PR #1, opened 2016-05-21
New structure of documentation and merge of contents

bipphilippe:
Merge of what was already in the documentation repository and was I wrote last weeks.

Split of the resulting documentation this way:
- the user's manual,
- the reference manual,
- the cookbook, so far empty - just as a placeholder and a call for contributors,
- the developer's guide.

Signed-off-by: Bernard Philippe <bip.philippe@laposte.net> (github: https://github.com/bipphilippe/openhab-docs)


bipphilippe:
Ooops...

I do not know why but the latest commits are not included in the PR.
I am investigating.

kaikreuzer:
Thanks - one question: Do you really insist on the xx_ prefixes? I understand that this makes it easier to have an overview on the file system, but I think it is very ugly to have this as part of the official urls in the end.
I think we should keep the urls as user-friendly as possible. I would therefore suggest to remove all prefixes as well as the underscores, so that we rather have something like http://docs.openhab.org/usermanual/basics/introduction instead of http://docs.openhab.org/10_users_manual/10_openhab_basics/00_openhab_basics_introduction

bipphilippe:
It was really helpfull in the merge process, but it is not critical for me to keep them.
I am going to remove them and do a PR.

kaikreuzer:
Thanks again :-)

bipphilippe:
New PR will be done very soon.
Just waiting for the last commit.

PR #4, opened 2016-05-30
Fixed English usage on index and overview

watou:
* In English, the noun __software__ never has an indefinite article preceding it ("a software").  Not present, but it is also never pluralized (softwares).  Information is also never pluralized (informations).
* In American English, __summarises__ is not recognized, whereas in British English, __summarizes__ is acceptable.  Should the documentation be written in American English?
* Provide missing URL to Eclipse SmartHome.
* Added commas so reader can breathe.

### DIFF
--- a/index.md
+++ b/index.md
@@ -6,18 +6,18 @@ layout: intro
 
 # Introduction to openHAB 2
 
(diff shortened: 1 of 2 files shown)


kaikreuzer:
Many thanks, please do continue :-)

> Should the documentation be written in American English?

I would say so, yes.

PR #12, opened 2016-06-04
Improve binding sorting (not 100%)

watou:
Workaround inability to [sort arrays from CSV files](https://github.com/jekyll/jekyll-help/issues/157) by pushing a new array and sorting that.  Also, sort by id because sorting by `field.toLowerCase()` does not work.  It's still not a perfect sort, but it's close.

### DIFF
--- a/addons/bindings.md
+++ b/addons/bindings.md
@@ -6,9 +6,11 @@ layout: documentation
 
 # List of Available Bindings
 
-{% assign bindings = site.data.bindings | sort: 'label.toLowerCase()' %}
+{% assign bindings = "" | split: "|" %}
+{% for binding in site.data.bindings %}{% assign bindings = bindings | push: binding %}{% endfor %}
+{% assign sorted_bindings = bindings | sort: "id" %}
 
 | Binding | Description |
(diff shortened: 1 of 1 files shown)


addons/bindings.md
```diff
@@ -6,9 +6,11 @@ layout: documentation
 
 # List of Available Bindings
 
-{% assign bindings = site.data.bindings | sort: 'label.toLowerCase()' %}
+{% assign bindings = "" | split: "|" %}
```
kaikreuzer: what is the "split" good for here? To make it an array? But doesn't it then contain an empty string as a first element?

watou: ["You cannot initialize arrays using only Liquid. You can, however, use the split filter to break a single string into an array of substrings."](https://help.shopify.com/themes/liquid/basics/types#initializing-arrays).  Some example (I forget where) showed this as a workaround, and my local testing didn't show an empty array element.

kaikreuzer: ok, thanks for the background - looks like a nasty workaround indeed ;-)

PR #23, opened 2016-06-12
updated rulesdsl.md 2nd attempt

udo1toni:
The first link had a missing # 
the second link makes no sense as it would point a dozen lines ahead without any additionally information.


kaikreuzer:
Perfect, thanks!

PR #24, opened 2016-06-12
Corrected configurations and added note about changes to HSBType

watou:


kaikreuzer:
Thanks!

PR #32, opened 2016-06-18
completed page regarding to existing pictures

udo1toni:


kaikreuzer:
Note that this file is generated.

udo1toni:
Ah. then I should have had more patience... ;)

PR #48, opened 2016-07-17
Fix digitalSTROM Readme typo

ThomDietrich:
This PR just fixes a small formatting problem. Check http://docs.openhab.org/features/bindings/digitalstrom/readme.html to see that the headline "Supported Things" is not shown as such.

However I found a bigger problem that might need improvement in the transformation system!? The digitalSTROM image (`DS-Clamps.jpg`) is too wide for my (small) notebook screen. Something like this should fix that:
```css
img {
    max-width: 100%;
}
```

### DIFF
--- a/addons/bindings/digitalstrom/readme.md
+++ b/addons/bindings/digitalstrom/readme.md
@@ -12,6 +12,7 @@ The integration happens through the digitalSTROM-Server, which acts as a gateway
 **Note:** All was tested with digitalSTROM-Server firmware version 1.9.3 to 1.10.3.
 
 ![various_digitalSTROM_clamps](doc/DS-Clamps.jpg)
+
 ## Supported Things
 
 ### digitalSTROM-Server


kaikreuzer:
Note that these files must not be edited in this repo (see https://github.com/openhab/openhab-docs#automatically-generated-parts).
I have created https://github.com/eclipse/smarthome/pull/1873 to fix the problem.

ThomDietrich:
Got it.
Is replacing the image by a smaller version really the solution?

kaikreuzer:
> Is replacing the image by a smaller version really the solution?

No, but a quick fix and I anyhow thought that the image is much too big (for any screen).
Feel free to come up with a css fix (but make sure that it does not have any unwanted impact at other places...)

ThomDietrich:
That's what I hoped to achieve with the line above. But you are right, we should be certain. I google searched for the problem and found this article, basically agreeing with me: https://wpbeaches.com/make-images-scale-responsive-web-design
To be precise, this setting will present every picture with the size of it's resolution as long as the width does not exceed the width of the screen.

(While this is the solution to the problem I saw, it's not good practice to have pictures with a random width in times of responsive webdesign. Some bootstrap grid magic would be better suited I guess. But that would __definitely__ have unwanted effects. So let's stay with max-width for now.)

PR #51, opened 2016-08-01
added explanation of Things

kaikreuzer:


ThomDietrich:
There is also the concept of a Bridge, which is documented nowhere (?). I came across it in the homematic article: http://docs.openhab.org/addons/bindings/homematic/readme.html
Should it be documented under concepts-things or is it homematic specific?

kaikreuzer:
They are mentioned here: http://docs.openhab.org/concepts/things.html

"A special type of thing is a “bridge”. Bridges are things that need to be added to the system in order to gain access to other things. A typical example of a bridge is an IP gateway for some non-IP based home automation system."

ThomDietrich:
Ah damn, had a quick look and didn't see a headline "Bridges" - but this seems sufficient. Thanks

Issue #49, opened 2016-07-28
Non-responsive images


ThomDietrich:
To resume https://github.com/openhab/openhab-docs/pull/48

I saw, that materialize is used for styling. So despite what I suggested in the other discussion, it would make more sense to rely on materialize to handle pics? http://materializecss.com/media-css.html

```css
<img class="responsive-img"...>
```
The effect is more or less the same: 
```css
img.responsive-img {max-width: 100%; height: auto;}
```
I have no experience with jekyll, would be great if somebody could quickly provide the needed change.

PR #56, opened 2016-08-04
Fix scrollbar bug, Add responsive images

ThomDietrich:
Fixes https://community.openhab.org/t/documenting-openhab-2/10568/84
Fixes https://github.com/openhab/openhab-docs/issues/49

Sorry for piggybacking :o)

The "bug" behind Link 1 is actually a feature added by the original author of style.css to be able to look at big pictures.
I included the changes suggested by Link 2 to fix that in the right way.
However not as a class, I like this solution better after all.

Outlook: At some point in the future, if one image needs to be half size or with a pink frame, this would be possible by adding a class:

```
![title](image.png){: .half-pink-img}
```

### DIFF
--- a/concepts/index.md
+++ b/concepts/index.md
@@ -22,7 +22,7 @@ Channels can be linked to multiple Items and Items can be linked to multiple Cha
 
 To illustrate these concepts, take a two-channel actuator that controls two lights:
 
-![](images/thing-devices-1.png){: .responsive-img}
+![](images/thing-devices-1.png)
 
 The actuator is the _Thing_. This might be installed in the electrical cabinet, it has a physical address and needs to be setup and configured in order to be used.
 The user is instead interested in the two lights, which are located at different locations in his home. These lights are the desired functionality, thus the _Items_ and they are linked to the _Channels_ of the actuator.
--- a/concepts/index.md
+++ b/concepts/index.md
@@ -22,13 +22,8 @@ Channels can be linked to multiple Items and Items can be linked to multiple Cha
 
 To illustrate these concepts, take a two-channel actuator that controls two lights:
 
-![](images/thing-devices-1.png)
+![](images/thing-devices-1.png){: .responsive-img}
 
 The actuator is the _Thing_. This might be installed in the electrical cabinet, it has a physical address and needs to be setup and configured in order to be used.
 The user is instead interested in the two lights, which are located at different locations in his home. These lights are the desired functionality, thus the _Items_ and they are linked to the _Channels_ of the actuator.
 A _Link_ can be regarded like a physical wire in this example.
-
-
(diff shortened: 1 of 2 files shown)


concepts/index.md
```diff
@@ -27,8 +27,3 @@ To illustrate these concepts, take a two-channel actuator that controls two ligh
 The actuator is the _Thing_. This might be installed in the electrical cabinet, it has a physical address and needs to be setup and configured in order to be used.
 The user is instead interested in the two lights, which are located at different locations in his home. These lights are the desired functionality, thus the _Items_ and they are linked to the _Channels_ of the actuator.
 A _Link_ can be regarded like a physical wire in this example.
-
```
kaikreuzer: Note that you actually need to fix [this file](https://github.com/eclipse/smarthome/blob/master/docs/documentation/concepts/index.md), since you are only looking at a copy here, which will get overwritten again.

css/styles.css
```diff
@@ -497,10 +497,6 @@ section#developers .section-header {
```
kaikreuzer: Could you create another PR for https://github.com/eclipse/smarthome/blob/master/docs/css/styles.css as well to keep it in sync?

ThomDietrich: Done. Will you merge both or is the openHAB one again replaces by the esh one?

kaikreuzer: We need it here as well as the styles.css is not copied by the doc build (shall we copy it as well?)

ThomDietrich: I can't really say. While I understand the need to 1. have two independent docs and 2. copied articles, I fear that the whole thing will explode on you at some point :O
The building process seems a bit unclear right now. There is no automatic build after PR merge, it is hard to understand what is copied and why there are duplicate files (Contribution for example).

Would it be better to have the whole eclise smarthome docs as a submodule (to make realtions clear to everyone) and to work with symlinks (if jekyll and friends like that)? Without symlink an idea would be to have an sh script (like you have now) copying files with an additional step, marking them as copied. An added YAML front matter "page.source" in the head of the file would be a nice approach I guess.

ThomDietrich: And to answer your question, yes I think so, we should copy as much as possible but add openhab specific things on top. A styles2.css loaded after style.css could be a place for everything orange and openhab'y.

kaikreuzer: > to work with symlinks (if jekyll and friends like that)?

I think this won't work, at least I wasn't very successful when I experimented with such a solution.
What we will indeed need sooner or later is a better manageable copy mechanism, doing it in the pom.xml as it is now will definitely become too complex at some point in time.

> yes I think so, we should copy as much as possible

Feel free to adapt the copy mechanism accordingly.

PR #65, opened 2016-08-07
Update installation overview and RPi article

ThomDietrich:
* a new installation overview article with all a user might need to know and decide
* the raspberry pi article with a bit more context 
* attention: I decided to move prerequisites into the overview as it felt right there. I'm open for a discussion

@-all: Would be great if you could proof read and tell me about missing information ;) Thanks


kaikreuzer:
Didn't proof read it in detail but it looks like a lot of useful information (as always), so I merge it right away!

ThomDietrich:
Thanks and Thanks... and Thanks ;)

PR #71, opened 2016-08-11
Add visual separators to menu

ThomDietrich:
Just an idea I was playing with...
What do you think?

2016-08-11 22_46_26-installation overview - openhab - empowering the smart home


kaikreuzer:
Yeah, why not!
Probably you should directly introduce it in the styles.css of ESH and copy it over here as discussed.

ThomDietrich:
Sure! I will use that moment to fix up a bunch of details I stumbled over.

PR #83, opened 2016-09-05
Update the snaphot upgrade explaination

ThomDietrich:


kaikreuzer:
Is this a temporary text? When do you plan to remove the "old" version?

ThomDietrich:
Yes it is temporary. I thought it would be best to make clear that reinstall is not needed anymore. I'll remove the whole thing as soon as yet another thing about installation changes. theo is working on the repository right now. I'm sure there will be things to document.

PR #87, opened 2016-09-07
Fixed README section for accessing Karaf console

eschizoid:


ThomDietrich:
I suspect @rkoshak to do some further changes to this article in the next few days. This is your heads up in case you already created a branch to modify docker.md ;)

PR #89, opened 2016-09-07
Changed openhab.list to openhab2.list

rkoshak:
As part of my efforts to run through a migration from OH 1 to OH 2 I hit a snag with the snapshot apt repos. While debugging this problem I noticed that the instructions have the user overwire openhab.list. But if they had already installed OH 1 through apt-get this file already exists, along with some others (openhab.list.DistUpgrade, openhab.list.save). 

To avoid intermingling the OH 1 and OH 2 repos in this way I propose moving to use openhab2.list.

If this is undesirable I can make a note in the migration docs instead.

### DIFF
--- a/installation/linux.md
+++ b/installation/linux.md
@@ -48,17 +48,17 @@ Alternatively resort to the [manual installation approach](linux.html#manual-ins
 Choose between the latest Beta release or a Snapshot with all incoming contributions, rebuild on [every change](https://oss.jfrog.org/webapp/#/builds/openHAB-Distribution).
 As openHAB 2 is still in an evolving state, the snapshot may be the **preferred choice**.
 
-The repository URLs will be stored in the file `/etc/apt/sources.list.d/openhab.list`.
+The repository URLs will be stored in the file `/etc/apt/sources.list.d/openhab2.list`.
 Be careful to not have conflicting repositories in your sources list.
 
 Decide between two options:
 
 * **Beta Release**
 
-  Add the **openHAB 2 Beta repository** to your systems apt sources list (will overwrite your existing `openhab.list`):
+  Add the **openHAB 2 Beta repository** to your systems apt sources list (will overwrite your existing `openhab2.list`):
 
   ```shell
-  echo 'deb http://dl.bintray.com/openhab/apt-repo2 testing main' | sudo tee /etc/apt/sources.list.d/openhab.list
+  echo 'deb http://dl.bintray.com/openhab/apt-repo2 testing main' | sudo tee /etc/apt/sources.list.d/openhab2.list
   ```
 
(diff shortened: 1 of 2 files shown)


kaikreuzer:
@ThomDietrich & @theoweiss Ok for you?

ThomDietrich:
I had this thought when writing the docs too.
Looking at our filesystem, there are a few places where "openhab2" is the used term (`/etc/openhab2`, `usr/share/openhab2`, `/lib/systemd/system/openhab2.service`). I agree with this change, in fact if there is any other file/folder only named "openhab", we should think about renaming this one too.

By the way, I'm already going this way in the openHABian setup ([source](https://github.com/ThomDietrich/openhabian/blob/refactor-menu/openhabian-setup.sh#L135-L136)) :smiley:

installation/linux.md
```diff
@@ -48,17 +48,17 @@ Alternatively resort to the [manual installation approach](linux.html#manual-ins
 Choose between the latest Beta release or a Snapshot with all incoming contributions, rebuild on [every change](https://oss.jfrog.org/webapp/#/builds/openHAB-Distribution).
 As openHAB 2 is still in an evolving state, the snapshot may be the **preferred choice**.
 
-The repository URLs will be stored in the file `/etc/apt/sources.list.d/openhab.list`.
+The repository URLs will be stored in the file `/etc/apt/sources.list.d/openhab2.list`.
 Be careful to not have conflicting repositories in your sources list.
```
ThomDietrich: Do you think there should be a mentioning that the user may need to delete `openhab.list`? Yes, that should be part of the migration tutorial but I'm almost certain this will hit users migratin usin their "best knowledge" and this guide.

If yes, needs to be changed further down too

rkoshak: > I had this thought when writing the docs too.

Great minds think alike. :-)

I plan on having a "clean-up" step in the migration docs that covers removal of OH 1 and it will 
include removal of the old openhab.list files among other things.

I'm trying to structure the migration steps so the OH 1 install is not broken until the OH 2 install is completed and proven to work.

If you know of other files or folders that should be removed or renamed let me know so I don't miss them in the setup. With your work on openHABian you might be one of the most knowledgeable on that front.

ThomDietrich: The only file I can think of right now, not (entirely) named openhab2 is `/var/log/openhab2/openhab.log`.

For the cleanup: depends on the setup. Here is what I would name the hammer method:

* backup config and userdata (icons etc)
* stop service
* purge
* delete
  * /opt/openhab
  * /etc/openhab
  * /usr/share/openhab
  * service file
  * logs
  * other known files by openhab1 or openhab2 if available 

I'm not sure about the exact path names.

theoweiss:
lgtm

Issue #91, opened 2016-09-08
Karaf console command and example do not match


rkoshak:
On http://docs.openhab.org/administration/console.html the instructions say:

> If openHAB runs a service, the console can be accessed using ssh to the openHAB host on port 8101. The default Username/Password is openhab/habopen.

However, the provided output of the command shows the user having logged in as user karaf.

> ssh karaf@localhost -p 8101

This should be

ssh openhab@localhost -p 8101

I'd fix it myself but I don't yet know how to have more than one pull request at a time.

ThomDietrich:
Hey @rkoshak, there is no same in that :) You can actually learn quite a lot going through the excellent GitHub Guides https://guides.github.com - have fun!

One pull request is always linked to one branch. A few minutes ago you [already learned](https://github.com/openhab/openhab-docs/pull/89/commits), that commits pushed to a branch will by that also be part of the pull request, which is great for tuning your contribution till it's accepted by the repository maintainers. The solution to what you are wondering is to always create a new branch for a new pull request. You should never commit to or pull-request from the main branch (h-pages in this case).

To now create a second pull request on the issue here create a new branch named "karaf-login" and commit your changes to it. Important! As your gh-pages now has it's own decoupled commits compared to the origin (the repository here), be sure to create the branch based on the latest commit from here!

rkoshak:
Thanks! I'll give that a try tomorrow.

PR #94, opened 2016-09-12
Fix the command to access the console

lolodomo:
and add an advice in case the command fails.

### DIFF
--- a/administration/console.md
+++ b/administration/console.md
@@ -17,12 +17,12 @@ The Karaf console offers the option to:
 
 Accessing the Karaf console depends on the start mode of openHAB. The manually start using shell/batch script ends directly in the Karaf console.
 
-If openHAB runs a service, the console can be accessed using ssh to the openHAB host on port 8101. The default Username/Password is **openhab/habopen**.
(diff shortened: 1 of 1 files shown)


lolodomo:
Will fix #91

ThomDietrich:
Thanks for the fix! I do however not feel comfortable with the "advice". This file is quite important in the security concept of an ssh installation and users should not be invited to blindly delete it.
What is the problem you are trying to solve by that step?

lolodomo:
https://community.openhab.org/t/ssh-session-timing-out-on-slow-hardware/14008
Something was changed in the security setup and I think my SSH connection with the old OH setup was saved in known.hosts, preventing a new connection with new OH security setup.

ThomDietrich:
In this case you should have seen a warning, telling you to delete one line from that file, just as Red_Wraith suggested. Deleting only one specific line when prompted to do so is okay. Did you get that warning? In this case I don't think a mentioning in docs is needed.

maggu2810:
You can delete the old entry in the known hosts file using "ssh-keygen -R ..." or you can use command line parameters for "ssh" to "ignore" the known hosts file for this connection.

ThomDietrich:
Hey @lolodomo, I've tuned your contribution a bit. Are you okay with them?

lolodomo:
No problem

ThomDietrich:
Merged. Thank you!

PR #96, opened 2016-09-19
Disable proxy_buffering in nginx example

BClark09:
Using the default, proxy buffering causes very slow synchronisation issues with states because it waits for eventsource types (such as http://openhab.domain.tld/rest/events?topics=smarthome/items/*/state) to complete.


BClark09:
Just to see interest would it be worth writing a basic tutorial for making an nginx reverse proxy and generating a valid certificate with certbot for use with openhab?

Issue #98, opened 2016-09-21
nginx example is missing two configurations


BClark09:
I believe that the documentation for the nginx reverse proxy is misconfigured: I believe that proxy_buffering by default should be disabled, since the server will try to wait for the full message recieved from event source objects such as //rest/events?topics=smarthome/items/*/state.
```
proxy_buffering                         off;
```

Secondly, the server at that configuration will not forward https links in REST unless the following is specified:
```
proxy_set_header X-Forwarded-Proto      $scheme;
```
In the documentation "X-Forwarded-Scheme" is set but I don't think this is a valid header. I will send a pull request for the above but wanted to bring it as an issue in case I've missed any reasoning behind this.

Finally, would it be worth it to make a full tutorial on using an nginx reverse proxy with openhab, which includes setup, basic authentication and ssh encryption/renewal?

ThomDietrich:
Thanks for your addition! I'm no expert in that field but it seems like you know what you are doing. Could you post one-two URLs regarding these topics?

"Tutorial" - absolutely! Are you talking about this page? http://docs.openhab.org/configuration/jetty.html
Please extend it as far as you like :wink: Especially interested in a writeup on authentication. Remember that there are already postings regarding that topic in the community forum. Please have a look to cover everything.

On a side note: I wanted to add both authentication and SSL to openHABian as optional steps. When you are in the process of collecting the needed steps, feel free to report them here: https://github.com/ThomDietrich/openhabian/issues/12 Thanks! ;)

BClark09:
Not a problem! I'm actually fairly new to nginx, but I was having a couple of issues with it to try and investigate the cause:

Firstly, when doing a simple benchmark on the Basic UI, I found that items were taking ages to update using the debugging tools you can see that one of the items gets stuck on pending:

![image](https://cloud.githubusercontent.com/assets/8983584/18717579/3f5878e2-8018-11e6-8ba4-52452498aa24.png)

Secondly, `curl -X GET --header "Accept: application/json" "https://xxx.xx.xx/rest/sitemaps"` always returned: ` "link": "http://xxx.xx.xx/rest/sitemaps/default",`

After reading up on how nginx caches the response from the server: https://www.nginx.com/blog/nginx-caching-guide/ I assumed that the server was holding the request until it completed, setting proxy_buffering to off cleared the first issue.

[...]

PR #99, opened 2016-09-21
Address problems with nginx configuration example

BClark09:
Address the two points raised in #98 

Include X-Forwarded Protocol
Disable Proxy buffering by default


ThomDietrich:
Following from https://github.com/openhab/openhab-docs/issues/98

Hey this looks pretty good already. What came to mind immediately is a problem with these two sentences: 
"This means you can proxy connections to http://openhab.mydomain.tld to your openHAB runtime."
"Once installed, you can test to see if the service is running correctly by going to http://openhab.mydomain.tld"

This sounds like I can have my openHAB installation answer at "openhab.whitehouse.com" Cooool!! :smile:  

I think it should be made clear, that the option of serving under your own domain is only relevant to you if you in fact own a domain - which 99% of users will not.
However this goes on. Even if you do own a domain and are able to assign a subdomain, you do not necessarily need a proxy server. You can just point the domain to the servers IP and off you go. Most users will not be able to do this because their provider doesn't offer a static IP address. And voila you need to use a DynDNS which in return will already assign a domain which you could now even CNAME in your DNS settings... 
Besides the above scenario you may own a public webserver and could connect both servers by a VPN or similar and use the reverse proxy as shown... but as I said, 99%...

To me it seems like the domain topic and possibilities to "get" one should be explained more!?

BClark09:
That's fair enough! It may be difficult to cover all of the options and reasons for doing so on here, (I for example own a domain name, which I link to a non-static IP Address, although it doesn't change unless I disconnect the router for a few days.) 

However, I do agree that adding a domain section to explain the basics is necessary, since the HTTPS section is only valid on a domain. Do you think that the necessary changes should be:

- Explain the options involved in obtaining a domain name (free or otherwise)
- Explain that if you don't want to use a domain, show that you can use the IP address instead.
   - Replace  *openhab.mydomain.tld* with *mydomain_or_myip*
- Say that the benefits of using a proxy makes authentication simple (atm, possible) and that using a domain name makes generating valid HTTPS easier.
- Explain the method of obtaining HTTPS without a domain name.

If we do the last point, would it be better to split the Reverse Proxy into a category, containing the following pages:

- **nginx.html**: For the basic setup of nginx
- **auth.html**: For Authentication with nginx, the steps of which are the same regardless of domain/ip
- **https.html**: For adding certificates with nginx, via Let's Encrypt with a domain, and protecting your server with additional parameters. (And to add a part about OpenSSL for those without a domain after this PR is merged)
- **complete.html**: To show examples of all three of these steps completed.

ThomDietrich:
This is more complicated the more we think about it :)
Let's stay with one article for now.

I would propose to start with authentication. It's an often asked for topic and a simple openHAB2+nginx+auth would be what most readers will look for. Auth is not directly related to https or a domain, readers are free to read on or stop at this point.

Next the proposed "domain section" would be great. I think it is enough to state the possibilities and maybe link external articles on these topics. DynamicDNS is probably what most users will go with.

At last your already available chapters can explain how to configure nginx for a domain and https.

Another idea I want to run by you: I realized, that it is hard to write an article for the docs completely without feedback. How would you feel about additionally posting your article to the community forum and receiving questions and feedback on it?

BClark09:
Thanks for the advice @ThomDietrich, here's another draft. I've added instructions about HTTPS without a domain name so that all bases are covered within the same page. Let me know what you think, I can put this latest version on the forum for feedback too.

ThomDietrich:
Thanks for the addition. I saw your forum posting, let's see what comes additionally out of that.

PR #104, opened 2016-09-26
Wrong link corrected

Boby71:
Correction of wrong link "own configuration" which led to a 404 error


ThomDietrich:
Thanks Boby for pointing out that wrong link! As the link to jetty or anywhere else is not quite where "own configuration" should point to, I removed the link completely *for now*.

Issue #103, opened 2016-09-23
Check for broken links regulary


ThomDietrich:
With the current need to rewrite and restructure docs articles, it could be handy to know which links need to be changed.

http://www.deadlinkchecker.com/website-dead-link-checker.asp?u0=http://docs.openhab.org
##

ThomDietrich:
Additionally, transform absolute docs links to relative

BClark09:
You can also use [linkchecker](https://wummel.github.io/linkchecker/)

e.g:
```bash
sudo apt-get install linkchecker
linkchecker -F text/invalidlinks.txt --check-html --check-css --check-extern --ignore-url=^.*/font/ http://localhost:4000/
```

Another (but much less forgiving) alternative is using [html-proofer](https://github.com/gjtorikian/html-proofer#using-with-jekyll) you could even run this with a travis build!

PR #119, opened 2016-10-20
Replace and fix invalid links

### LINKED ISSUE #117: 404 Page Breaks after third 'depth'
My latest change works up until the third sub-directory for example: http://docs.openhab.org/one/two/three

It looks like the header and footer isn't applying base.html as intended and I'm not entirely sure of the reasoning behind the process. Why is {{base}} used all over in the first place? Surely a working  `<script src="{{base}}/js/jquery.min.js"></script>` is the same as `<script src="/js/jquery.min.js"></script>` etc. which sets the initial level? 

What is the necessity to calculate {{base}} by breaking down the target url?
##

BClark09:
Fixes #117 (and the change here is discussed in the comments) and addresses all of the current broken links to openHAB pages at the time of the PR.

### DIFF
--- a/_includes/base.html
+++ b/_includes/base.html
@@ -1,5 +1,4 @@
 {% assign base = '' %}
-{% assign base = '' %}
 {% comment %} <!--Disabled since the base in docs.openhab.org can only be '/'-->
 {% assign depth = page.url | split: '/' | size | minus: 1 %}
 {% if    depth == 0 %}{% assign base = '.' %}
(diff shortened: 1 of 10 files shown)


ThomDietrich:
Great contribution as always ;) Thanks

ThomDietrich:
How did you check for invalid links? I used this page https://github.com/openhab/openhab-docs/issues/103 the last time. If you've got a bash script ready to do the job: Care to share?

BClark09:
I had a difficult time trying to make a bash script to test broken links within the sites are because of how some of them are autogenerated.

I used linkchecker (which should be available for most debian packages) to help me.

```
sudo apt-get install linkchecker
linkchecker -F text/invalidlinks.txt --check-html --check-css --check-extern --ignore-url=^.*/font/ http://localhost:4000/
```

ThomDietrich:
Good enough :+1: Can you add this to the issue?

PR #120, opened 2016-10-25
Correct to set the IP address of the Hue Bridge in Things?

Ion3:
Is it correct to place the IP address here? If I do so, Paper UI shows me two Bridges even though I uesed the correct IP address of the bridge in my environment. The one I added in the Things shows up as "Hue Bridge UNINITIALIZED" and the original discovered in the Paper UI is still called "Philips hue (192.168.178.18) ONLINE".

### DIFF
--- a/addons/bindings/hue/readme.md
+++ b/addons/bindings/hue/readme.md
@@ -92,7 +92,7 @@ In this example **Bulb1** is a standard Philips HUE bulb (LCT001) which supports
 ### demo.things:
 
 ```
-Bridge hue:bridge:1 [ ipAddress="192.168.0.64" ] {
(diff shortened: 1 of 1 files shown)


ThomDietrich:
PaperUI will still show a thing if the identifier of this thing is different to the one you used ("hue:bridge:1"). I saw that behavior with other bindings and it doesn't make problems. If you want to get rid of the thing in PaperUI, name your own one the same. If you have problems getting your Bridge to work, it's probably not related to the fact that there is "another" bridge in PaperUI.

Either way, please ask for guidance in the forum, your chances are way higher to receive an answer: https://community.openhab.org/tags/hue
Please do not abuse the PR feature of GitHub to ask questions. If you found the solution to your question and there is in fact something wrong with the Hue article in the documentation, please reopen this ticket!

PR #121, opened 2016-10-26
 Update working synology devices

cniweb:
openhab/openhab-syno-spk#29

Signed-off-by: Christian Häussler c-n-i@web.de (github: cniweb)


ThomDietrich:
Thanks for the addition! Do you think it's wise to link the exact versions inside the table? May these become outdated over time?

PR #123, opened 2016-10-30
changed JDK to JVM

kaikreuzer:
fixes https://github.com/openhab/openhab-distro/issues/297


ThomDietrich:
I never knew why this was a must. I guess that mystery is solved now. ^^

PR #125, opened 2016-11-01
Update Amazon dash button Addon  readme.md

beinnlora:
Updated amazon dash button addon readme with example, to highlight how to trigger rules using Channel events, as opposed to the more common Item events

### DIFF
--- a/addons/bindings/amazondashbutton/readme.md
+++ b/addons/bindings/amazondashbutton/readme.md
@@ -86,6 +86,7 @@ Background discovery is supported. Just press the button in order to put it into
 ## Example Use
 
 The binding uses a trigger channel which doesn't have to be linked to an item. You can easily use it in your rules like this:
+
 ```
 rule "Dash button pressed"
     when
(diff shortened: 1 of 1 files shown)


addons/bindings/amazondashbutton/readme.md
```diff
@@ -82,3 +82,15 @@ Background discovery is supported. Just press the button in order to put it into
 ## Channels
 
 * Press: Channel for recognizing presses on the Amazon Dash Button
+
+## Example Use
+
+The binding uses a trigger channel which doesn't have to be linked to an item. You can easily use it in your rules like this:
+```
```
kaikreuzer: Could you please add an empty line here as Jekyll will otherwise break the layout on docs.openhab.org?

beinnlora: Before the code comment - between lines on line 88 and 89? I have now added that. Sorry, I am not very familiar with Jekyll or github.

kaikreuzer:
Yes, exactly like this, thanks!

kaikreuzer:
I didn't notice that this was a change on a generated file when merging. 
FTR: this change will be overwritten by https://github.com/openhab/openhab2-addons/pull/1380.

beinnlora:
Neither did I, thanks for the info.

PR #127, opened 2016-11-04
Owntracks 1.9 binding information added

smar000:


_data/oh1addons_infos.yml
```diff
@@ -161,6 +161,10 @@
   description: This binding allows openHAB to act as an MQTT client, so that openHAB items can send and receive MQTT messages to/from an MQTT broker
   wiki_url: https://github.com/openhab/openhab/wiki/MQTT-Binding
 
+- label: Owntracks Binding
+  description: Owntracks (formerly known as Mqttitude) provides device location data via an mqtt bridge to openHAB
```
watou: It sort of doesn't matter, but it was called MQTTitude before OwnTracks (note differences in capitalization).  Also mqtt would be better as MQTT.

_data/oh1addons.csv
```diff
@@ -57,6 +57,7 @@ binding,Nikobus Binding
 binding,Novelan Heatpump Binding
 binding,OpenEnergyMonitor Binding
 binding,OneWire Binding
+binding,Owntracks (mqttitude) Binding
```
smar000: Ok. Can I leave this as is for now or do I need to do anything else?

_data/oh1addons_infos.yml
smar000: Done.

_data/oh1addons.csv
kaikreuzer: you can leave it, it will be automatically overwritten once I update the generated content.

PR #132, opened 2016-11-10
added audio and voice console commands

kaikreuzer:
and sorted the rest a bit


ThomDietrich:
Thanks. I was always curious why there are no ":" in between the commands.
I've tweaked the markdown formatting. Pay attention that angle brackets are interpreted as html tags:
`this does <not> work` : this does <not> work

kaikreuzer:
Thanks for fixing the markdown :-)

ThomDietrich:
Removed the headers again. Just didn't look good in the end.

PR #139, opened 2016-11-17
Use adduser rather than useradd for adding groups to user

toby200:
`useradd` will only work if user is new, and apt-get install process creates the user already, so changed to use usermod instead.

`adduser` would be viable for most systems and is simpler, but usermod is probably more widely available, so a bit of a toss up between the two. Happy to change it if it's preferred to use `adduser user group` instead.

Signed-off-by: Toby Shepheard <shepheardt@yahoo.com> (github: toby200)

### DIFF
--- a/installation/linux.md
+++ b/installation/linux.md
@@ -398,8 +398,8 @@ The following example shows how to add Linux user `openhab` to the often needed 
 Additional groups may be needed, depending on your hardware and software setup.
 
 ```shell
-sudo useradd openhab dialout
-sudo useradd openhab tty
+sudo adduser openhab dialout
+sudo adduser openhab tty
 ```
 
 Additionally it's needed to allow the java environment to access the serial port of the connected peripheral.


BClark09:
Yes, I agree! I missed this when I reviewed the recent PR that added this, but unless @ThomDietrich has any reason to keep the previous then I think this would be much better.

ThomDietrich:
Hey guys, at first I was confused about this Issue. Then I realized that the problem is in the original PR! It's supposed to be adduser instead of useradd. I prefer adduser over user*. 
http://askubuntu.com/a/381646
If you agree, please adapt the PR and I'll merge

toby200:
Sorry the PR title was wrong, I started out thinking the problem was just the syntax, when in fact it was that the user already existed, and so amended the original commit but never updated the PR title.

Should I amend the original commit and force push, or create a new commit on top of the original?

ThomDietrich:
adduser is the better option imho. Thanks for adapting. @BClark09 do you agree?

BClark09:
Yep,  different Linux implementations use the address script in a manner more suitable to them so this command is probably safer.

PR #140, opened 2016-11-17
Rewrote the Migration Tutorial

### LINKED ISSUE #86: Add step by step instructions and lessons learned to Migration from 1.x tutorial
The existing "Migration from openHAB 1 to openHAB 2" has a lot of important information about what will be different between OH 1.x and OH 2 but it really isn't a tutorial. there are no step by step instructions. I'm going to start the migration process myself and will use my experiences to beef up these instructions.
##

rkoshak:
Rewrote based on the [tutorial on the forum](https://community.openhab.org/t/migration-from-oh-1-x-to-oh-2-tutorial/13947). I've done a good deal of editing to preserve the content that was already there and to add in instructions for using addons.cfg. Spelling and grammar checks out and links have been tested.

Closes #86 

Signed-off-by: Rich Koshak <rlkoshak@gmail.com> (github: rkoshak)


tutorials/migration.md
```diff
@@ -4,96 +4,846 @@ layout: intro
...
+practices?
+- Have you wanted to migrate to a container (e.g. 
+[Docker](../installation/docker.html) or a virtual machine?
+- Do you want to start over fresh on a brand new OS install?
+
+Your answers to these and other similar questions will help guide you to 
+prepare your installation before you start the migration. For example, if you
+want to start using git to back up and configuration manage your openHAB 
+configurations, or if you are considering moving to 
+[Docker](../installation/docker.html) or a VM or starting over fresh (e.g.
```
BClark09: same as above.

tutorials/migration.md
```diff
@@ -4,96 +4,846 @@ layout: intro
...
+
+The following procedure primarily focuses on an in-place migration. If you are
+migrating to a fresh install, container, or another VM some of the following
+steps can be skipped. Those steps are identified.
+
+One major consideration is that at the time of this writing openHAB 2 does not
+implement authentication and authorization (i.e. no username and password). If
+you are relying upon port forwarding to access your openHAB server remotely
+instead of via a VPN, SSH tunneling, or [my.openhab](https://my.openhab.org/),
+we highly recommend [setting up a reverse proxy](../configuration/nginx.html).
```
BClark09: same as above, won't comment on further uses of ../

tutorials/migration.md
```diff
@@ -4,96 +4,846 @@ layout: intro
...
+
+There are three approaches one can use to install and configure bindings, an 
+all text based one, using the Karaf console, or using one of the administration 
+GUIs (i.e. PaperUI or Habmin 2). While all three approaches are presented below, 
+the text based approach is the recommended one for those coming from openHAB 1.x 
+as it will be more familiar and flexible.
+
+###Text Based
+
+In your configurations folder for openHAB 2 (`/etc/openhab2/conf` on apt-get
```
BClark09: This is just `/etc/openhab2`

tutorials/migration.md
```diff
@@ -4,96 +4,846 @@ layout: intro
...
+- `autoapprove:enabled=false` - makes it so you must manually approve 
+automatically discovered Things in the Inbox (see below)
+- `org.eclipse.smarthome.links:autoLinks=false` - prevents openHAB from 
+automatically creating an Item for each automatically discovered Thing.
+
+Next open for editing `addons.cfg`. This file contains parameters related to 
+add-ons including allowing one to choose which add-ons to install. For now just
+set the following parameters:
+
+- `package = minimal` - minimal or standard are good choices here. Standard
```
BClark09: Should we really recommend a minimum package to a standard user?

tutorials/migration.md
```diff
@@ -4,96 +4,846 @@ layout: intro
...
+PaperUI for those migrating to openHAB 2 from openHAB 1.x.
+
+###Configure Add-ons
+
+As bindings and other add-ons are installed, you can watch for errors in the 
+logs. Errors at this time may not be important but make note of those that did
+generate errors during installation for special attention later.
+
+Once a binding or add-on is installed, it will create a 
+`conf/services/<add-on name>.cfg` file. Unlike in openHAB 1.x where all the
```
BClark09: ...for manual installations, since apt packages are /etc/openhab2/services/<add-on name>.cfg

tutorials/migration.md
```diff
@@ -4,96 +4,846 @@ layout: intro
...
+##Preparation
+
+Now is the time to consider and plan for your newly installed openHAB. Some 
+questions to ask and answer include:
+
+- Are you happy with your current deployment and maintenance of the deployment?
+- Do you have or need to change your backup and configuration management 
+practices?
+- Have you wanted to migrate to a container (e.g. 
+[Docker](../installation/docker.html) or a virtual machine?
```
BClark09: The base URL is usually set to {{base}}/installation/docker.html, if the tutorials moved into a new folder (say "guides/tutorials" then {{base}} will point to a different url.

tutorials/migration.md
```diff
@@ -4,96 +4,846 @@ layout: intro
...
+  Unresolved requirement: Import-Package: org.openhab.binding.astro.internal.calc
+```
+
+Note, the above error is for illustration purposes. You should not be 
+installing Astro this way.
+
+###Final Add-ons Installation Steps
+
+openHAB 2 has a different 
+[folder layout](http://docs.openhab.org/installation/linux.html#file-locations).
```
BClark09: This shouldn't be an absolute path, consider `{{base}}/installation/linux.html#file-locations` (or  at least `../` as you have used previously)

tutorials/migration.md
```diff
@@ -4,96 +4,846 @@ layout: intro
...
+
+Note that the path to the Sitemap UIs have changed. If one goes
+to `http://<hostname>:8080` you will see a list of all the installed UIs, 
+including the administration UIs. To get to ClassicUI the URL is
+`http://<hostname>:8080/classicui/app?sitemap=<your sitemap>`. If you previously
+used `default.sitemap` as your file name so you could skip the `?sitemap=` part
+you must now rename the file to `_default.sitemap` and change the very first
+line of the file to read `sitemap _default label="<sitemap name>`. 
+
+openHAB 2 only supports static webviews. The dynamic webview some
```
BClark09: That's not strictly true, the compatability layer still hold support for dynamic webviews, but the formatting is different and slightly confusing:

My basic-ui:

image

tutorials/migration.md
```diff
@@ -4,96 +4,846 @@ layout: intro
...
+In addition, to test an Item to see if it is Undefined one tests against `NULL`
+instead of `Undefined`. For example `if(MyItem.state == Undefined)` becomes
+`if(MyItem.state == NULL)`. Note that `NULL` is not to be confused with `null`.
+`NULL` is only valid when testing an Item's state to see if it is undefined.
+`null` is used pretty much everywhere else to mean "no value" and usually 
+indicates no result or an error.
+
+Finally, the `HSBType` state type can no longer be constructed using a `java.awt.Color` object, and there is no longer a `toColor()` method.  Use the following alternatives:
+
+~~~~
```
BClark09: Does "\`\`\`java" not work in this case? (FYI, Other pages have been using "\`\`\`xtend")

tutorials/migration.md
```diff
@@ -4,96 +4,846 @@ layout: intro
...
+If you haven't already, configure default persistence now. You can do this 
+through `runtime.cfg` or PaperUI.
+
+Now openHAB 2 is ready to receive your other configurations, items and such. 
+Watch the log file as you transfer over files for errors. These errors will 
+have to be fixed.
+
+Copy over your files in the following order:
+
+~~~~
```
BClark09: Wouldn't \`\`\`bash be more appropriate?

tutorials/migration.md
```diff
@@ -4,96 +4,846 @@ layout: intro
...
+
+When you save this file openHAB will begin the download (if using the Online 
+version) and installing the add-ons.
+
+Note: if running in Docker, create <openHAB 2 home>/userdata/persistence prior to saving the 
+file if there are any persistence add-ons.
+
+An example addons.cfg:
+
+~~~~
```
BClark09: I believe this should be at least \`\`\`text

tutorials/migration.md
```diff
@@ -4,96 +4,846 @@ layout: intro
...
+- Old syntax that was `[minbound,maxbound]` has to be changed to 
+`[minbound..maxbound]`. 
+- Note that you now have the ability to exclude bounds from the ranges (e.g. 
+`]minbound..maxbound]`) and also define open ranges`[minbound..]`
+
+
+###Sitemap
+
+If you use png icons, you must change the default icons from svg to png for
+ClassicUI and BasicUI. This can be done in PaperUI in `Configuration -> Service->BasicUI and ClassicUI`.
```
BClark09: Space between first -> but not the last.

tutorials/migration.md
```diff
@@ -4,96 +4,846 @@ layout: intro
...
+
+In openHAB 1.x (and importantly still for 1.9 version bindings running in openHAB
+2) one had to add 
+["binding configuration"](https://github.com/openhab/openhab/wiki/Explanation-of-items#binding-config) 
+in curly brackets to the end of Item definitions in order to link that Item to 
+some control point on a physical device or API. The 
+[openHAB wiki](https://github.com/openhab/openhab/wiki) gives a nice examples of 
+how this looks:
+
+~~~~
```
BClark09: Again, wouldn't \`\`\` be more appropriate? I won't comment on further uses of this.

tutorials/migration.md
```diff
@@ -4,96 +4,846 @@ layout: intro
...
+have to be fixed.
+
+Copy over your files in the following order:
+
+~~~~
+cp <openHAB 1.x conf>/configurations/transform/* <openHAB 2 conf>/transform/*
+cp <openHAB 1.x conf>/configurations/scripts/* <openHAB 2 conf>/scripts/*
+cp <openHAB 1.x conf>/configurations/persistence/* <openHAB 2 conf>/persistence
+Copy any custom icons added from webapps/images to <openHAB 2 conf>/icons/classic
+Copy any custom webviews from webapps tp <openHAB 2 conf>/html
```
BClark09: Why do these last two lines follow a different format to the rest of the chunk?

tutorials/migration.md
rkoshak: I was using one of the other pages as my example. Wish I remember which one because apparently that one needs to be fixed too.

tutorials/migration.md
```diff
@@ -4,96 +4,846 @@ layout: intro
...
 
-- The `configurations` folder has been renamed to `conf`
-- Instead of the global `configurations/openhab.cfg` file, there is now an individual file per add-on in `conf/services`.  The `<key>:` prepended to each configuration property is removed in the new `<key>.cfg` files.
-- You can still put additional jar files in the `addons` folder to install bindings that are not available in the distribution, but please note that removing a jar file from this folder will not uninstall it. To uninstall it, you will have to use the console command bundle:uninstall.
-- The OSGi console commands are now available as "smarthome", not as "openhab" anymore.
-- The webapps folder has been discontinued, but static resources can be placed in `conf/html`.
+## Table of Contents
 
-### User Interfaces
+[1. Preparation](#preparation)
```
rkoshak: I was unaware of this.  I will fix it. More automation is more better.

tutorials/migration.md
rkoshak: That one got away. I thought I had removed all those mistakes. Good catch,.

tutorials/migration.md
rkoshak: I could not get the ``` code fences to work at all. The only way I could get a block of text to show up as code was to use ~~~~ (I had to look that up). When I used ``` it wouldn't format the block as code at all whether I used a language e.g text, java, xml, sh, etc) or not.

tutorials/migration.md
rkoshak: That never should have been an absolute path. Not sure what happened here.

tutorials/migration.md
rkoshak: It didn't work. The only way I could get it into code formatting was to use ~~~~

tutorials/migration.md
rkoshak: Because they are more involved than simply copying everything in those folders over. For the icons one only wants to copy over the custom ones, not all of them.

Similarly there may be things in webviews (e.g. habmin) that one would not want to copy over to the OH 2 install.

tutorials/migration.md
rkoshak: I never was able to get it to work in OH 2 even with the compatibility layer. Perhaps the different path part was a problem. If you can provide some more details I'll make sure it gets added. appropriately.

tutorials/migration.md
rkoshak: Didn't work

tutorials/migration.md
rkoshak: When I first wrote it I used ```java. It did not work for me. It basically treated all the text as normal text. I could only get code formatted text using ~~~~.

tutorials/migration.md
BClark09: Fair enough!

tutorials/migration.md
BClark09: That's strange! I don't think there's anything different to other pages that work,  I'll see if I can compile and get the same result.

tutorials/migration.md
kaikreuzer: I was actually considering to change these package types due to discussions we had.
The idea would be to have a:
- "simple" package, which only includes 2.0 bindings, Paper UI, HABPanel and the new rule engine and which uses simple mode by default.
- "expert" package, which includes 1.x-compat-layer, all UIs, the transformation services that were part of 1.x and which comes with simple mode deactivated.

My idea would be that those "packages" are just samples that declare a default set of addons. Ideally, they should simply adapt the addons.cfg itself, so that the user has a chance to also remove stuff at any time - this is currently an issue with e.g. the demo package as you cannot uninstall bindings from it, unless you switch away from demo.

Does this make sense in general to you? I am not sure yet, how I can achieve this technically, but I think it would make stuff easier - and I hope that the necessary adaptions in this tutorial won't be too big.

tutorials/migration.md
BClark09: For openHAB 1 webapps, dynamic resources go in `<userdata>/webapps/` and the static content goes in `<conf>/html`. For example for a webview from the 1.x weather binding:

```xtend
Webview url="/weather?locationId=home&layout=example&iconset=colorful" height=7
```

- The `layout/` and `images/` folders can go in `<conf>/html/weather/`
- The html page that the webview points to should be located at `<userdata>/webapps/weather-data/layouts/example.html` and this needs to be edited to contain the new location of the static content: e.g. `src="/static/weather/layouts/example.js"`.

Note that the \`\`\`xtend syntax works here, and this is what the ideal markup would look like in the docs.

tutorials/migration.md
```diff
@@ -4,96 +4,846 @@ layout: intro
...
+
+&nbsp;&nbsp;[11.1 Managing the Inbox using PaperUI](#managing-the-inbox-using-paperui)
+
+&nbsp;&nbsp;[11.2 Managing the Inbox Using the Karaf Console](#managing-the-inbox-using-the-karaf-console)
+
+&nbsp;&nbsp;[11.3 Linking Channels to Items](#managing-the-inbox-using-the-karaf-console)
+
+&nbsp;&nbsp;[11.4 Manually Creating Things](#manually-creating-things)
+
+##Preparation
```
BClark09: There needs to be spaces between # and the section title, otherwise jekyll won't render it. Multiple instances of this throughout this PR.

tutorials/migration.md
BClark09: The expert package sounds like the sort of package that migration users would be looking for and I think this tutorial won't need to change too much to use this.

rkoshak: It won't take too much at all. It would mainly let us remove some content (e.g. how to install the compat-layer since it will already be there.

tutorials/migration.md
ThomDietrich: There where two times code fences did not get compiled for me. 
1. when there are spaces behind them
2. Discourse: At some point Discourse stops rendering. I do not recall what the reason was. Something with special characters or Headlines... Anyhow that's not related to Jekyll,  just put it in here.

tutorials/migration.md
ThomDietrich: There are some Markdown litter for that. I use remark with good results

tutorials/migration.md
rkoshak: just so you don't think I'm crazy:

image

tutorials/migration.md
rkoshak: Using ~~~~java seems to do something. I'm not sure if it is right or not. ~~~~xtend does nothing.

rkoshak:
I just pushed some updates which I believe address all of the comments above. Keep them coming.

BClark09:
Yeah that looks better! Only one more comment, and that is the censorship on the myopenhab-account.png, would this look better without quick brushing over, i.e: 
image

Or if you add example text?

rkoshak:
I completely forgot about that image. I intended to cover over it like you did and put account@email.com in that spot. Even if it looks fake it will show them what to expect. Honestly I don't really care if my email ends up there but I wanted to keep everything pretty generic where I could. Good catch. Sometimes you sirens so much time focused on details the big things completely get by one's notice. I'll fix the image after the little one's bath.

BClark09:
This all looks good to me now!  A really well written  tutorial @rkoshak!

tutorials/migration.md
ThomDietrich: Are we talking Discourse or Jekyll/both? Fact is I had this problem once but don't remember the cause. Just that I were able to solve it. I would look but I'm not on a Pc till next weekend

rkoshak: We are talking Jekyll and the problem ended up being I was running 3.0. upgrading to 3.3 made it work.

ThomDietrich:
Hey guys, I would read through the whole article and look into the markdown problem discussed above after 27th. Please decide on your own if you want to wait for that, otherwise tell me and I can merge right away (and still check later). Totally up to you Rich ;)

BClark09:
The markdown problem is resolved now, It was simply an out of date Jekyll.

rkoshak:
I'm ok with it going live as is. Of course I'm biased. :-) If something comes up we can create a new issue to fix it. I've already had some cases where I wish I'd had this updated version to point to on the forum.

kaikreuzer:
Let's go then - million thanks to @rkoshak and looking forward to the next PRs flowing in :-)

PR #143, opened 2016-11-24
adapted demo and migration tutorial to new package types

kaikreuzer:
See https://github.com/openhab/openhab-core/pull/81

### DIFF
--- a/tutorials/migration.md
+++ b/tutorials/migration.md
@@ -203,7 +203,7 @@ An example addons.cfg:
 #   - minimal  : Installation only with dashboard, but no UIs or other addons
 #   - simple   : Setup for using openHAB purely through UIs - you need to expect MANY constraints in functionality!
 #   - standard : Default setup for normal users, best for textual setup
-#   - expert   : Setup for expoert users, especially for people migrating from openHAB 1.x
+#   - expert   : Setup for expert users, especially for people migrating from openHAB 1.x
 #   - demo     : A demo setup which includes UIs, a few bindings, config files etc.
 #
 package = standard
--- a/tutorials/migration.md
+++ b/tutorials/migration.md
@@ -158,13 +158,12 @@ Next open for editing `addons.cfg`. This file contains parameters related to
 add-ons including allowing one to choose which add-ons to install. For now just
 set the following parameters:
 
-- `package = minimal` - minimal or standard are good choices here. Standard
-will include the standard UIs, though you can choose those individually later
-- `remote = true` - required for the installation of legacy 1.x and 
-experimental add-ons
+- `package = expert` - expert is good choices here. It will include the standard
+UIs, all transformation services and the 1.x compatibility layer,
+though you can choose your UIs and add-ons individually later.
+- `remote = true` - required for the installation of legacy 1.x add-ons
 - `legacy = true` - required to install the 1.9 version of an add-on for which 
 there is a 2.0 native version
-- `experimental = true` - required to install experimental add-ons
 - `binding = ` - a comma separated list of the bindings from openHAB 1.x that
 you use. If you are in doubt of the name of a binding, look in openhab.cfg for
 that binding's configurations. The first part of the tag in openhab.cfg will be
@@ -202,10 +201,12 @@ An example addons.cfg:
 # The base installation package of this openHAB instance (default is "standard")
 # Valid options:
 #   - minimal  : Installation only with dashboard, but no UIs or other addons
-#   - standard : Typical installation with all standards UIs
+#   - simple   : Setup for using openHAB purely through UIs - you need to expect MANY constraints in functionality!
+#   - standard : Default setup for normal users, best for textual setup
+#   - expert   : Setup for expoert users, especially for people migrating from openHAB 1.x
 #   - demo     : A demo setup which includes UIs, a few bindings, config files etc.
 #
-#package = standard
+package = standard
 
 # Access Remote Add-on Repositories
 # Defines whether the remote openHAB add-on repository should be used for browsing and installing add-ons.
@@ -214,16 +215,10 @@ An example addons.cfg:
 #
 remote = true
 
-# Include legacy 1.x bindings. If set to true, it also allows the installation of 1.x bindings for which there is
+# Include legacy 1.x bindings. If set to true, it also allows the installation of 1.x bindings for which there is 
 # already a 2.x version available (requires remote repo access, see above). (default is false)
 #
 legacy = true
-
-# Include experimental add-ons. If set to true, it also allows the installation of extensions that are not (yet) part
-# of the official distribution and need further testing (requires remote repo access, see above). (default is false)
-#
-experimental = true
-
 
 # A comma-separated list of bindings to install (e.g. "sonos,knx,zwave")
 binding = zwave,astro1,http,mqtt,nest,networkhealth,ntp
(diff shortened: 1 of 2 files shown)


tutorials/migration.md
```diff
@@ -158,13 +158,12 @@ Next open for editing `addons.cfg`. This file contains parameters related to
 add-ons including allowing one to choose which add-ons to install. For now just
 set the following parameters:
 
-- `package = minimal` - minimal or standard are good choices here. Standard
-will include the standard UIs, though you can choose those individually later
-- `remote = true` - required for the installation of legacy 1.x and 
-experimental add-ons
+- `package = expert` - expert is good choices here. It will include the standard
```
BClark09: Expert is a good choice here

kaikreuzer:
Thanks @BClark09, fixed it.

ThomDietrich:
@rkoshak I was waiting on a comment from you regarding the migration tutorial. Lgtm.

rkoshak:
Sorry, I've been struggling to keep up the last couple of weeks and didn't even see this.

kaikreuzer:
@ThomDietrich & @rkoshak: I'd actually like to discuss the package options with you as I am not completely sure yet, whether these should be the final ones - I somehow feel that 5 are too many and 4 might be a better number.

But let me first briefly explain what this packaging at all does now: This setting is merely used at startup now to define an initial userdata/etc/org.openhab.addons.cfg file, if it does not exist (plus it sets simple linking mode for the "simple" package and it adds the demo files for the "demo" package). The addons file defines what should be installed for a start, but as Paper UI as well as conf/services/addons.cfg settings modify this, this is not "lasting", but the user is free to customize his setup in any way he wants.

Having to define this initial package in conf/services/addons.cfg before a first launch is at the moment only kept for backward compatibility. My plan would be to put the [package line in addons.cfg](https://github.com/openhab/openhab-distro/blob/master/distributions/distribution-resources/src/main/resources/conf/services/addons.cfg#L9) also in comments. If you do that for a test on a freshly unzipped distro, you will see a new intro page instead of the dashboard:

screen shot 2016-12-05 at 14 21 36

This page might explain you, why I prefer 4 instead of 5 options ;-) It is just an initial draft, it would still have to be polished with texts & images.
Clicking one of these options will simply set the package in the configuration, the very same way as it is done through the line in addons.cfg. What do you think of this mechanism? I think it is much more natural then having to go to a rather hidden file conf/services/addon.cfg before a first launch. My idea would also be to enhance this page in future to get hold of some other information, like the [geo location of the runtime](https://github.com/eclipse/smarthome/issues/1694) etc.
If you agree to pursue this way, we should probably also add dedicated pages on docs.openhab.org, which describe the package and its content - the setup page could then refer to these pages directly.

rkoshak:
>  I somehow feel that 5 are too many and 4 might be a better number.

I agree. 5 seems too many. 4 seems right independent of how it looks on the page ;-). What isn't clear to me is what the difference is between Expert and Minimal. I would expect Expert to be just the core runtime also and I wouldn't expect anyone but an expert to use Minimal.

>  I think it is much more natural then having to go to a rather hidden file conf/services/addon.cfg before a first launch.

I agree. This is a great way to bootstrap the system for new users, particularly as we move more and more towards the UIs being fully capable for almost all users. And this can and will feed into a Getting Started tutorial.

In addition to getting geolocation, I can see this initial page being a great place to set just a few overall runtime settings like localization/language (inferred from geolocation perhaps?), and the initial user and password when authentication makes it back into OH. 

> If you agree to pursue this way, we should probably also add dedicated pages on docs.openhab.org, which describe the package and its content - the setup page could then refer to these pages directly.

Indeed and and these will also feature in many of the other tutorials so having the separate pages to point to makes a lot of sense.

kaikreuzer:
>  I wouldn't expect anyone but an expert to use Minimal.

Well, "Expert" at the moment is the setup for OH1 users, i.e. it brings the compat layer and all the transformation services (which were part of the OH1 runtime and now became add-ons).
 "Minimal" just means that it does not install ANY add-on from the start and leaves it to the user. Right, this is something only experts will do. Maybe not worth to offer it as an option on the Setup page, but I wanted to provide a way to NOT start downloading add-ons that one in the end might not want to use at all for a custom setup.

BClark09:
If they're looking at that page then they're already looking at a minimal setup right? Why not change the "Please choose your installation type:" to "Please choose your installation type, or to skip this process click [here](#):

Then offer the 4 other options?

rkoshak:
> Well, "Expert" at the moment is the setup for OH1 users, i.e. it brings the compat layer and all the transformation services (which were part of the OH1 runtime and now became add-ons).

That makes a whole lot of sense. I forgot about the transformations.

Hmmmm, that pushes me back to five. Grrr. Having a configuration like this will make it much easier for OH 1.x migraters. 

I like @BClark09's suggestion. "Skip" becomes your minimal which can be a smaller button under the four options which will maintain the more even look while also giving the five options.

kaikreuzer:
> I like @BClark09's suggestion. "Skip" becomes your minimal which can be a smaller button under the four options which will maintain the more even look while also giving the five options.

Ok, I like that as well. I can adapt the setup page accordingly.
But in the end it means we still have the 5 different packages, so the description in the addons.cfg in this PR stays unchanged, right?

rkoshak:
> so the description in the addons.cfg in this PR stays unchanged, right?

I think so. Even if the GUI has the skip, the cfg file still needs to keep the minimal option.

kaikreuzer:
Here's how it now looks like:

screen shot 2016-12-08 at 17 38 52

I'd like to add a link to the docs for some details on the packaging types. 
@ThomDietrich Any suggestion where this should point to? I can provide the content for such a page if you give me a space for it :-)

rkoshak:
Nice! I like the background images. It really sets them apart. 

Does it make sense to put the docs under Installation? I could easily see these fitting in on the Overview page as it is something almost everyone will have to deal with no matter what approach they choose to install.

ThomDietrich:
Nice! I like the idea, the design and the descriptions. The idea of @BClark09 comes out great in the new screenshot. This seems very intuitive and in line with how other systems would do this kind of thing :+1: 

One small things: "there are 4 pre-defined" -> "there are four pre-defined"?

Regarding the article: I would not put this inside "Installation". Installation addresses everything AROUND openHAB (OS, Hardware, Designer, ...). I would suggest to put the package details inside an article under Configuration ([screenshot](https://cloud.githubusercontent.com/assets/2870104/20943083/65305658-bbfe-11e6-9642-3076a16d01f0.png)) and (to make the link as stable as possible) as it's own document. How about `configuration/packages.md` which is linked as "Initial Setup" above the PaperUI entries in the screenshot?

kaikreuzer:
> One small things: "there are 4 pre-defined" -> "there are four pre-defined"?

Addressed with https://github.com/openhab/openhab-core/pull/91

> How about configuration/packages.md which is linked as "Initial Setup" above the PaperUI entries (which are not final) in the screenshot?

Sounds good to me!

ThomDietrich:
@kaikreuzer Done. https://github.com/openhab/openhab-docs/pull/148/commits/a64814b135dee2716f73063df5c3d4f89a0a62c0

How do you want to continue? Send me the content now or create your own PR after #148 is merged...

kaikreuzer:
Won't be able to do it this weekend, so better after #148 has been merged.

PR #154, opened 2016-12-15
Small improvements to the add-ons the menu

kaikreuzer:
- removed rules from 3rd party integration
- changed menu text of myopenhab


ThomDietrich:
LGTM, I agree with both. I'm actually not sure why the rules are under 3rd party.

ThomDietrich:
Just realized that I maybe was a bit quick here as you maybe wanted to add further changes or merge yourself. Will be more careful next time...

kaikreuzer:
All fine, once a PR is pushed out, it should be considered to be ready to be merged - that was the case here :-)

PR #155, opened 2016-12-15
Remove www from myopenhab.org link

cniweb:
SSL secret warning!
Signed-off-by: Christian Häussler <c-n-i@web.de> (github: @cniweb)


kaikreuzer:
Thanks for noticing!
Note that this is a generated file that you try to edit here.
I have fixed the original source here: https://github.com/openhab/openhab2-addons/pull/1576/files

cniweb:
Ahh, OK

PR #158, opened 2016-12-17
Add documentation for item tagging

pfink:
I had to figure out on my own where exactly to put tags, so it would be good to have this documented as well...

### DIFF
--- a/configuration/items.md
+++ b/configuration/items.md
@@ -32,7 +32,7 @@ Doing so you will have full IDE support like syntax checking, context assist etc
 Items are defined in the following syntax:
 
 ```xtend
-itemtype itemname ["labeltext"] [<iconname>] [(group1, group2, ...)] [{bindingconfig}]
+itemtype itemname ["labeltext"] [<iconname>] [(group1, group2, ...)] [["tag1", "tag2", ...]] [{bindingconfig}]
 ```
 
 Note: Parts in square brackets ([]) are optional.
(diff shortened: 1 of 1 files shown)


ThomDietrich:
Thanks! Would you also add a section introducing and describing tags?

pfink:
Unfortunately, I also don't know more stuff than I already included to this PR. I thought  directly adding that what I know is at least better than just opening issues and let others do the complete work...

ThomDietrich:
Absolutely right. Asking doesn't hurt. Thank you for your contribution!
@kaikreuzer: Anyone else or a document/discussion somewhere capable to shed some light on this topic?

kaikreuzer:
Tagging is indeed still in its infancy. The idea of it and the discussion around it can be found at https://github.com/eclipse/smarthome/issues/1093.

I think what could be moved to the documentation right now is the current list of valid tags, which is documented in [the HomeKit binding](http://docs.openhab.org/addons/io/homekit/readme.html#item-configuration).

pfink:
Maybe we should start with this PR and open an issue for the rest?

configuration/items.md
```diff
@@ -49,7 +49,8 @@ The example above defines an item:
 * with name `Sensor_Temperature`
```
itheiss: As the item in the example has changed from Sensor_Temperature to Temperature shouldn't this be "with name `Temperature`"?

ThomDietrich: Most certainly. In fact now that I look at this again, "Temperature" is a horrible item name. In my home this would be called LR_Temperature. Would you come up with a PR for LivingRoom_Temperature or whatever you think is appropriate?

itheiss: Ok. Done. Please review.

Issue #53, opened 2016-08-03
sitemap article, missing details


ThomDietrich:
Hello busy bees,
the sitemaps article is almost done. I've created pull request https://github.com/openhab/openhab-docs/pull/54 to merge the article in it's first version. I would like to point out a few open questions:
- these three [sitemap elements](https://github.com/eclipse/smarthome/tree/master/extensions/ui/org.eclipse.smarthome.ui.basic/src/main/java/org/eclipse/smarthome/ui/basic/internal/render) have been left out, can somebody comment on them? List (NPE in BasicUI), Mapview, Page (I didn't try them, found out about them only now.)
- A few links in the document are not final, don't worry, I will soon create the needed articles, first as dummys though.
- I am personally not using the Charts element (InfluxDB+Grafana+ImageRenderer ❣) I would like to ask one of you guys to [update the article](https://github.com/ThomDietrich/openhab-docs/blob/patch-5/features/sitemap.md#element-type-chart) (new PR or as comment here) on this one element and to provide a screenshot. Beware, that all screenshots were created with the sitemap definition embedded at the end of the article as a comment. Be sure to update it. Thanks ;)
- The [slider](https://github.com/ThomDietrich/openhab-docs/blob/patch-5/features/sitemap.md#element-type-slider) and the [colorpicker](https://github.com/ThomDietrich/openhab-docs/blob/patch-5/features/sitemap.md#element-type-colorpicker) are two other elements I never used. Thanks to anyone adding missing information (see comments in the code).
- A native English speaker is invited to touch up my text. If you find real problems, please lecture me about them ;P 
##

kaikreuzer:
> these three sitemap elements have been left out

- Page: This is not a valid widget (see the "real" list [here](https://github.com/eclipse/smarthome/blob/master/bundles/model/org.eclipse.smarthome.model.sitemap/src/org/eclipse/smarthome/model/Sitemap.xtext#L14))
- MapView: shows an OpenStreetMap map for a Location item. Try [this](https://github.com/eclipse/smarthome/blob/master/distribution/smarthome/conf/sitemaps/demo.sitemap#L9).
- List: Good question, I think no UI supports this and nobody uses it. Just leave it away for now.
- The widget that is missing is "Default": This let's the system choose the widget type for the UI automatically based on the item type (see [here](https://github.com/eclipse/smarthome/pull/1535)).

[...]

PR #159, opened 2016-12-17
Add `Default` element type to sitemap documentation

pfink:
see eclipse/smarthome#1535

### DIFF
--- a/configuration/sitemaps.md
+++ b/configuration/sitemaps.md
@@ -98,6 +98,7 @@ The following element types can be used in a sitemap definition file.
 | Element                                                                     | Description                                                                                             |
 |-------------------------------------------|-----------------------------------------------------------|
 | [Frame](#element-type-frame)              | Area containing various other sitemap elements.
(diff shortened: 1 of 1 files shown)


ThomDietrich:
Hey Patrick!
Thanks for your contribution! Adding default was one of the open tasks discussed in https://github.com/openhab/openhab-docs/issues/53. Would you like to tackle the other once as well? If not, just let me know and we can go on with the current PR content ;)

pfink:
Unfortunately, I don't know the other ones. I always work this way: I'm figuring out the things I currently need (which is mostly of course just a small subset from what is available), and if parts of it were not documented, I directly open PRs with the content I figured out to prevent that others run into the same issue. In my opinion that's a good compromise between being an overall maintainer which is quite time consuming or, the other extreme, not contributing anything at all.

So you'll may see more Mini-PRs by me in the future that'll improve stuff bit by bit.

ThomDietrich:
Thanks @pfink for both your PRs!

PR #162, opened 2016-12-19
Added tested device DS216+II

heinbm:


ThomDietrich:
@cniweb do you think it's a good idea to reference release versions in this document? Will certain devices only work with certain versions?

cniweb:
@ThomDietrich Yes, certain models only run with a specific release. But I am already planning a version that will be compatible for all models.

ThomDietrich:
Okay then.
Thanks @heinbm!

PR #173, opened 2016-12-30
also correctly sort bindings in the menu

kaikreuzer:
The order is already correct on the overview page, but the menu still lists the ESH bindings first and then the OH2 bindings.

### DIFF
--- a/_includes/user-menu.html
+++ b/_includes/user-menu.html
@@ -60,7 +60,10 @@
         <ul>
           <li><a href="{{docu}}/addons/bindings.html">Overview</a></li>
           <hr />
-          {% for binding in site.data.bindings %}
+          {% assign bindings = "" | split: "|" %}
+          {% for binding in site.data.bindings %}{% assign bindings = bindings | push: binding %}{% endfor %}
+          {% assign sorted_bindings = bindings | sort: "id" %}
+          {% for binding in sorted_bindings %}
           <li><a href="{{docu}}/addons/bindings/{{ binding.id }}/readme.html">{{ binding.label }}</a></li>
           {% endfor %}
         </ul>


ThomDietrich:
Hey Kai, I'm on vacation and away for a few days (besides other things installing openhab in my brother's house :D). I'd look at this PR on Wednesday/Thursday, if you can wait this long.

kaikreuzer:
waited ;-)

ThomDietrich:
:sweat_smile: okay I've looked at the result. I see the improvement but would have hoped to get rid of this problem:

* vitotronic
* Belkin Wemo
* YahooWeather

Any reason why you decided to sort be id instead of label? I've just done that and everything looks fine. If you agree, please also change bindings.md ;)

kaikreuzer:
> Any reason why you decided to sort be id instead of label?

Well, in this PR I did it exactly the same way as the binding list is done, so I think it is fully in sync and there is no reason not to merge it :-)

Whether we want to change both places to use a different ordering can be discussed. It is not completely unlogical though as it is: People will look for a "Hue" and a "WeMo" binding and will find it at the right place. It is a bit like sorting "The Police" in your albums under "P".

ThomDietrich:
I didn't pay attention to sorting when touching the bindings. I would definitely have sorted by label, everything else just seems random. "The Police" is a special case...

My preferred solution would be both sorted by label.

ThomDietrich:
@kaikreuzer

kaikreuzer:
Tried to change it to sorting by label, but I failed to apply a "toLower" operation - without it, "digitalSTROM" is sorted at the end.

So I still vote for merging this and anybody else more fluent in Jekyll/Liquid can at anytime come up with a new PR.

ThomDietrich:
That's not ideal indeed! Okay, let's go. Thanks

PR #148, opened 2016-12-04
Big User Manual structure revision

ThomDietrich:
## Update, Ready for Review

The open points below are not structure related and can be attended to by their own PRs.
I'll create individual issues for all of them.

---

- [x] Add hr rulers to menu structure
- [x] Rearrange Installation and Setup order
- [x] Rearrange Features articles to Configuration
- [x] Add redirects for all old article locations
- [x] Add placeholders for not yet existing articles
- [x] Insert a link to wiki or whatever suited in new articles (till filled with content)
- [ ] Write Configuration/Overview article with the different ways of configuration (text vs. database, rules files exclusive, sitemap vs. HABPanel, ...)
- [ ] Write PaperUI configuration article
- [ ] Update openHABian article
- [x] Update RaspberryPi article with openHABian details
- [ ] Add SSH disabled info to the RPi article
- [ ] Help: How to post a problem, How to contribute a solution
- [x] Rename menu entry openHAB Designer to Smarthome Designer
- [x] Better Arrangement of "System Integration"
- [x] IFTTT as another "System"
- [ ] ...

-----

### Motivation

I tried to follow these ideas:

**Reducing the menu to the logical parts:**
* Preamble (Introduction, Concepts)
* Prerequisites (Installation and Setup)
* Howto (Configuration and Usage)
* Further possibilities (Addons, Administration)
* Outlook (Help, FAQ, Troubleshooting)

Horizontal rules were introduced as **visual separators** between different topics, without the need to name them or have a collapsing menu for them.

Quite a few articles shifted in their position. All **previous locations** are [set up as redirects](https://raw.githubusercontent.com/openhab/openhab-docs/0c4daa4052341667a057fd740b1de48c4b839f3b/features/items.md). Old links in e.g. the community will not be a problem.

While building the menu, I took care to have a **menu entry for every aspect** of a topic. A [placeholder article](https://github.com/openhab/openhab-docs/blob/0c4daa4052341667a057fd740b1de48c4b839f3b/configuration/transform.md) exists for every article not created yet. The content of these articles is not clear yet. In one place I decided to have two menu entries pointing at the same article, there is room for improvement.

Every article not yet finished (especially placeholders) include a **Contributions Wanted** note with a link to the issue tracker.

------

### Local Environment

If you want to see and test the new arrangement, the probably quickest way is to clone the Repo and execute `vagrant up`. See: https://github.com/ThomDietrich/openhab-docs#alternative-vagrant-vm

PR #178, opened 2017-01-08
Added option to generate a sitemap.xml file for Google

MHerbst:

### DIFF
--- a/404.md
+++ b/404.md
@@ -1,6 +1,8 @@
 ---
 layout: intro
 title: Page not found
+permalink: /404.html
+
 ---
 
 {% include base.html %}
--- a/_config.yml
+++ b/_config.yml
@@ -1,11 +1,10 @@
 # Site settings
-title: openHAB 2
+title: openHAB 2 
 description: Empowering the Smart Home
 baseurl: "" # the subpath of your site, e.g. /blog/
 url: "http://docs.openhab.org" # the base hostname & protocol for your site
 twitter_username: openhab
 github_username:  openhab
-permalink: /404.html
 # Build settings
 markdown: kramdown
 exclude: ["CNAME", "pom.xml", "README.md", "CONTRIBUTING.md"]
--- a/_config.yml
+++ b/_config.yml
@@ -9,3 +9,7 @@ permalink: /404.html
 # Build settings
 markdown: kramdown
 exclude: ["CNAME", "pom.xml", "README.md", "CONTRIBUTING.md"]
+# Additional gems for sitemap generation
+gems:
+  - jekyll-sitemap
+  
--- a/_includes/header.html
+++ b/_includes/header.html
@@ -16,6 +16,7 @@
 
   <!-- Font -->
   <link href='https://fonts.googleapis.com/css?family=Lato:300,400,700' rel='stylesheet' type='text/css' />
+  <link rel="canonical" href="{{ site.url }}{{ page.url }}" />
 
   <script>
   (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){


ThomDietrich:
For reference: https://help.github.com/articles/sitemaps-for-github-pages

Is the result as desired?

MHerbst:
It seems that it will not really work with openhab-docs :-(. For some reason I did not notic it my local tests because it look like the sitemap of the foundation's site.
The URLs generated into the sitemap.xml do not contain extensions. 

After some tests I found out that this is caused by the *permalink* option in `_config.yml`:
```
permalink: /404.html
```
This options looked a bit strange to me. I have removed this line and then sitemap and generated pages were OK.

BTW: this option was entered in the wrong file. It must be set in the frontmatter of 404.md (see: https://help.github.com/articles/creating-a-custom-404-page-for-your-github-pages-site/)

I will change and commit it.

Anyway I would say that there is a bug in the sitemap plugin.

BTW: with the foundation's site it is possible to access the pages with and without extension (that's why the generated sitemap, which looks similar, works here). 
As an example, you can access the contact page with URL http://www.openhabfoundation.org/contact and also with URL http://www.openhabfoundation.org/contact.html

@kaikreuzer Do you know why this works on the foundation's site?  The site mixes both URL formats and all links are working .... Is this configured on Cloudflare?

ThomDietrich:
(only on phone currently) links with and without extension do also work at docs. 
One example I know for sure: http://docs.openhab.org/tutorials/migration

MHerbst:
You are right. Probably I tried it only with the local Jekyll server and there it does no work.

MHerbst:
I really don't know what I tested yesterday evening, but now I tried it again. It is very interesting but also the Jekyll Server supports the links with and without extension. So the local test environment behaves similar to the deployed website. Also the links in the sitemap are OK. I have tried several of them and all were OK (I will leave the final test to Google ...).

I saw that links used on the website sometimes are using the extension and sometimes not. E.g. in the main menu we have these variants:

- http://docs.openhab.org/tutorials
- http://docs.openhab.org/introduction.html
- http://docs.openhab.org/developers

All three would work with and without extension. 

But there is another point, that is more important. All pages can be addressed by at least two different URL.s. For some page there are even three valid URLs:

- http://docs.openhab.org/tutorials/
- http://docs.openhab.org/tutorials/index.html
- http://docs.openhab.org/tutorials/index

I would not even think about changing the permalink formats, because this would lead to lots of 404 errors due to the many links to pages on docs.openhab.org.

But Google does not really like these "duplicates". This will be solved by the sitemap.xml. But to be sure we should also generate a canonical link into the header of each page that contains the preferred URL. I can add this tomorrow to _includes_header.html. 

BTW: I saw that each page contains some Twitter Javascript at the bottom. I think we should remove it.

ThomDietrich:
Hello Martin,

> in the main menu we have these variants

The main menu could just be the only menu I didn't touch in #148.

> I would not even think about changing the permalink formats, because this would lead to lots of 404 errors due to the many links to pages on docs.openhab.org.
> But Google does not really like these "duplicates".

Which links are we talking about? I would not expect too many problems from that. In #148 I've addressed a similiar problem this way: https://raw.githubusercontent.com/openhab/openhab-docs/gh-pages/features/items.md -> Could be a solution

> But to be sure we should also generate a canonical link into the header of each page that contains the preferred URL. I can add this tomorrow to _includes_header.html.

Sounds good ;)

> BTW: I saw that each page contains some Twitter Javascript at the bottom. I think we should remove it.

I think so too.

MHerbst:
Hi Thomas,

> Which links are we talking about? I would not expect too many problems from that. In #148 I've addressed a similiar problem this way: https://raw.githubusercontent.com/openhab/openhab-docs/gh-pages/features/items.md -> Could be a solution

I was talking about the pages that could be accessed with and without .html extension. The only thing that we need to avoid problems with google is to add a canonical link to each page. But in general it looks better it the site uses the same link format (without .html) in the menus and the internal looks.
Last week while modifying the Foundation website I learned that the possibility to access pages with our without extension seems to be a Cloudflare configuration. Kai and I also had to find out that some links in the deployed page behaved different than in the Jekyll test environment.

Since monday evening the Foundation website works perfectly with https. Next week I will check again my modifications in this PR. We will also have to add a privacy policy page with the possibility to opt-out from Google Analytics (we can copy it from the foundations's site) and an imprint (it's in work).

ThomDietrich:
Hi Martin,
did you want to add more to this PR or can it be merged?

MHerbst:
Hi Thomas,
I would like to add the canonical links. Will be probably ready this afternoon.

ThomDietrich:
Great. Okay I'll wait then.

MHerbst:
@ThomDietrich Ready to merge now :-)

PR #186, opened 2017-01-12
Add missing openHAB 1.x binding information

wborn:
This PR should add the missing binding information in the table on: http://docs.openhab.org/addons/bindings.html#compatible-1x-bindings

### DIFF
--- a/_data/oh1addons_infos.yml
+++ b/_data/oh1addons_infos.yml
@@ -53,6 +53,18 @@
   description: Monitor and control Anel NET-PwrCtrl power sockets and relays
   wiki_url: https://github.com/openhab/openhab/wiki/Anel-Binding
 
+- label: Bticino Binding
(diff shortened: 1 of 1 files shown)


ThomDietrich:
Great addition. Just out of interest: How did you compile this list? Are you certain it's now complete?

wborn:
1. I checked the generated [HTML](http://docs.openhab.org/addons/bindings.html#compatible-1x-bindings) for what was missing
2. I got the label from `oh1addons.csv` (which is generated)
3. Then I searched for the right [wiki](https://github.com/openhab/openhab/wiki) and a proper description from the Wiki (and sometimes had to look at the code if the Wiki was not clear enough)

So still a lot of manual labor! :-) It should thus be up to date with the bindings of the generated content in `oh1addons.csv`.

ThomDietrich:
Thank you!

PR #210, opened 2017-01-16
Fix rendering on openhab docs site

Jamstah:
This table doesn't render correctly on the openhab docs site (http://docs.openhab.org/addons/bindings/ipp/readme.html), my best guess is because there is no newline after the table.


ThomDietrich:
That is correct :) You will need to correct this readme over at https://github.com/openhab/openhab2-addons/blob/master/addons/binding/org.openhab.binding.ipp/README.md though. Thanks!

Jamstah:
I did wonder if it was an automated build. Thanks.

Issue #150, opened 2016-12-06
Add Search function


ThomDietrich:
https://github.com/openhab/openhab-docs/pull/148#issuecomment-264916793

Plugins can not be used on GitHub Pages.
One option: http://digitaldrummerj.me/blogging-on-github-part-7-adding-a-custom-google-search/

cniweb:
I think a search function is useful. But we must have a google account for use the google search engine. Have we one?

MHerbst:
We have an account for Google Search and I have already created a search engine. You can test it with this URL: https://cse.google.com/cse/publicurl?cx=003707250976775348452:h9hxadbatfq

@ThomDietrich I can add it to the website but where do we want to put the search field? One option would be in the menu at the right with an expandable edit field.
Or in the left sidebar on top of the menu items. I would configure the search function so show the results as  overlay.

ThomDietrich:
Sounds fantastic. I was not able to do a search (no progress after an empty "Loading" page) but I'm sure that's just a small problem.
I like both your options. It would feel more natural to have it in the header, on the top right. Is the horizontal space sufficient for it? Looking forward to that PR! 🎉

MHerbst:
>  It would feel more natural to have it in the header, on the top right. Is the horizontal space sufficient for it?
We can try it. On my website I am using an "expandable" input box. It gets wider when you click into it. But I am not sure whether this is possible with the Google search. We will see :-)

MHerbst:
I have just created the PR for Google Search. Probably the look & feel needs some tweaking. I especially don't like that I had to put the magnifying glass symbol in the menu into a separate li-tag to avoid a line break. Please feel free to improve the styling :-)

kubawolanin:
Should this issue be closed? BTW. I've added small fix for the placeholder in #257

Confectrician:
> Should this issue be closed?

Bump.
Implemented a long time ago.

PR #222, opened 2017-01-17
Added Google Search

MHerbst:
Solves #150

### DIFF
--- a/_includes/header.html
+++ b/_includes/header.html
@@ -29,6 +29,7 @@
 
 </head>
 
+
 <body class="{{ include.bodyClass }}">
   <div id="header" class="navbar-fixed">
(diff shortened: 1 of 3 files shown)


ThomDietrich:
@MHerbst the search function didn't work for me at first. The reason was Ghostery in my browser, blocking [Google AJAX Search API](https://apps.ghostery.com/en/apps/google_ajax_search_api).

Do you now of a way to show a warning in that case?

ThomDietrich:
Besides that, is the design as you anticipated it?

grafik

MHerbst:
Yes that's how it looks on my machine. Will try to optimize it. 
Regarding Ghostery I will try to find out whether it is possible to detect it.

PR #224, opened 2017-01-17
add more information to the Device Tagging section

dsuguy2007:
Add the info that the device tag goes after the group if a group is specified.


ThomDietrich:
Hello @dsuguy2007 Thank you for you contribution. Sadly these articles are just copied from here: https://github.com/openhab/openhab2-addons/tree/master/addons/io/org.openhab.io.hueemulation

Please open your PR there. Regarding the content: Are you sure that this is the only allowed constellation? In this case you should maybe show a full example, including a binding channel.

Thanks and best regards

Issue #233, opened 2017-01-19
Yahoo Binding Example in Migration Tutorial is no longer correct


rkoshak:
Apparently the Yahoo binding channel ID syntax has changed slightly since the tutorial was written. See: 

https://community.openhab.org/t/items-syntax/20672

PR #236, opened 2017-01-19
Changed Yahoo binding example to match README

rkoshak:
Updated the Yahoo Weather Binding example to match the one in the README. In particular removing the reference to "unit" (is that supported any longer?), removing the quotes around the woeid in the thing config, and replacing the "#" with ":" in the Item definition.

Corrects #233 

Signed-off-by: Richard Koshak rlkoshak@gmail.com (github: rkoshak)

### DIFF
--- a/tutorials/migration.md
+++ b/tutorials/migration.md
@@ -599,7 +599,7 @@ a .things file (located in conf/things) with the line:
 
 
 ```java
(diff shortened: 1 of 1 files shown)


ThomDietrich:
Thanks!

Issue #190, opened 2017-01-15
1.x Bindings note


ThomDietrich:
1.x Bindings are differently used than ESH/OH2 bindings. The main differences are
1. they do not know the concept of things and channels
2. items using them need to be set up in configuration files
3. binding configuration https://community.openhab.org/t/mqtt-openhab2-configuration/20283/2

These three rules plus whatever else is needed should be mentioned here: http://docs.openhab.org/addons/bindings.html#legacy-1x-bindings

From https://github.com/openhab/openhab-docs/issues/167:
> we should change [this](http://docs.openhab.org/addons/bindings.html#legacy-1x-bindings) title anyhow to just call it "1.x Bindings" and in [here](http://docs.openhab.org/addons/bindings.html#further-1x-bindings) we could add a new list of all legacy 1.x bindings.

kaikreuzer:
@ThomDietrich Beware that you use the term "legacy binding" differently than it is done in the Paper UI (resp. the openhab-addons-legacy.kar).

There, it is defined as " 1.x bindings for which there is already a 2.x version available", i.e. not all 1.x bindings are referred to as legacy, but bindings for which there isn't any 2.x equivalent yet, are considered as regular add-ons.

ThomDietrich:
Oh, I did not know that 😅  thanks!

ThomDietrich:
Example case: https://community.openhab.org/t/itead-switches-and-sockets-cheap-esp8266-wifi-mqtt-hardware/15024/49

sihui62:
https://community.openhab.org/t/itead-switches-and-sockets-cheap-esp8266-wifi-mqtt-hardware/15024/53?u=sihui

sihui62:
PR [#247](https://github.com/openhab/openhab-docs/pull/247) created

ThomDietrich:
Might still be needed: #275

PR #247, opened 2017-01-22
added instructions for splitting up openhab.cfg

sihui62:
Signed-off-by: Siegfried Huismann <sihui@gmx.de>


addons/bindings.md
```diff
@@ -27,13 +27,29 @@ The following bindings can be installed and configured directly from within Pape
...
+## 1.x Bindings
 
-Bindings developed for the first version of openHAB can also be used with openHAB 2.
-Please be aware, that these bindings function and behave differently.
+Bindings developed for the first version of openHAB can also be used with openHAB 2. If there is already a 2.x binding version available, those are referred to as "Legacy 1.x Bindings", otherwise as regular add-ons.
+Please be aware, that these bindings function and behave differently, f.e. there is no autodiscovery function for openHAB2 things and openHAB2 channels are not available.
 For beginners it is recommended to resort to native bindings if possible.
 
-To be able to see and install 1.x bindings needs to be activated in the system settings.
+To be able to see and install 1.x bindings one has to activate "Access Remote Repository" and "Include Legacy 1.x Bindings" in openHAB2 system settings.
```
watou: I don't think this is true.

addons/bindings.md
```diff
@@ -27,13 +27,29 @@ The following bindings can be installed and configured directly from within Pape
...
+
+When using config examples for the openhab.cfg file from the [openHAB 1 wiki](https://github.com/openhab/openhab1-addons/wiki), those need to be splitted into individual binding config files for openHAB2:
```
watou: The past tense of the verb split is split.

addons/bindings.md
```diff
@@ -27,13 +27,29 @@ The following bindings can be installed and configured directly from within Pape
...
+
+After installing a Legacy or regular 1.x binding an individual `<binding_name>.cfg` file is created in the `/services` folder. If this file is not created automatically it needs to be added manually.
+The naming convention has to be: `<binding_name>.cfg` without the trailing "1" in the binding name grabbed from "PaperUI -> Add-Ons -> Bindings", f.e. the binding name from PaperUI for the MQTT binding is "binding-mqtt1-1.9.x", the config file you have to create is "mqtt.cfg".
```
watou: The abbreviation "f.e." is not recognized by English speakers.  If an abbreviation is needed, "e.g." can be used or preferably "for example" spelled out will be understood.

addons/bindings.md
sihui62: Thanks for the quick reply. My knowledge about this statement is from here:
https://community.openhab.org/t/oh2-upgrade-of-today/15560/12?u=sihui
remote=true and legacy=true

addons/bindings.md
sihui62: fixed

addons/bindings.md
sihui62: fixed

addons/bindings.md
watou: But that message says something different:

> On order to add legacy 1.x bindings through configuration files....

sihui62: Okay ... but it does not make any difference if those two config options are set through PaperUI or addons.cfg.
Anyway, do you know which option has to be set, if anyß
Thx.

sihui62: Okay, found the source:

https://community.openhab.org/t/a-single-distro-from-now-on/19903?u=sihui

> For the ones that loved the offline distro (like myself): There is an easy way to get it back! Just additionally download the file openhab-addons-2.0.0-SNAPSHOT.kar and place it in your local addons/ folder - this will immediately make all add-ons available offline and you can safely deactivate the online repo access (in addons.cfg or through the Paper UI). You can even do the same now with the legacy add-ons (which were only available online before) - simply download the openhab-addons-legacy-2.0.0-SNAPSHOT.kar and put it into the addons/ folder.

For me that means, if you don't download the *.kar, you have to enable both options.

watou: Is it not that the legacy switch is only needed for legacy bindings and not for access to any 1.x binding?  That was my main point, but the .kar point applies as well.

sihui62: It looks like there is some confusion lately about the terms "legacy" and "regular" (I'm confused now too):
https://github.com/openhab/openhab-docs/issues/190
I will leave the changes as is at the moment, I'm pretty sure @ThomDietrich will take a look before merging anything ...

watou: https://github.com/openhab/openhab-docs/issues/190#issuecomment-272685141 seems clear to me, but I'm not in a hurry for you to resolve it now as long as incorrect info isn't merged and forgotten. :smile:

ThomDietrich:
@sihui62 thanks for helping out! All details we talked about are mentioned 👍 I've broken down the lines (better with git) and done a little rephrasing here and there. I hope that's okay with you.
@watou I believe the Legacy topic is addressed correctly now. I've not mentioned the kar package.

addons/bindings.md
```diff
@@ -29,27 +29,34 @@ The following bindings can be installed and configured directly from within Pape
...
+Binding configuration was done in the global `openhab.cfg` file under openHAB 1.x, which is **not** part of openHAB 2.
+When using config examples from the [openHAB 1 wiki](https://github.com/openhab/openhab1-addons/wiki), please be aware, that those need to be split into individual binding config files for openHAB2:
 
-After installing a Legacy or regular 1.x binding an individual `<binding_name>.cfg` file is created in the `/services` folder. If this file is not created automatically it needs to be added manually.
-The naming convention has to be: `<binding_name>.cfg` without the trailing "1" in the binding name grabbed from "PaperUI -> Add-Ons -> Bindings", for example the binding name from PaperUI for the MQTT binding is "binding-mqtt1-1.9.x", the config file you have to create is "mqtt.cfg".
-Next step is to remove the binding prefix name from the config lines in openhab.cfg and add the remaining part to the `<binding_name>.cfg` file, for example
+After installing a legacy or regular 1.x binding, an individual `<binding_name>.cfg` file is used in the `/services` folder.
+If this file is not available, please added it manually.
```
ThomDietrich: "created" was better
added -> add

sihui62: > I hope that's okay with you.

👍

ThomDietrich:
@sihui62 @watou good to go?

sihui62:
> good to go?

From my side: yes.

addons/bindings.md
```diff
@@ -27,13 +27,36 @@ The following bindings can be installed and configured directly from within Pape
...
 Bindings developed for the first version of openHAB can also be used with openHAB 2.
 Please be aware, that these bindings function and behave differently.
+They do for example not support the auto-discovery feature of openHAB 2 and do not define [openHAB 2 things and channels]({{base}}/concepts/things.html).
 For beginners it is recommended to resort to native bindings if possible.
 
-To be able to see and install 1.x bindings needs to be activated in the system settings.
+If there is already a openHAB 2 version of a binding available, old bindings are referred to as "Legacy 1.x Bindings".
+To be able to see and install legacy 1.x bindings, one has to activate "Include Legacy 1.x Bindings" through either Paper UI or `addons.cfg`.
+
+Configuration and usage of 1.x bindings is best done in configuration files.
```
watou: > best done

only true in the case where the 1.x add-on does not yet have an `ESH-INF/binding/binding.xml` or `ESH-INF/config/config.xml` file.  If it does, then the user has a choice of configuring via Paper UI, .cfg file, or if they really want to, the Karaf console `config` commands.

watou: Otherwise, LGTM.

ThomDietrich: Okay I get that this was too absolute.
So the statement should be that "some" 1.x bindings can also be configured via Paper UI? I don't like that either :) Suggestions?

sihui62: What about:
Setup and usage of 1.x bindings, which are not yet revised for openHAB 2, have to be configured through text based files.

ThomDietrich: @watou just an idea. Would you be possible to add or provide info in the possibilities of karaf around `config` in an independent PR/Issue? That would be most amazing 🥇

ThomDietrich:
Thanks!

PR #266, opened 2017-01-24
Search input hotfix

kubawolanin:
Apparently [placeholder property](https://css-tricks.com/almanac/selectors/p/placeholder/) is not very well standardized.
Fixed the margin overwriting materialize.css rule (thus the infamous `!important` ;-))

Signed-off-by: Kuba Wolanin <hi@kubawolanin.com>

### DIFF
--- a/css/openhab.css
+++ b/css/openhab.css
@@ -24,13 +24,19 @@ li.search:hover {
   border: none !important; 
   transition: 0.2s ease-in;
   display: block;
-  margin: 17px 0 0 0;
+  margin: 17px 0 0 0 !important;
 }
 
-.search-form ::-webkit-input-placeholder,
-.search-form ::-moz-placeholder,
-.search-form ::-ms-input-placeholder {
-    color: #fff;
+::-webkit-input-placeholder {
+  color: #fff;
+}
+
+::-moz-placeholder {
+  color: #fff;
+}
+
+::-ms-input-placeholder{
+  color: #fff;
 }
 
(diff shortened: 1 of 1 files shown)


css/openhab.css
```diff
@@ -24,13 +24,23 @@ li.search:hover {
...
 
-.search-form ::-webkit-input-placeholder,
-.search-form ::-moz-placeholder,
-.search-form ::-ms-input-placeholder {
-    color: #fff;
+.sticky .search-form-input {
+  margin: 9px 0 0 0 !important;
+}
+
+::-webkit-input-placeholder {
```
MHerbst: Using this without class or id is dangerous. There is at least one other input box that already has a white background, e.g. the input box that appears on mobile devices (id is querymob). 

Same is true for the other placeholder pseudo elements.

css/openhab.css
```diff
@@ -24,13 +24,23 @@ li.search:hover {
   border: none !important;
```
MHerbst: Some lines before are duplicate width entries (I probably forgot the remove one ...).

kubawolanin: Done!

MHerbst:
Looking good now  ... at least on my desktop in full width. If you make the window smaller (like on a mobile) the search input box is not visible at all because the placeholder is not shown. Maybe we should add for #querymob a bottom-border line. @kubawolanin

ThomDietrich:
Could you also structure the file by the help of comments, please

PR #273, opened 2017-01-26
Add `[config|custom].properties` to update list

### LINKED ISSUE #244: Docu Windows Update
I followed the new description how to update openhab on http://docs.openhab.org/installation/windows.html

Running the start.bat ´does not start openhab and I get following errors:
```
Launching the openHAB runtime...
karaf.bat: Ignoring predefined value for KARAF_HOME
WARN: file:/C:/openhab-offline/userdata/etc/config.properties is not found, so not loaded
null
WARN: file:/C:/openhab-offline/userdata/etc/config.properties is not found, so not loaded
Error occurred shutting down framework: java.lang.NumberFormatException: null
java.lang.NumberFormatException: null
        at java.lang.Integer.parseInt(Unknown Source)
        at java.lang.Integer.parseInt(Unknown Source)
        at org.apache.karaf.main.ConfigProperties.<init>(ConfigProperties.java:208)
        at org.apache.karaf.main.Main.updateInstancePidAfterShutdown(Main.java:226)
        at org.apache.karaf.main.Main.main(Main.java:191)
```

Something must be wrong in the description

NoTechi

BClark09:
Fixes #244 (hopefully)

### DIFF
--- a/installation/windows.md
+++ b/installation/windows.md
@@ -71,6 +71,8 @@ There is currently no automatic update script for Windows. To update manually, d
  - `userdata\etc\all.policy`
  - `userdata\etc\branding.properties`
  - `userdata\etc\branding-ssh.properties`
+ - `userdata\etc\config.properties`
+ - `userdata\etc\custom.properties`
  - `userdata\etc\distribution.info`
  - `userdata\etc\jre.properties`
  - `userdata\etc\profile.cfg`


ThomDietrich:
@BClark09 the error said "config.properties is not found", because of that @DimitrisSar removed config.properties from the list of files. Was that incorrect?

BClark09:
I'm not so sure, These two files are present in the zip file that a user wants to upgrade to:

image

The user is expected to say "Do not replace" for anything that exists in this folder. So we remove these files so that when we copy and paste from zip to the existing folder, they new versions are present. 

Perhaps a batch file will make this much easier:

openHABUpgrade.bat C:\openhab openhab-2.0.0.zip

Let me see what I can do...

kaikreuzer:
> Perhaps a batch file will make this much easier

As mentioned in https://github.com/openhab/openhab-distro/issues/299#issuecomment-268253707: I think we need such an "upgrade" script for both Linux and Windows, which should then probably be used by manual, apt, synology, docker, qnap, snap, etc. upgrades likewise.

ThomDietrich:
Yeah it's about time. It's a bit embarrassing to ask users to delete a bunch of files, then do this, then that, just to upgrade to a new version.

@BClark09 Would be great if you could come up with such a bat file for windows. I'd suggest to bundle it inside the zip file. Then the user can call it from inside the extracted new version folder. I've seen that approach a lot.

There could also be a bat script in the existing installation, prompting for the zip to upgrade on, but that could create conflicts with needed new changes inside the script.

Which approach would you suggest?

DimitrisSar:
I think that I shouldn't have removed the reference to delete the `config.properties` file during the upgrade steps... It exists in the distro zip file.

I haven't found any free time to progress on the batch script story for windows :(

Also: There is a small error in the PR that I created recently and it was merged:
In section "Installation", steps #3 & #4 are miss-numbered (auto numbering didn't work and they appear as "1").  Maybe you can fix this small issue.

BClark09:
Will do!

So far, this batch script might work. When placed in the main openHAB directory, simply drag the zip into it and it should upgrade.

https://gist.github.com/BClark09/74984daa39e0535521da7589e976575f

To do: Auto backup first!

BClark09:
I've finished the batch script, seems to work well on a number of Windows operating systems, but I haven't tested it fully. Is there anyone willing to do this?

I've never made a batch script before, so go easy on me if I've made silly mistakes:

https://gist.github.com/BClark09/74984daa39e0535521da7589e976575f

kaikreuzer:
@BClark09 Question: Would we find a way where the information, which files need to be copied/overwritten/moved/renamed could reside in a file OUTSIDE of the upgrade script, so that this info could be used from both Windows and Linux upgrade scripts and we would not have to maintain both at the same time, if some new file to copy comes into the game?

BClark09:
Yes, I believe if the list of these files were stored in a text file inside the distribution, then the batch script can iterate down the list i.e.

```
for /f %%a in (systemfilelist.txt) do (
[[Do something special to %%a]]
)
```

Issue #258, opened 2017-01-23
Header logo link


paulianttila:
Should openHAB logo link in the header go back to openHAB main page (www.openhab.org) rather than docs.openhab.org index page? If not, then header probably should have a link to openHAB main page (www.openhab.org).

MHerbst:
Linking he log in the header to the openHAB main page would be very confusing. A click on the log normally brings you to the home page of a site (in this case docs.openhab.org).
But I agree with you that a link to would be fine.  

I am planning to a add footer line (#223) for links to imprint and privacy. Would could a link to the openHAB website to this line.

Issue #223, opened 2017-01-17
Add Imprint and Privacy Policy to website


MHerbst:
The website uses Google Analytics. So we should add a Privacy Policy page with an opt-out policy for Google Analytics. To avoid any problems I would recommend to add an imprint too.

For a Privacy Policy page we can use the first half of http://www.openhabfoundation.org/privacy. The domain openhab.org is owned by Kai, so we would need a different text for the imprint if we want to add one.

I would add the links to a footer line because the pages must be accessible easily and in the footer the links do not disturb.

MHerbst:
Hast been solved quite a long time ago ...

PR #277, opened 2017-01-28
Added Imprint and Privacy Page

MHerbst:
Also removed unused twitter script coding from the footer.

Solves #223 
Solved #258 

Expected look:
image


kaikreuzer:
I just noticed that the footer now flows into the menu on the left:

screen shot 2017-02-01 at 11 39 07

@MHerbst Could you have a look at that? Btw, I think the font size might be a bit smaller as well.

MHerbst:
> I just noticed that the footer now flows into the menu on the left:

This happens only on pages where the expanded menu is larger than the text. This can be a bit complicated because of the navigator sidebar that is dynamically expanded. As a quick bypass I could change the z-Order so that the text will lie behind the menu.

> Btw, I think the font size might be a bit smaller as well.
Of cousre, I think 0.8 rem would be sufficient.

ThomDietrich:
I'd suggest to move the footer inside the content container. In the long run we should try to bring the whole page more in-line with the ideas and principles supported by Materialize... http://materializecss.com/footer.html

MHerbst:
@kaikreuzer  As I already thought it is not that easy to move the footer in a better position (due to the current structure of the page). As there are only some pages affected I only modified the z-order and the footer line is now partially hidden by the menu if the menu is larger then the text.

>I'd suggest to move the footer inside/besides the content container.

I've tried it but this would mean a larger refactoring of the theme. The biggest problem is that it does not really make use of the grid provided by materializecss.

> In the long run we should try to bring the whole page more in-line with the ideas and principles supported by Materialize... 

Very good idea! For my taste there is sometimes too much plain HTML content in the pages. Would be greate if we could imrove this and get as much pure markdown as possible. 

When you are done so far that I could help you, please tell me.

AngelosF:
2 very small issues that I noted for the http://docs.openhab.org/privacy.html:

a) There is no first section title in the page (should it be: "Privacy Policy" or something similar?) and
b) the auto-generated link to the second section title "Privacy Statement for the Use of Google Analytics" is not working (URL changes to `http://docs.openhab.org/privacy#undefined`)
privacy

ThomDietrich:
Good points @AngelosF 

Regarding b) That's a general problem with level 4 (and 5) headlines. Has been in my head for a while but I didn't look into that yet. Probably an easy fix.

MHerbst:
Thanks @AngelosF for your remarks. I have copied the texts from www.openhab.org and did not see that the heading was missing. Ouch! 

Will do the changes this weekend. Especially for d) I will have to check and change some other pages too :-(.

@ThomDietrich In this case it is a level 3 headline.

AngelosF:
For b): I usually use the `#` sign to define headers and then the auto-generated Anchors work fine.
Example: 
```
# This is an <h1> tag
## This is an <h2> tag
###### This is an <h6> tag
```

I see that you define the headers using html (`<h3>Privacy Statement for the Use of Google Analytics</h3>`)

Maybe this is the reason why the Anchors are not created ?

References:
1) https://gist.github.com/asabaylus/3071099#anchors-in-markdown
2) https://guides.github.com/features/mastering-markdown/

MHerbst:
Good point @AngelosF. As far as I can see it should be no problem to use Markdown for this page. Will modify it.

PR #279, opened 2017-01-30
Fix typo

milo0:
namp -> nmap


kaikreuzer:
Thanks - please note that the source of this page is [here](https://github.com/openhab/openhab2-addons/pull/1803/files), so I have ported your fix there.

PR #283, opened 2017-02-01
Modified footer

MHerbst:
- smaller font size
- added non-breaking spaces to prevent unwanted line breaks
- put footer in background if menu is larger than text

See also my remarks in #277 


kaikreuzer:
Many thanks!

Issue #306, opened 2017-02-07
Clarification: Item Type Group in the Frame element of a sitemap


AngelosF:
We should add a clarification to the http://docs.openhab.org/configuration/sitemaps.html#element-type-frame that Item Type Group cannot be used in the Element Type Frame.

Potentially, add an example on how to correctly display a Group within a Frame:
```
Frame {
        Group item=gGroup_Name
}
```

The following does not work:
```
Frame item=gGroup_Name
```

Ps: https://community.openhab.org/t/automatic-frame-generation-from-group-items/22117

kaikreuzer:
I wasn't even aware that you can define a "normal" item for a frame. What does that do? How is that rendered? I am not sure that this actually makes much sense at all... The frames are really only meant to do some visual separation with a title, nothing more than that.

AngelosF:
It just renders the item info in the "header" of the frame
Example:
```
Frame item=WU_Temp
```

With item definition:
```
Number	WU_Temp		"Temperature [%.2f °C]"		<temperature>	(Weather,gInflux)	{weather="locationId=HomeR, type=temperature, property=current"}
```

Will result to:
![frame_item](https://cloud.githubusercontent.com/assets/25465966/22684309/7e5fc81e-ed24-11e6-879f-91814817d4ed.png)

kaikreuzer:
That's what I was wondering - showing some item status simply does not make sense for the frame. I would rather suggest to remove this "feature" completely.

AngelosF:
I also agree...
Should I close this issue then or we should "re-use" to remove from the sitemaps docs entry the option to use item in the Frame element syntax?

kaikreuzer:
For me ok to re-use as the title is about a "clarification" :-)

PR #310, opened 2017-02-08
Remove `item=` parameter from Element Type Frame

AngelosF:
Closes: #306

Signed-off-by: Angelos agf@itcs.ro (github: AngelosF)

### DIFF
--- a/configuration/sitemaps.md
+++ b/configuration/sitemaps.md
@@ -148,14 +148,12 @@ This provides the flexibility to present items in the way desired in your home a
 ### Element Type 'Frame'
 
 ```xtend
-Frame [label="<labelname>"] [icon="<icon>"] [item=<item>] {
+Frame [label="<labelname>"] [icon="<icon>"] {
         [additional sitemap elements]
 }
 ```
 
 Frames are used to create visually separated areas of items.
-
-* `item` will provide data presented in the head of the frame, above all included elements.
 
 ![Presentation of the frame element in BasicUI](images/sitemap_demo_frame.png)
 


watou:
Excellent -- this was a documentation mistake in openHAB 1.x that some widgets had `item=` documented but they didn't do anything and made no sense. 😄

ThomDietrich:
I've actually tested all of these when rewriting the sitemaps article and "item" is indeed supported by the Frame element type.
However I saw the discussion in #306 and would also agree to remove it.

Thanks @AngelosF

watou:
>  "item" is indeed supported by the Frame element type

While it may be supported syntactically, if it has no meaning and leads to confusion, it shouldn't be documented! 😄

ThomDietrich:
> has no meaning

Well, that's your opinion.

2017-02-08 12_45_38-w66a openhab installation - pale moon

watou:
> Well, that's your opinion.

I understood Kai's comment yesterday to be a definitive statement.  From https://github.com/openhab/openhab-docs/issues/306#issuecomment-277931667:

> I wasn't even aware that you can define a "normal" item for a frame. What does that do? How is that rendered? I am not sure that this actually makes much sense at all... The frames are really only meant to do some visual separation with a title, nothing more than that.

ThomDietrich:
Ah I see.
I think the decision made here was "Either fix the brackets or remove the possibility". And as - as we seem to all agree - a value for a frame might seldom make sense, we just removed the documentation part for it.
Is an actual fix to remove or correct the presentation in the UI/API of interest? Maybe.

Timtam:
Just to warm this up again, I just found this issue to be rather helpful to me. I'm currently building a forecast sitemap which shows the next 7 days weather forecast and thus each frame should have the date of the day in its label.
Because dynamic labels don't seem to be possible, I created a few items which get updated by a rather simple rule and now get used via Frame item=date_forecast_0d etc.
I don't find the brackets to be too disturbing, but I'd like to have the Frame element with its item parameter to be a feature rather than an undocumented hint somewhere within the forums.

ThomDietrich:
Hey @Timtam I've since continued to use this as well. Would you like to drive this? Please create an issue here https://github.com/openhab/openhab-webui/issues for BasicUI to fix the brackets issue. Reference back to here and as soon as the issue got solved we can reintroduce the documentation part.

PR #317, opened 2017-02-11
[Addons] “since” badges have no value for actions or persistence add-ons

watou:


watou:
Superseded for actions; not correct for persistence.

PR #320, opened 2017-02-12
added logos to add-on lists

kaikreuzer:


kaikreuzer:
@ThomDietrich & @watou: I played around with all kinds of option to position the logo (right column as discussed or as a new column at the very left), but I didn't like most of them.
The imho best solution was to replace the label text, if a logo is available as in >95% the logo is really making the label obsolete. This is how this now looks like:

screen shot 2017-02-12 at 22 55 59

Ideally, we should have logos for almost all add-ons, then it would be even better - I only provided a subset for the start.

I still think that the badges catch the eye too much, especially as they should just be a "minor detail" - but I do not have any better suggestion.

ThomDietrich:
Hey Kai,
This looks great. I agree the Logos can replace the label but I'd suggest to set the label as both alt and title tag of the image.

kaikreuzer:
Right, makes sense!
Will wait for @watou's feedback before I adapt the PR.

ThomDietrich:
Also in the css this might be a better match: 
```css
#body.documentation section#documentation .content img.logo {
.addon-table img.logo {

```

kaikreuzer:
I tried that, but the setting was overwritten by something in styles.css.
I am a total CSS noob, but if you have a good advice how to make it work more elegantly, I'm all ears :-)

watou:
Looks nice!  Some thoughts:

* You might want to move the icon column in the .csv files near the beginning of the line, in the odd case that there is a comma in the description column.  Just to note that the bindings/actions/persistence tables no longer use the description column from their .csv files.
* With the labels no longer text for entries with icons, "search in page" in the browser will be less reliable, so it's a reason to try to make sure the description contains the keywords one might search for.
* The badge "install: auto" could be suppressed, since it isn't described among the checkboxes at the top of the page, meaning that only "install: legacy" and "install: manual" badges would be shown of the "install" badges.  I think the remaining badges are important information and I don't think there's a better spot for them.
* I agree with @ThomDietrich's suggestions above, including having the logo class in openhab.css.  It might be possible to center the image with `margin-left: auto !important; margin-right: auto !important;` in the CSS, but that is one of those things one has to experiment with.

kaikreuzer:
> set the label as both alt and title tag of the image.

done

> move the icon column in the .csv files near the beginning of the line

done

> "search in page" in the browser will be less reliable

Valid point, but most add-ons indeed seem to repeat the name in the description, so I think we can live with that.

> The badge "install: auto" could be suppressed

We could have a try how that looks like then, but it is out of scope of this PR for me.

>  center the image

I prefer left alignment as the label texts are also aligned on the left. I wouldn't want to treat the logo as a label replacement differently.

watou:
I pulled your PR and suggest a small change, otherwise LGTM.

css/openhab.css

```diff
-body.documentation section#documentation .content img.logo {
-  max-height: 60px;
-  max-width: 150px;
+.addon-table img.logo {
+  max-height: 60px !important;
+  max-width: 150px !important;
```

kaikreuzer:
thx, updated.

watou:
A few more logos for your consideration:

https://github.com/watou/openhab-docs/commit/1416817b85369afb5e19d58850f069f61ac14412

kaikreuzer:
Nice, but you should trim the white space around them.

kaikreuzer:
@ThomDietrich Up to you to merge this.
@watou Can do follow up PRs with additional logos then.

ThomDietrich:
Just to be sure: Did you add alt and label *only* to bindings logos on purpose?

kaikreuzer:
Err, no, no purpose at all - just a mistake ☹️

kaikreuzer:
updated

ThomDietrich:
LGTM!

ThomDietrich:
Great results! :clap:

PR #327, opened 2017-02-15
Update OLA links

peternewman:

PR #284, opened 2017-02-01
Factor OH1 add-on docs into site

watou:
and filter lists by source and install status.  

Fixes #280

Signed-off-by: John Cocula <john@cocula.com> (github user: watou)

PR #328, opened 2017-02-15
Identify source file in front matter for all OH1 add-ons

watou:
Attempting to minimize situations like [this](https://github.com/openhab/openhab-docs/pull/327#issuecomment-280059224) by implementing some of what @ThomDietrich requested [here](https://github.com/openhab/openhab-docs/pull/284#issuecomment-276951514):

> (If possible) please add an additional front matter (e.g. "source") directly pointing to the repository file as a full URL. This will be helpful in the above case and for an "Improve this article" button I was thinking about adding.

Same cannot be done for OH2 addons due to multiple repositories, but there may be a solution.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -347,8 +347,13 @@
 		        						}
                 			 		}
                 				}
+                				def toYaml = { '---\n' + it.collect{ /$it.key: $it.value/ }.join('\n') + '\n---\n' }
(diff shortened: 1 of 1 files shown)


watou:
Superseded by #332

ThomDietrich:
Right, thanks for adding these!!

PR #332, opened 2017-02-18
Put content on addons/io.md page

### LINKED ISSUE #330: IO/System Integration articles 404 from menu
@watou could you please have a look? http://docs.openhab.org/addons/io.html

watou:
Fixes #330

Also:

* Removed manually maintained submenu contents
* Updated and added icons
* Fixed small bug in actions page


ThomDietrich:
Looks good.
I'd leave the menu though. Otherwise the individual articles will not have a reference in the menu at all and the menu will fold completely. This problem now already exists with the Bindings: http://docs.openhab.org/addons/bindings/airquality/readme.html

watou:
> This problem now already exists with Bindings

I will look again to see if there is a way to keep the menus from folding, instead of filling the menu with dynamic content that could be very very long.  That would address all add-on menus (if possible). Would that work for you @ThomDietrich?

ThomDietrich:
Yes that would be okay with me.
I do not understand why you are so eager to remove content!? I'd just leave the menu as it is, I have build it only a few months ago for good reasons...

watou:
> I do not understand why you are so eager to remove content!? 

The alternative is to dynamically generate the submenus so they contain identical links to the main `addons/<type>.html` page, which seems redundant and for some add-ons is a very long list and not really appropriate for a menu, IMHO.  (To have static content in the menu but dynamically build the `addons/<type>.html` content seems wrong to me, as this would require PRs for user-menu.html every time an add-on changes disposition.)  It's also not possible to filter the menu based on criteria like is done in the main `addons/<type>.html` page, so all addons of a given `<type>` would appear on the menu with no further context.

> I'd just leave the menu as it is, I have build it only a few months ago for good reasons...

I assumed one reason was a lack of content for the main addons/type.html page, which is now solved for actions, bindings, io and persistence.  Another reason would be to work around Jekyll's menu collapsing since the addons are one level lower than expected for that to work by default.  Were there other reasons to put an manually curated list of add-ons in user-menu.html?

ThomDietrich:
>  Were there other reasons to put a manually curated list of add-ons in user-menu.html?

Of course it's better to have an automatically generated list e.g. for the bindings. For system integration I went the manual way as there were only a hand full and IFTTT was a bit special, hence the `<hr>`.

Till now **all** docs pages have a menu entry. The reason is simple: consistency. Everything should be accessible through the menu - at least I tried to follow that principle. I'm not sure if we should brake with that.

> a very long list and not really appropriate for a menu, IMHO

Hmmm you know I actually caught up with that change for bindings when I was building around in my automation and needed different bindings docs... suddenly I couldn't access them as before anymore :( Imho a long menu doesn't hurt anyone who doesn't use it but will be missed by those who do.

Now, there is one downside of the bindings menu I do not want to conceal: Responsiveness. The bindings list will be shown in whole on a smartphone above the article and the long list is a problem. However on a smartphone the menu shouldn't be shown unfolded by default to begin with, imho. These are all things I wanted to address when updating the page [in regards to Materialize](http://materializecss.com/side-nav.html)... I can't promise any deadline for that right now 😢 

Okay, let's try to find a compromise. I'd vote to stay with a menu. I don't have a strong believe how this menu is defined, as long as it is complete and working. Wdyt?

watou:
Do you think that if the menu included all add-ons, that there is no problem with a menu Tellstick being followed directly by another menu item Tellstick?  Also, on mobile devices, should there be a 200+ long menu to scroll through before the user can reach the main content page?  These are two of the problems I see in having all add-ons appear in the menu hierarchy.

There are now too many entries (in my opinion) to appear in the menu hierarchy for all add-ons, and there will likely be many, many more in the future.  The lists of add-ons are more like rows in a database and not part of the static structure of a website's sitemap.  I feel it would be a mistake to represent them in the menu hierarchy.  Not doing so across all types of add-ons would achieve the consistency objective that we both share. :-)

watou:
I looked at user-menu.html again, and immediately noticed that this manually curated file was lacking links to dropbox and gcal, thereby giving users the incorrect impression that those services did not exist.  I don't think that either manually maintaining add-ons in the user menu, or programmatically generating them regardless of length of the list or context (install: manual, etc.) is correct.  However, a fix for the collapsing menu when drilled down to a specific add-on would be nice in a further PR.

ThomDietrich:
Hey @watou, I took a few minutes to think about it 

> Not doing so across all types of add-ons would achieve the consistency objective that we both share. :-)

If it's similar across all add-on sections we've achieved that indeed. Okay let's do it. Regarding the menu: I've tuned the scripts behind it at an earlier point and believe that might be possible. Let's look at this at a later point though, after I had a chance to do the Materialize update. This is more than needed...

Okay, I'm good with removing the user-menu subentries and thereby have to say only one last thing: lgtm 😉

watou:
Thanks a lot @ThomDietrich!   

My only idea so far about avoiding the collapsing menus ends up being a bit kludgy, where instead of `addons/${typedir}/${id}/readme.md`, the build process leaves the processed README.md file at `addons/${typedir}/${id}.md` but then adds a [`permalink` style](http://jekyllrb.com/docs/permalinks/#pages):

> if you use a permalink style that omits the .html file extension, each page is rendered as an index.html file inserted inside a folder with the page’s name (for example, mypage/index.html).

like:

```yaml
permalink: /addons/${typedir}/${id}
```

this would generate `/addons/${type}/${id}/index.html`, which *might be* enough for the menus to not collapse any more.  The remaining issue is to resolve `readme.html` again so no one's links are broken again!  If GitHub Pages supports the [jekyll-redirect-from](https://github.com/jekyll/jekyll-redirect-from) plugin, this could be achieved with

```yaml
redirect_from: /addons/${typedir}/${id}/readme.html
```

So yeah, kind of kludgy. :-)

watou:
I also noticed a new strangeness, where clicking on an add-on doc link quickly goes to an `#undefined` anchor first, and then to the correct add-on doc page.  This seems to be new, since after the "addons-table" PR but before this PR.  Do you see the same thing on your side?

ShaunCurrier:
Joining the convo from #344 and only reading this issue halfway until I got lost in the weeds, here's my unsolicited opinions and observations.

- It seems that the current systems integration page available [here](http://docs.openhab.org/addons/io.html) is not *currently* so long as to preclude the presence of each individual item in the menu.
- Even if the list gets longer and causes minor pain for small screen users, maybe that's an acceptable price to pay in exchange for the ease of discovery in the docs, which is important - maybe more important than keeping small screen users from scrolling long lists.
- There is something to be said for the docs menu structure mirroring the tabs that appear in the Paper UI of OpenHAB (for example the IO tab).  If the docs mirror the tabs, then users will be better able to make the connection between what they read in the docs and what they're seeing in the default OpenHAB administration GUI.

...Just my opinion as someone with a fresh mind who doesn't carry the burden of all the valuable rationale that went into how the docs are structured already.  Maybe my opinion's valuable... or maybe not.

watou:
Thank you for sharing your opinions on this!  Doing so can only result in an improved outcome.  

Removing confusion or the potential for confusion would certainly be Job 1 for the documentation.  Consistency between the product and the docs (whether it be that PaperUI matches the doc layout or vice versa) makes good sense.  Navigability of the docs site (whether mobile, desktop or huge screen) should also not add to confusion.

I would encourage you to open new issues against the docs for each and every discrete deficiency you see, and open pull requests to match, if you are so inclined.  The beauty of open source! 😄

Issue #278, opened 2017-01-29
Invalid links


MHerbst:
Page http://docs.openhab.org/concepts/guidelines.html contains an invalid link to http://docs.openhab.org/concepts/guidelines.html. 

I think the correct target would be http://docs.openhab.org/developers/development/ide.html

On page http://docs.openhab.org/addons/uis.html all links in section "App" are invalid. They must be changed to "../addons/**uis**/apps/..."

Page: http://www.openhab.org/technologies.html links to http://docs.openhab.org/addons/bindings/gcal/readme.html which no longer exists.

ShaunCurrier:
I think your first point intends to say that the IDE Setup link is wrong.  I can confirm.

I also confirm that the App links are invalid.

I'm unclear which of the links on the technologies page is broken.  I can't find anything related to "gcal" on that page.  Which image did you click?

ShaunCurrier:
I fixed these issues in #353 except for the last part dealing with http://www.openhab.org/technologies.html.  I think that problem might need to get routed elsewhere.

kaikreuzer:
> I'm unclear which of the links on the technologies page is broken. I can't find anything related to "gcal" on that page. Which image did you click?

I think this does not exist anymore since I regenerated that page one or two weeks ago.
Many thanks for fixing the rest, @ShaunCurrier!

MHerbst:
> I think this does not exist anymore since I regenerated that page one or two weeks ago.

@kaikreuzer 's assumption is correct. I have checked the and they are OK now. Thanks for fixing the links @ShaunCurrier

ThomDietrich:
Can the issue be closed then?

MHerbst:
I wanted to wait until the PR was merged before closing this issue.

ThomDietrich:
Well then ;)

PR #353, opened 2017-02-26
Fix invalid links to IDE and apps docs (#278)

ShaunCurrier:
Does not fix the part of #278 affecting http://www.openhab.org/technologies.html as I think that might belong to another repository and I wouldn't know how to fix it.

Please review the way the relative links were done as I'm new to this mechanism.


ThomDietrich:
Thanks!

PR #359, opened 2017-03-02
Fix #358

### LINKED ISSUE #358: Addons: fix error on addons/persistence.html (Exec)
Per https://github.com/openhab/openhab-docs/pull/357#discussion_r103924707

The code [here](https://github.com/openhab/openhab-docs/blob/gh-pages/process_addons.groovy#L71) attempts to mate up installable features from feature.xml files to the various package names, but because sometimes "misc" features come from "io" packages, "transformation" features come from "transform" packages, and maybe some other incongruities, the code is overly lenient in the matching process.  This has led to one known mistake, where the "org.openhab.persistence.exec" package is mistakenly matched to the "openhab-transformation-exec" feature, so the Exec persistence service is incorrectly tagged as being from 2.x and installable, which at the moment is not true.

The issue can be worked around for the moment by simply testing the exec persistence service under OH2 and creating a feature for it, but ideally the matching logic could be improved.

watou:


ThomDietrich:
Thanks!! Happy to see that you went the extra mile for the probably more complicated solution!

Issue #204, opened 2017-01-15
Initial Setup: How to reset


ThomDietrich:
https://community.openhab.org/t/a-single-distro-from-now-on/19903/40

>Actually, the package selection does nothing else than creating the file `userdata/etc/org.openhab.addons.cfg`. So if you delete this file and set a different package in `conf/services/addons.cfg` it will create a new file with the addons specified of that package. But there really isn't a need to switch package, because you can simply change the installed addons yourself at any time and for your personal needs.

halfbaked:
@ThomDietrich PR here https://github.com/openhab/openhab-docs/pull/371

Confectrician:
There is only one package (standard) left in openHAB 3

PR #371, opened 2017-04-11
Changing initial package selection

halfbaked:
Added section to configuration/packages on how to change the package from the initial selection made.
Solution taken from here: https://community.openhab.org/t/a-single-distro-from-now-on/19903/40
Addresses issue: https://github.com/openhab/openhab-docs/issues/204

Signed-off-by: Eamonn O'Connell <eamonnoconnell@gmail.com> (github:halfbaked)

### DIFF
--- a/configuration/packages.md
+++ b/configuration/packages.md
@@ -77,3 +77,10 @@ This includes:
 - MAP transformation service as this is heavily used by the sample files
 - Sample configurations files `demo.items`, `demo.things`,  `demo.sitemap`, `demo.rules`, `demo.script`, `rrd4j.persist`, `de.map` and `en.map` for use with the MAP transformation
 
+# Changing Initial Package Selection
+
+It is common to have initially selected one package (e.g. Demo or Simple) during installation and to later want to go back and change to another (e.g. Standard). To avoid redoing the installation again from scratch, do the following:
+
+1. Remove the config file `/var/lib/openhab2/etc/org.openhab.addons.cfg` which was first created by your initial package selection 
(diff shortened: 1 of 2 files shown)


ThomDietrich:
Hey, this was quick and is a good read! ;)

I'm not entirely sure if we should state it like this. First we should make clear, that the initially selected package is not *that* important. One can just install addons and packages as he wants. As Kai said:

> But there really isn't a need to switch package, because you can simply change the installed addons yourself at any time and for your personal needs.

Also it would be better to be able to see the Initital Setup Screen again instead of resorting to `addons.cfg`. Did you try that?

It might also be worth mentioning that this change will only influence packages. If one wants to start over from zero they should to an uninstall+reinstall.

Lastly it may be good to link this new chapter in the introduction.

Side Note: I was surprised to see your changes to be _italic_. Turns out this is related to the `_blank`s throughout the text. Maybe it would be wise to escape the underscore (`\_blank`) but I'm not even sure if the link would still work then, let's just leave it.

halfbaked:
Yeah, I prefer to do it, than leave it hanging over me :)

Is it true to say: a package is a bundle of add-ons to get the user
started, where the add-ons include UIs, bindings, persistence services, and
actions?

If I start with the standard package, and then switched to the simple
package, it's not going to uninstall add-ons right?

ThomDietrich:
I'm afraid we would need to test this. I could use one of my openHABian testing systems for that. This evening maybe, no guarantees... If you want to test yourself, that would be great.

PR #374, opened 2017-04-21
Update guidelines.md

martinvw:
The date in the about.html should not always be changed.

See https://github.com/eclipse/smarthome/issues/3270#issuecomment-296121716


martinvw:
@kaikreuzer can you review this one?

PR #381, opened 2017-05-03
added getting started to android UI

halfbaked:
Signed-off-by: Eamonn O'Connell <eamonnoconnell@gmail.com>
(github:halfbaked)

### DIFF
--- a/addons/uis/apps/android.md
+++ b/addons/uis/apps/android.md
@@ -34,8 +34,16 @@ The latest release version of the app is always available through Google Play.
 
 **Getting Started:**
 
-When first installed the app is in demo mode. To connect the android app to your openhab server, first navigate to Settings and uncheck the Demo mode option.
-Then set the openHAB URL to the URL of your server. For example: `http://192.168.1.3:8080`. The display will now be determined by the sitemaps defined on your server. 
+When first installed the app is in demo mode. To connect it to your openHAB server, first navigate to Settings and uncheck the Demo mode option.
+Then set the openHAB URL to the URL of your openHAB server; depending on your setup, the URL might look like one of the following examples:
+* `http://192.168.1.3:8080` (IP address)
+* `http://openhabian:8080` (local dns)
+* `http://openhabian.local:8080` (local dns)
+
+Once the URL is set correctly, the display of the app will be determined by the sitemaps defined on your server. 
+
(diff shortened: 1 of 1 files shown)


addons/uis/apps/android.md
```diff
@@ -32,6 +32,11 @@ The latest release version of the app is always available through Google Play.
   <div class="col s12 m5"><img src="images/android_02.png" alt="Demo Widget Overview"></div>
 </div>
 
+**Getting Started:**
+
+When first installed the app is in demo mode. To connect the android app to your openhab server, first navigate to Settings and uncheck the Demo mode option.
```
kaikreuzer: This is indeed nasty that it is required (but I agree that atm it should be in the docs) - my hope is that https://github.com/openhab/openhab.android/issues/179 gets implemented one day.

ThomDietrich:
Hey, thanks for the addition! Two little requests:
1. In addition to the IP example would you show a dns name one, e.g. openhabianpi?
2. An additional sentence regarding the "openHAB Remote URL" address option would be great.

halfbaked:
Hey Thomas,

No problem at all.

I can add a second example of http://openhabianpi.local:8080/ no problem
Maybe it is my networking setup but if I just use the local DNS as
described here without the .local it will not work for me:
 http://docs.openhab.org/installation/openhabian.html#raspberry-pi

I've not yet tried the remote url, so am not familiar. If you know enough
to explain to me when the android app will know to use the remote url and
when it will use the other I can add something.
Otherwise, I suggest to park that aspect for now.

ThomDietrich:
I've once read up on this topic myself. When or when not to use the .local address seems to depend on your device and network and how DNS resolution is handled. The ".local" name is as far as I know provided by mDNS, for which openHABian also contains avahi-daemon. On subnets I'm working with, the ".local" is not needed.
Also see https://www.howtogeek.com/167190/how-and-why-to-assign-the-.local-domain-to-your-raspberry-pi

Regarding the remote URL, I'd guess it's used after a connection on the local address was not successful. Maybe @digitaldan can shed some more light here. For documentation a simplistic description would suffice I'd say.

halfbaked:
Hey Thomas,

thanks for the link, it cleared up the whole .local mystery for me :)

I've pushed changes on both items you asked for.

Have a great weekend,
Eamonn

ThomDietrich:
@halfbaked Thanks!!

PR #383, opened 2017-05-04
rfxcom unload / load docs wrong way round

chrisns:


ThomDietrich:
Hello and Thank You for the contribution!
Many parts of the documentation are imported from other openHAB repositories, e.g. the component specific repository. This way we ensure that the articles are always in sync with the actual source code of the component (Add-ons, Bindings,...).

https://github.com/openhab/openhab2-addons/blob/master/addons/binding/org.openhab.binding.rfxcom/README.md

chrisns:
Cool, thanks for pointing me in the right direction I figured that I wasn't in the right place but couldn't find the original source.
raised https://github.com/openhab/openhab2-addons/pull/2250

PR #394, opened 2017-06-14
Chart element: value "2D" accepted as period

lolodomo:
Mention about not working refresh option for chart suppressed as it was fixed recently in Basic UI


ThomDietrich:
Thanks Nr2! ;)

Issue #402, opened 2017-06-17
[Items] Continue article, Attend to PR comments


ThomDietrich:
All comments from this PR are still not included: https://github.com/openhab/openhab-docs/pull/250

Other issues are available related to the items article.

ThomDietrich:
Further changes needed...

rkoshak:
Some suggestions. Over all it looks good to me.

In openHAB "Items" represent all properties and capabilities of the user’s home automation.
The quotes around Items are not necessary. If you want o emphasize it use italics instead.

This is wording correct but comes off as overly technical. Maybe reword to something closer to:

"In openHAB, Items represent all the data and controls of the user's home automation."

While a device or service might be quite specific, Items are unified substitutions inside the openHAB world.
+Items can be Strings, Numbers, Switches or one of a few other basic Item types, a programmer can rightly compare Item Types with base variable types of a programming language.
One unique feature of openHAB Items (in comparison to normal variables) is the possibility to connect them to the outside world (via Bindings, more about that later).
+An Item does not simply hold a certain information (e.g., "No Error", 3.141, OFF), the information is synchronized with the real world in both ways.
I fear this might cause confusion among new and non-programmer users. Something like:

"Items are the basic representation of all devices and services in openHAB. Items can be Strings, Numbers, Switches or one of a few other basic Item types. The Item type indicates how openHAB interacts with it (e.g. commanding a Switch to turn ON or OFF) and allows openHAB to ineract with them independent of what technology the Item may be connected to. Consequently, Items are used to store data and to interact with the outside world through bindings.

For example, all openHAB needs to know is an Item is a Switch, not that it is a Switch that happens to be linked to the ON/OFF channel of a Hue Light Bulb. And when openHAB sends an ON command to that Switch through one of the UIs or a Rule, that ON command gets turned into a signal to turn on the Hue Bulb by the linked binding."

I don't know if this is any better.

Items can be bound to Bindings or linked to Channels.
Would it be confusing to make a distinction this soon that bound to bindings is for OH 1.x and linked to Channels is for 2.x?

Other 1.x and legacy Bindings do not offer this path.
Strike this line. You basically say the same thing a couple lines later.

2.x Bindings can also be used with this method
Change to

"2.x Channels can be used with either method."

The last example above defines an Item with the following parts:
Does it make sense to focus on the 2.x channel linking example instead of a 1.x binding example here?

The Item name is the unique identified of the Item.
identifier

A good naming schema can be advised.
[...]

PR #373, opened 2017-04-20
Corrected default behavior for group functions

triller-telekom:
Signed-off-by: Stefan Triller <stefan.triller@telekom.de>

PR #250, opened 2017-01-22
Items article revision

rkoshak:
Addresses Issue #152.

Made 'Item' and 'Group' consistently capitalized
Minor grammatical and word choice corrections
Clarified the behavior of icons when used with a map
Added the table of contents
Clarified binding/linking to Items.

Signed-off-by: Richard Koshak rlkoshak@gmail.com (github: rkoshak)

PR #404, opened 2017-06-18
[Items] Continue items article revision, #250

### LINKED ISSUE #291: Items: Add use of 'autoupdate'
The docs entry for items currently does not provide any explanation of the feature to add `autoupdate=false` to the binding definition. 

http://docs.openhab.org/configuration/items.html 
or
https://github.com/openhab/openhab-docs/blob/gh-pages/configuration/items.md

More info:
https://github.com/openhab/openhab1-addons/wiki/Samples-Item-Definitions

ThomDietrich:
Continue the items article revision with all comments in #250.
The article is far from perfect. Many things still need to be added and improved.
This PR is another iteration and ready to be merged for now.

Related to #250 #402
Resolves #291 
Next look into #373

Kindly requesting a quick look by @metbril @Rossko57 @AngelosF and @rkoshak

### DIFF
--- a/configuration/items.md
+++ b/configuration/items.md
@@ -7,10 +7,16 @@ title: Items
 
 # Items
 
-openHAB uses a strict separation between the physical world and the application, which is built around the notion of "Items" (also called the virtual layer).
-Items represent functionality that is used by the application (mainly user interfaces or automation logic).
+In openHAB "Items" represent all properties and capabilities of the user’s home automation.
 
-This page is structured as follows:
+While a device or service might be quite specific, Items are unified substitutions inside the openHAB world.
+Items can be Strings, Numbers, Switches or one of a few other basic Item types, a programmer can rightly compare Item Types with base variable types of a programming language.
+
+One unique feature of openHAB Items (in comparison to normal variables) is the possibility to connect them to the outside world (via Bindings, more about that later).
+An Item does not simply hold a certain information (e.g., "No Error", 3.141, OFF), the information is synchronized with the real world in both ways.
+
+But let's not get ahead of ourselves.
+The rest of this page contains all details regarding Items and is structured as follows:
 
 {::options toc_levels="2..4"/}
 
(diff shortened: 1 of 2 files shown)


rkoshak:
Some suggestions. Over all it looks good to me.

> In openHAB "Items" represent all properties and capabilities of the user’s home automation.

The quotes around Items are not necessary. If you want o emphasize it use italics instead.

This is wording correct but comes off as overly technical. Maybe reword to something closer to:

"In openHAB, Items represent all the data and controls of the user's home automation."

>While a device or service might be quite specific, Items are unified substitutions inside the openHAB world.
> +Items can be Strings, Numbers, Switches or one of a few other basic Item types, a programmer can rightly compare Item Types with base variable types of a programming language.
One unique feature of openHAB Items (in comparison to normal variables) is the possibility to connect them to the outside world (via Bindings, more about that later).
>  +An Item does not simply hold a certain information (e.g., "No Error", 3.141, OFF), the information is synchronized with the real world in both ways.

I fear this might cause confusion among new and non-programmer users. Something like:

"Items are the basic representation of all devices and services in openHAB. Items can be Strings, Numbers, Switches or one of a few other basic Item types. The Item type indicates how openHAB interacts with it (e.g. commanding a Switch to turn ON or OFF) and allows openHAB to ineract with them independent of what technology the Item may be connected to. Consequently, Items are used to store data and to interact with the outside world through bindings.

For example, all openHAB needs to know is an Item is a Switch, not that it is a Switch that happens to be linked to the ON/OFF channel of a Hue Light Bulb. And when openHAB sends an ON command to that Switch through one of the UIs or a Rule, that ON command gets turned into a signal to turn on the Hue Bulb by the linked binding." 

I don't know if this is any better.

> Items can be *bound to Bindings* or *linked to Channels*.

Would it be confusing to make a distinction this soon that bound to bindings is for OH 1.x and linked to Channels is for 2.x?

> Other 1.x and legacy Bindings do not offer this path.

Strike this line. You basically say the same thing a couple lines later.

>  2.x Bindings can also be used with this method

Change to 

"2.x Channels can be used with either method."

> The last example above defines an Item with the following parts:

Does it make sense to focus on the 2.x channel linking example instead of a 1.x binding example here?

> The Item name is the unique identified of the Item.

identifier

> A good naming schema can be advised.

"A good naming schema is advised. Here are a few recommendations:

- Use meaningful names. Instead of "Node12" or "Relay27" use names like "Bedroom_Lamp".
- Name related Items similar to each other. For example, if you have a Timer Item using the Expire Binding related to the Bedroom_Lamp Item, name it something like "Bedroom_Lamp_Timer".
- Be consistant in how you name Items"

I think that will address another Issue I opened (or planned to, I can't remember now). I couldn't find one so maybe it wsa just a plan.

> This relation can be exploited in sitemaps or in automation rules.

"This relationship can be exploited in sitemaps, automation rules, and for persistence."

> By default, if no function is provided to the group, the Group uses OR.

I don't think this is true anymore as of 2.1 with the change in how Group states are calculated. You now must supply the function otherwise NONE is the default.

> +Because the Group state is an aggregate, every change in the Item members does not necessarily result in a change to the Group's state.
> +But, this also means that any single change to a member of the Group will result in multiple updates to the aggregate state of the Group.

I beleive the changes made to Groups mentioned above is a first step to fix this behavior. I haven't kept up with where they are on that though.

> Tagging is a new feature under development.

See the conversation on #411 

> Binding Configuration

"Binding Configuration and Channel Linking"

When ever we have a choice to default to something that is 1.x verses 2.x, we should default to the 2.x. So if we don't use the "and" in my suggestion we should change it to "Channel Linking".

> If you need to use legacy openHAB 1.x Bindings

"If you need to use legacy openHAB 1.x Bindings, i.e. 1.x Bindings for which there is 2.x version, .... This will make those 1.x version Bindings show up in PaperUI along side the 2.x versions."

> After downloading the legacy .jar files, they need to be placed in the `/addons/` folder.

"If one wants to install a 1.x binding that is not part of the official distribution or part of the IoT Marketplace, download the binding's jar file and place it in the `/addons/` folder."

There is a disctinction between the Legacy 1.x switch and enabling the 1.x compatibility layer documented here: http://docs.openhab.org/developers/development/compatibilitylayer.html#how-to-use-openhab-1x-add-ons-that-are-not-part-of-the-distribution.

> but the expire Binding will also listen and eventually modify the Item state.

"but the Expire Binding will change the Item's state to -999 when the mysensors binding fails to update the Item for more than 5m."

> Exception `autoupdate`:**
 
This section begs for an explanation of when you would want to use this. I have never used it so can't write a suggestion myself.

ThomDietrich:
@rkoshak Thanks! Could you please move the whole comment over to #402 ?

rkoshak:
Done.

PR #405, opened 2017-06-18
Updated mac-adress notation in the usage Example

Confectrician:
The notation of mac-adresses changed in one of the last updates.
This very small change fixes the example.

### DIFF
--- a/_bindings/amazondashbutton/readme.md
+++ b/_bindings/amazondashbutton/readme.md
@@ -101,7 +101,7 @@ __ Caution:__  You have to be aware that other Amazon devices might pop up in yo
 ```
 rule "Dash button pressed"
     when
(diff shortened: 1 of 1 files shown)


ThomDietrich:
Hello and Thank You for the contribution!
Many parts of the documentation are imported from other openHAB repositories, e.g. the component specific repository. This way we ensure that the articles are always in sync with the actual source code of the component (Add-ons, Bindings,...).

https://github.com/famake3/openhab2-addons/blob/master/addons/binding/org.openhab.binding.amazondashbutton/README.md

Confectrician:
This totally makes sense for keeping everything up to date 
and i will open a request in the addons repository.

Thanks for Clarification

PR #406, opened 2017-06-20
[restdoc] Elaborate installation of local rest api doc.

130s:
Being new to OpenHAB, I had to spend half an hour to find how I can install and where I can see the rest api doc.


ThomDietrich:
Thanks!

PR #413, opened 2017-06-23
fix icon links

mueller-ma:


ThomDietrich:
Hello and thanks for your contribution!
Many parts of the documentation are imported from other openHAB repositories, e.g., the Binding- or add-on specific repository. This way we ensure that the articles are always in sync with the actual source code.
Long story short, you'll need to propose your change in the correct repository: https://github.com/eclipse/smarthome/blob/master/extensions/ui/iconset/org.eclipse.smarthome.ui.iconset.classic/README.md

ThomDietrich:
@mueller-ma did you open a new PR?

Issue #343, opened 2017-02-22
Rules:  System-based Triggers language changes


rgerrans:
The following needs to be modified for the behavior of 'System started' since it runs on any rule refresh, not just the ones that it is in.

```
System-based Triggers

Currently, you schedule rules to be executed either at system startup or shutdown. _Note that newly added or modified startup rules are executed once, even if openHAB is already up and running._ They are simply executed once as soon as the system is aware of them. Here’s the syntax for system triggers:

System started
System shuts down
```

Suggested edit:

Note that adding or modifying any rules will cause startup rules to execute once, even if openHAB is already up and running. 

For reference:  https://community.openhab.org/t/startup-rules-are-executed-even-though-oh-already-running-for-hours/14202?u=rgerrans

bgilmer77:
I support the addition of the note.  This is NOT expected behavior, and could cause unintended consequences for OH2 users.  This seems like a bug, but for now, at a minimum, we should put users on notice that this behavior is normal.

ThomDietrich:
I feel the same way. "System started" suggests a different behavior and this should (1) be documented and (2) solved.
@kaikreuzer I did a quick search and didn't find an issue where this topic is discussed. "System started" should only trigger when openHAB is started, another keyword could be introduced to execute on every change. Is there an issue or discussion we can refer to, do you feel we should start one?

kaikreuzer:
No, there isn't any discussion.
The reason for the behaviour is rather simple: There is no "start time" of openHAB, as all components can come and go dynamically. So the system makes sure that every "system started" rule is executed as early as possible as soon as the system knows about it.

I wouldn't want to technically change anything there until we might have https://github.com/eclipse/smarthome/issues/1896 in place, which would give us a proper "system started event".

bgilmer77:
Thank you Kai for the explanation.  I have been running OH1 for a while and am now moving to OH2.  As a *NIX SysAdmin and native English speaker, I want to give back by contributing to the documentation, and I have just started working with Thom to learn how to do that.  I will contribute a little text around this to help others who come after me understand the behavior of the System started trigger.

kaikreuzer:
Many thanks @bgilmer77, I very much appreciate your efforts!

bgilmer77:
@kaikreuzer I have created a pull request, adding a little text to explain the behavior of the System started trigger at #416 .  You might want to look at it to see if I have described the behavior correctly.  This is the behavior I experience on my system.
Thanks.

[...]

PR #416, opened 2017-06-27
Add note on 'System started' trigger

bgilmer77:
The 'System started' trigger causes actions to execute when the system is started (i.e. a reboot).  But the trigger also fires when alterations are made to any rules files.  This could lead to unanticipated system behavior, since users may expect that the only time the trigger would fire is when the system initializes for the first time.

Sign-off: Signed-off-by: Brad Gilmer brad@gilmer.tv (github: bgilmer77)

### DIFF
--- a/configuration/rules-dsl.md
+++ b/configuration/rules-dsl.md
@@ -149,7 +149,7 @@ Two system-based triggers are provided as described in the table below:
 
 Trigger  |  Description
 ---------|-------------
-System started | Rules using the 'System started' trigger exeucte when openHAB starts and when changes or additions to *any* rule are recognized by openHab.  This is normal behaivor and should be anticipated when writing rules using this trigger.
+System started | Rules using the 'System started' trigger exeucte when openHAB starts and when changes or additions to *any* rule are recognized by openHAB.  This is normal behaivor and should be anticipated when writing rules using this trigger.
 System shuts down| Rules using the 'System shuts down' trigger execute when openHAB shuts down.
 
 ## Scripts
--- a/configuration/rules-dsl.md
+++ b/configuration/rules-dsl.md
@@ -151,6 +151,7 @@ Currently, you schedule rules to be executed either at system startup or shutdow
 System started
 System shuts down
 ```
+These system-based triggers work as expected.  Rules using the 'System started' trigger execute when openHAB starts.  Rules using the 'System shuts down' trigger execute as openHAB shuts down.  Additionally, rules using the 'System Started' trigger will execute when changes or additions to rules files are recognized by openHAB.  This is normal behavior and should be anticipated when writing rules using these triggers.
 
 ## Scripts
 
--- a/configuration/rules-dsl.md
+++ b/configuration/rules-dsl.md
@@ -145,13 +145,12 @@ You may also use [CronMaker](http://www.cronmaker.com/) or the generator at [Fre
 
 ### System-based Triggers
 
-Currently, you schedule rules to be executed either at system startup or shutdown. Note that newly added or modified startup rules are executed once, even if openHAB is already up and running. They are simply executed once as soon as the system is aware of them. Here's the syntax for system triggers:
+Two system-based triggers are provided as described in the table below:
 
-```java
-System started
-System shuts down
-```
-These system-based triggers work as expected.  Rules using the 'System started' trigger execute when openHAB starts.  Rules using the 'System shuts down' trigger execute as openHAB shuts down.  Additionally, rules using the 'System Started' trigger will execute when changes or additions to rules files are recognized by openHAB.  This is normal behavior and should be anticipated when writing rules using these triggers.
+Trigger  |  Description
+---------|-------------
+System started | Rules using the 'System started' trigger exeucte when openHAB starts and when changes or additions to *any* rule are recognized by openHab.  This is normal behaivor and should be anticipated when writing rules using this trigger.
+System shuts down| Rules using the 'System shuts down' trigger execute when openHAB shuts down.
 
 ## Scripts
 
--- a/configuration/rules-dsl.md
+++ b/configuration/rules-dsl.md
@@ -149,8 +149,23 @@ Two system-based triggers are provided as described in the table below:
 
 Trigger  |  Description
 ---------|-------------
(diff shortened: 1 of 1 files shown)


ThomDietrich:
Hello @bgilmer77,

first please add a new line in front of the new paragraph. All code blocks need to be surrounded by empty lines to be formatted correctly by the Jekyll markdown engine (kramdown).

What's the idea behind the first sentence? If not working as expected was normality, we could stop here 😄 

 I'd suggest to restructure the whole section. Instead of having the two triggers listed followed by one long descriptive text, how about a table or itemization listing the two triggers to the left and the description of both separated on the right in short sentences. This way you can also make the double meaning of "system started" clearer.

bgilmer77:
@ThomDietrich I agree that restructuring as you described would be best.  I am being careful because I do not want to make anyone upset.  For this reason, I only added the text after the original section.  However, your suggestion is better.  I will work on making these changes.

ThomDietrich:
You don't have to be overly cautious! As long as your changes are better than before you are adding value everyone appreciates.

Of course if the changes are worse, hell will rain down upon you! ......... just kidding 😆

bgilmer77:
I am used to it ;-).

bgilmer77:
Okay - cleaner and simpler.  See what you think.

configuration/rules-dsl.md
```diff
@@ -145,12 +145,12 @@ You may also use [CronMaker](http://www.cronmaker.com/) or the generator at [Fre
...
-Currently, you schedule rules to be executed either at system startup or shutdown. Note that newly added or modified startup rules are executed once, even if openHAB is already up and running. They are simply executed once as soon as the system is aware of them. Here's the syntax for system triggers:
+Two system-based triggers are provided as described in the table below:
 
-```java
-System started
-System shuts down
-```
+Trigger  |  Description
+---------|-------------
+System started | Rules using the 'System started' trigger exeucte when openHAB starts and when changes or additions to *any* rule are recognized by openHab.  This is normal behaivor and should be anticipated when writing rules using this trigger.
```
kaikreuzer: please replace openHab by openHAB :-)

bgilmer77: Absolutely.  Thank you for the correction.  It will take a little while for me to learn the documentation conventions.  I appreciate your comment.

ThomDietrich: "any" might be misleading. As far as I know this is limited to the one file the rule is located in.

bgilmer77: @ThomDietrich On my installation, making a change in an unrelated rules file causes the System started trigger to execute the actions in that file.  I verified this before I wrote this up.  In fact, this behavior is what got me to think that having this behavior documented could be important for users.

ThomDietrich: Wait...

ThomDietrich:
```
2017-06-29 18:15:30.119 [INFO ] [el.core.internal.ModelRepositoryImpl] - Refreshing model 'astro.rules'
2017-06-29 18:15:38.709 [INFO ] [e.smarthome.model.script.astro.rules] - SYSTEM STARTED TEST
2017-06-29 18:16:03.354 [INFO ] [el.core.internal.ModelRepositoryImpl] - Refreshing model 'kodi.rules'
```

@kaikreuzer what do you think?

kaikreuzer: It should definitely only executed the ones within the refreshed files. So my expectation seems to be in line with @ThomDietrich's observations.

AngelosF: In my setup, I don't use `when System started` condition in my rules but I would expect to be triggered only on the file that I am creating/editing (not globally)... as @ThomDietrich described.

Ps: small spell checks should be done: `exeucte` & `behaivor`

bgilmer77: @AngelosF, thank you for your comment.  I have a question - do you want to use U.K. spelling or U.S. spelling in your documentation?  Execute is definitely spelled incorrectly, and I will fix that.  But behavior is spelled correctly in the U.S., but U.K. spelling would be behaviour.  What is the preferred spelling for this project?

bgilmer77:
@ThomDietrich I will do some more testing.  I did two brief checks just now and the System started trigger did NOT fire when I changed unrelated rules - it worked as expected.  If I am unable to duplicate the case where the trigger fires on unrelated rules, I will revise my contribution accordingly.

bgilmer77:
Verified - but slightly different cause...  I just created a home.item file which did not exist before.  When the system read the new file, the System started trigger fired.  I was not able to make the System started trigger fire with edits to any existing rules files.  I also was not able to make the System started trigger fire by creating a new rule.

bgilmer77:
Update - I just found that deleting a line in a .item file causes the 'System started' trigger to fire as well.

bgilmer77:
@ThomDietrich Update - don't know if this is useful, but I continue to find situations where the 'System started' trigger fires.  Here is another one - I enabled a new channel on a multisensor Thing and then manually linked the channel to an Item.  I did all of this in the paper-ui.  When I saved the change, the 'System started' trigger fired.

ThomDietrich:
@bgilmer77 interesting findings. The whole behavior will hopefully clear up when the new feature mentioned by Kai in https://github.com/openhab/openhab-docs/issues/343#issuecomment-310859914 is available.

For now let's finish this PR with a more general description. *System started* is triggered after openHAB startup, after the rule file was modified or after items related to that rule file are modified. The last might not be completely correct but will make users aware of the possibility. All of this should not be a problem if the rule actions themselves are safe. Which brings me to the last point:

Would you like to add a paragraph discussing possible use cases of the system started trigger? In my opinion the trigger should be used to initialize items **if** they are not yet set. This is interesting if they are used inside your sitemap or rules.

Here is a typical system started rule of mine:

```java
rule "Speedtest init"
when
    System started
then
    createTimer(now.plusSeconds(30)) [|
        if (Speedtest_Summary.state == NULL || Speedtest_Summary.state == "") Speedtest_Summary.postUpdate("unknown")
    ]
end
```

ThomDietrich:
Thanks!

configuration/rules-dsl.md
AngelosF: behavior or behaviour is fine... It doesn't really matter U.S. or U.K. spelling.
`behaivor` (as currently written) is misspelled :)
I am a Greek citizen that was educated in the U.S., so I prefer U.S. spelling (but don't mind me... i am just a simple user of OH :))

PR #435, opened 2017-07-07
Added note about automatic java install

BClark09:

### DIFF
--- a/installation/linux.md
+++ b/installation/linux.md
@@ -147,7 +147,7 @@ sudo apt-get install openhab2
 ```
 
 **Note:** Apt will try to install Java 8 on your system if it does not detect a Java package.
-If you have installed one manually and want to continue without installing a package, add `--no-install-recommends` after `install`:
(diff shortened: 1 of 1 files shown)


ThomDietrich:
Do you want to merge this PR right away, even if these instructions are for latest snapshots?

BClark09:
What's the best practice? To only have documentation relating to the most common release (stable) or say "From version 2.2.0..." I'm inclined to think that the latter would just confuse people into thinking "There's a new version?!"

ThomDietrich:
Maybe.
We were actually already discussing (in another thread) to add different version snapshots of the documentation. This is not yet implemented and might take a few more days...
The change in this PR is rather uncritical but we should keep that in mind.
Let's merge.

BClark09:
Ah, that sounds good! Will have a look for that. 

> Last comment: Is the ":" at the end a typo?

Yes :stuck_out_tongue:

ThomDietrich:
Full Stop ^^

PR #438, opened 2017-07-11
[demo] Deprecate an old demo page. Fix non-existent link.

130s:
With the old link I get the following error on my browser. I'm not sure the new link in this commit/PR is what we want though. If someone tells me the right url I can update this PR:

```
HTTP ERROR 404

Problem accessing /job/openHAB-Distribution/lastSuccessfulBuild/artifact/distributions/openhab-offline/target/openhab-offline-2.0.0-SNAPSHOT.zip. Reason:

    Not Found
```


ThomDietrich:
The demo tutorial was removed from the menu structure long ago because it was heavily outdated. The basic tutorial was supposed to be a replacement but even this one could use some refinement.

I did not know that there was still a link pointing to the article. Would both of you agree to remove the link and article from documentation?

AngelosF:
sounds good to remove the http://docs.openhab.org/tutorials/demo.html and the link from http://docs.openhab.org/installation/#getting-started

I did a quick [check](https://goo.gl/CieCn5) and that's the only place that I found the demo.html referenced.

130s:
>> How did you navigate from docs.openhab.org to the demo tutorial?

[Today Google still shows this page as the first result for me](https://www.google.com/search?client=ubuntu&channel=fs&q=openHAB+demo+install&ie=utf-8&oe=utf-8).

>> I like a lot the http://docs.openhab.org/tutorials/beginner/index.html and I don't know if we need the demo
> 
> I did not know that there was still a link pointing to the article. Would both of you agree to remove the link and article from documentation?

+1. That new? series of tutorials look good basically.

I've added a commit to deprecate the demo page and remove the undesired link.

AngelosF:
well... if it's the first result on google, we should keep it :)
I would prefer a redirect from demo to the beginner tutorial (instead of adding a disclaimer that the page is deprecated).
I don't know how to implement the redirect in these pages :( so I can't recommend a method

ThomDietrich:
In this case I'd delete the article completely. Google will remove it soon enough after.
A redirect would also be possible: https://raw.githubusercontent.com/openhab/openhab-docs/gh-pages/features/addons.md

130s:
In general I'd prefer not to removing even pages that contain only obsolete info for archiving purpose, but in this case I'll try to migrate information to existing tutorial pages in #439 as much as possible, so +1 for deleting the old demo page, which I just added a commit for.

ThomDietrich:
Thanks for your help 🍻 
Looking forward to your additions in #439!

PR #469, opened 2017-08-05
Adding description on using sendCommand/postUpdate fixes #185

### LINKED ISSUE #185: Rules: sendCommand/postUpdate actions warning
Add section to Rules to document warnings about using primitives with sendCommand/postUpdate actions

See https://community.openhab.org/t/sendcommand-documentation/9534/9

Question, that posting has a lot of background and generalizations to explain why using primitives with the command/update Actions is a problem. Should they be kept or should the docs update only include a succinct warning to watch out?

LightIsLife:
The use of sendCommand and postUpdate in rules is explained with an emphasis on discouraging the generic method sendCommand(Item,arg) over the method Item.sendCommand(arg) that provides much more reliable type conversions. Details as to the background have been added
Signed-off-by: Markus Lipp <lipp.markus@gmail.com> (github: LightIsLife)

### DIFF
--- a/configuration/rules-dsl.md
+++ b/configuration/rules-dsl.md
@@ -179,7 +179,30 @@ end
 ```
 
 
-## Manipulating Item States
+## Scripts
+
+The expression language used within scripts is the same that is used in the Xtend language - see the [documentation of expressions](http://www.eclipse.org/xtend/documentation/203_xtend_expressions.html) on the Xtend homepage.
+
+The syntax is very similar to Java, but has many nice features that allows writing concise code.
+It is especially powerful in handling collections.
+What makes it a good match for openHAB from a technical perspective is the fact that there is no need to compile the scripts as they can be interpreted at runtime.
+
+To be able to do something useful with the scripts, openHAB provides access to
+
+- all defined items, so that you can easily access them by their name
+- all enumerated states/commands, e.g. `ON, OFF, DOWN, INCREASE` etc.
+- all [standard actions](https://github.com/openhab/openhab/wiki/Actions) to make something happen
+
+Combining these features, you can easily write code like:
+
+```java
+if (Temperature.state < 20) {
+    Heating.sendCommand(ON)
+}
+```
+
+{: #manipulating-item-states}
+### Manipulating Item States
 
 Rules are often used to manipulate the state of an Item, for example switching lights on and off under certain conditions.
 Two commands can change the value or state of an Item within rules:
@@ -200,7 +223,7 @@ The following table summarizes the impact of the two manipulator commands on the
 Besides the specific manipulator command methods `MyItem.sendCommand(<new_state>)` and `MyItem.postUpdate(<new_state>)`, generic manipulators in the form of `sendCommand(MyItem, <new_state>)` and `postUpdate(MyItem, <new_state>)` are available. The specific versions is normally recommended.
 
 {: #sendcommand-method-vs-action}
-### MyItem.sendCommand("new state") versus sendCommand(MyItem, "new state")
+#### MyItem.sendCommand("new state") versus sendCommand(MyItem, "new state")
 
 Using the methods `MyItem.sendCommand(<new_state>)` and `MyItem.postUpdate(<new_state>)` is often preferable.
 These are methods of Objects that can accept a variety of types.
@@ -232,63 +255,6 @@ For example, if the name of the Item to receive an update or command was calcula
 ```java
 val int index = 5
 sendCommand("My_Lamp_" + index, ON)
-```
-
-#### Details:
-
-As all object-oriented computer languages, Java and the Rules DSL have implemented the concept of inheritance. 
-However, inheritance only applies to Objects and does **not** apply to primitives. 
-Inheritance allows to take an existing Object type, called a Class, and adding to it to make it into something different. 
-This “something different” becomes a Child of the original Class, the parent. The Child still can do everything the parent could do. 
-The top level base Class for all Objects in Java and the Rules DSL is called simply `Object`. 
-
-In addition to other useful things, the class `Object` implements a method called `toString`. 
-And since `Object` is the parent of all Objects, ALL Classes also implement a `toString` method. 
-_However primitives do not inherit from Object. 
-They don't inherit from anything and they don't have any methods at all which includes the lack of a toString Method._
-
-The `sendCommand` is a generic action and needs to be able to work with all Item types. 
-Actions only support two String arguments as all Objects will support the conversion `toString`. 
-`sendCommand (MyItem, new_state)` will automatically use the `MyItem.toString` method to convert MyItem into a String. 
-It will also attempt to do so with the second argument if `new_state` is not already a String. 
-However, if the second argument is a primitive, and not an Object, it does not carry a method `toString`. 
-Thus, Rules DSL will not be able to cast `new_state` as a String. 
-As a consequence, the use of `sendCommand(MyItem, primitive)`, using a primitive as the second argument, will almost always fail. 
-
-The different syntax for the generic and the objective-specific differs and is given in the table below:
-
-| Generic (Action)                 | Specific (Method)               |
-|----------------------------------|---------------------------------|
-| `postUpdate(MyItem, new_state)`  | `MyItem.postUpdate(new_state)`  |
-| `sendCommand(MyItem, new_state)` | `MyItem.sendCommand(new_state)` |
-
-`MyTimes.sendCommand()`, however, will use the `sendCommand` method that is suitable to make the necessary type conversions.
-For example, the `NumberItem` class would have a `sendCommand(int)`, `sendCommand(long)`, `sendCommand(float)`, `sendCommand(double)`, `sendCommand(Number)`, `sendCommand(DecimalType)`, and `sendCommand(String)` method. 
-Each of these separate methods is individually written to handle all of these different types of Objects. 
-MyItem will automatically apply the method that corresponds to the argument type.
-
-In a nutshell, using the syntax `MyItem.sendCommand(new_state)` or `MyItem.sendUpdate(new_state)` will help avoid many problems.
-
-## Scripts
-
-The expression language used within scripts is the same that is used in the Xtend language - see the [documentation of expressions](http://www.eclipse.org/xtend/documentation/203_xtend_expressions.html) on the Xtend homepage.
-
-The syntax is very similar to Java, but has many nice features that allows writing concise code.
-It is especially powerful in handling collections.
-What makes it a good match for openHAB from a technical perspective is the fact that there is no need to compile the scripts as they can be interpreted at runtime.
-
-To be able to do something useful with the scripts, openHAB provides access to
-
-- all defined items, so that you can easily access them by their name
-- all enumerated states/commands, e.g. `ON, OFF, DOWN, INCREASE` etc.
-- all [standard actions](https://github.com/openhab/openhab/wiki/Actions) to make something happen
-
-Combining these features, you can easily write code like:
-
-```java
-if (Temperature.state < 20) {
-    Heating.sendCommand(ON)
-}
 ```
 
 {: #implicit-variables}
--- a/configuration/rules-dsl.md
+++ b/configuration/rules-dsl.md
@@ -178,6 +178,49 @@ then
 end
 ```
 
+
+## Causing effects on items
+
+Rules come to life by using input to cause something else to happen, to cause effects, such as change the state of an item and, e.g., switch on a light at sunset. The two most common commands that can change item values or states within rules are `sendCommand` and `postUpdate`. The full syntax to use them is: 
+* `myItem.postUpdate(state_or_value)`
+* `myItem.sendCommand(state_or_value)`
+
+A `postUpdate` is often used to change the status of in item without causing any action to reflect any status changes that may be caused by other means (physically switch on a light). A `sendCommand` will cause a change in status and trigger potential actions. 
+
+It is important to understand that they act slightly different and interact in specific ways with (event-based rule triggers)[{{base}}/configuration/rules-dsl.html#event-based-triggers]. For example `postUpdate` will will fire `received update`-based triggers, but not fire `received command`-based triggers.Correspondingly, `sendCommand` will fire `received command`-based triggers, but not fire `received update`-based triggers.
+
+### For those who want to have more background, here the long answer:
+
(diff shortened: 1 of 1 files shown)


LightIsLife:
I am not sure about the etiquette here, but certainly want to raise this point: given the fact that leaned quite some text and concepts from @rkoshak (see link he posted in issue #185), should that be acknowledged here and if so how?
I saw in other posts (c3a1a1dc7bdaad61dedab4a3b6d3a1c029b59163) that there was a line that reads something like `also-by`, is this the way to go here? Or any other way? Either way I would be totally happy to acknowledge any any all from whom I learned, @rkoshak of course on the top of the list.

rkoshak:
From a personal perspective, I think any acknowledgments should certainly be brought up in the Issues, PR comments, and discussions here on Github but not in the docs themselves. I like sighting sources because it helps tell those who care what source materials were used to build the new text. It can also help the maintainers judge the writing itself. Just be sure to include both the author and the source (forum postings in this case).

If direct quotes are used in the PR itself then attribution should be supplied with the direct quote, just like one would do in a paper.

I'm not sure these acknowledgments need to be all that formal though. I think the also-by tag has a specific meaning in the signoff so I can't say that it is the appropriate way to make the ack.

rkoshak:
Some comments on the PR itself:

Line 184: Consider rewording:

Rules come to life by **generating events** to cause something else to happen, to cause effects**. For example**, change the state of an item or switch on a light at sunset.

Line 188: Consider rewording:

A `postUpdate` is used to change the status of an item without causing any further action. `postUpdate` is used to update openHAB Items to reflect any status changes that may be caused by other means (physically switch on a light) or to indicate state changes that do not require a corresponding action. 

Line 190: Missing a space between . and Correspondingly

Line 192: This section contains information that probably shouldn't be skipped (i.e. there is a preference to use the methods rather than the Actions. I would change the title to something like:

MyItem.sendCommand("new state")  versus sendCommand(MyItem, "new state")

Add an intro paragraph to explain that the that the method is the preferred method. The current opening lines make the reader think that this has already been stated but it hasn't . In this intro include a sentence or two to mention that the Actions can only accept Strings as arguments versus the methods that can take a variety of types depending on the Item.

Line 202: Consider rewording slightly to emphasize when it is most appropriate to use which approach:

- Actions: when all you have is the name of the Item as a String instead of a reference to the actual Item itself (e.g. when the name of the Item to receive an update or command was calculated in the Rule by building up a String).
- Methods: when you have the reference to the Item directly because its name is hard coded or the Item was pulled out of a Group.

Don't lose the rest of this paragraph though. This is just a quick intro or a tl;dr for the next section but my fear is that if it isn't included here many readers will skip over this info in the next section.

Line 204: Get rid of the "Even More" from the title, assuming you made the change to 192

Line 206: Inheritance allows **one** to

Line 210: work with all **Item types**.

Finally, this is probably beyond the scope of this PR, but I want to capture somewhere that when/if we bring in the article started [here](https://community.openhab.org/t/type-conversions/32684), right after this section is probably where it should go.

Great work! Thanks for doing this. It is really appreciated.

LightIsLife:
Great comments, thanks so much. Incorporated all. I apprecaite your feedback.
Are the attributions above (and in the issue #185 enough to address your comments? I am happy to do whatever is appropriate or necessary. As mentioned above, this commit is based on the contribution by @rlkoshak here: https://community.openhab.org/t/sendcommand-documentation/9534/9; I don't think I quoted verbatim from this though.  

@rkoshak as for your other comment:
> Finally, this is probably beyond the scope of this PR, but I want to capture somewhere that when/if we bring in the article started here, right after this section is probably where it should go.

I could give this a try. I really like your forum entries, and it would be great if we could capture this one. How is this started, do you want to raise an issue and assign it to me? (or not, but as said, I think I can do that if desired).

rkoshak:
Well, I wrote that article at a prompting from Thom. One of the approaches we have used for creating new parts of the docs (i.e. those not copied from somewhere else like the OH 1.x wiki) is to create and update a posting on the Forum. Once the document is mature enough we then incorporate it into the docs.

This gives us a good place to get content out fast and reviewed by a wider audience. This worked well for the Migration Tutorial.

The article there is not quite ready to be added to the docs. There are at least three Item types and at least one (maybe more) State Types that are not yet written up:

Item         State
----------------------
Location  PointType
Call          ???
Image      ???

Go ahead and create a new issue, point it at that article and that way we will have the action captured and tracked. Then feel free to start/continue a discussion over at the thread. Once we are happy the content is complete we will copy it over to the Docs and complete the Issue. 

Sound good?

LightIsLife:
@rkoshak just having second thoughts about whether I interpreted your suggestions correctly. 
You suggested to use 
> MyItem.sendCommand("new state") versus sendCommand(MyItem, "new state")

more specifically I am referring to your suggestion to use `"new state"` while originally I had used `state_or_number`; Now I had no issues either way and updated my committ following your nomenclature suggestion. 
Where I am starting to have question is the use of quotation marks in this context: `MyItem.sendCommand("new state")` 
Considering that it would be rather common practice to think that everything in quotation marks refers to a string, makes this not wrong, but maybe too restricted; if I understand it correctly the argument here does not need to be a string, but could be a primitive or any Object (depending on the Item of course). Would it be better to write:
`MyItem.sendCommand(new state)` (no use of quotation marks) of course we would leave the quotation marks for actions as only strings can be used here: `sendCommand(MyItem, "new state")`

Or am I overthinking this?

rkoshak:
You might be over thinking it. I honestly didn't put that much thought into it myself. I think my main point was that I was trying to be consistent in showing that the same thing is being passed to the Action and the method.  I used the quotes in Action to emphasize that we are passing a String and then I used the quotes in the method to be consistent with the Action.

But it might be more instructive to show passing an Object to the method and the Object.toString to the Action. The important thing to capture is that the Action ONLY takes Strings. Sometimes the Rules engine is able to convert an Object to a String but often it fails to do so.

Maybe it would be better to use something like `MyItem.sendCommand(newState)` and `sendCommand(MyItem.toString, newState.toString)`. The toString isn't strictly required for MyItem but it emphasizes that the Action is taking a String, not the Item object itself. The toString on the newState helps to emphasize that the Action is taking a String for the state.

LightIsLife:
Thanks much, appreciate you sharing your thoughts.
Maybe we'll leave it for now, not sure what a better way could be. The new format that you suggested just now would be, for me, just adding confusion. In the text we right that it is primitives do not have such a .toString method and the we use it in argument......
For me it would be clearer to leave it as a "string" at least for the actions.  
And you reminded me with your comment:
> The toString isn't strictly required for MyItem but it emphasizes that the Action is taking a String, not the Item object itself. The toString on the newState helps to emphasize that the Action is taking a String for the state.

that we should add a sentence explaining that in an action the MyItem will be automatically case using the .toString method that is available in all Items; but of course the user can make this transformation manually too use a string with the name of the item in the action. I am not sure this si sufficiently clear hear for users not totally familiar with OOP.

LightIsLife:
@ThomDietrich appreciate your comments/review/input, thank you. 
Attempted to address everything in new commit d824bec

ThomDietrich:
Did you forget the one-sentence-per-line thingy?

LightIsLife:
Hmm, not sure what I missed.....let me check tomorrow

LightIsLife:
@ThomDietrich: I think I got it now....sorry long and convoluted writing is sort of an occupational hazard for me....hope I got it shortened enough. If not let me know what the desired word/sentence or syllables/sentence metrics is for the docs and I will rewrite. Thanks for your guidance.

ThomDietrich:
Ooooh I'm deeply sorry, I thought you already know about this requirement of mine / of our young evolving [style guide](https://github.com/openhab/openhab-docs/blob/gh-pages/styleguide.md). Each sentence has to be on its own line in the document. That's it, no need to shorten sentences.

Example: 

```
This is example text file content.
Every sentence is on it's own line.
Why you ask?
Git works on a per line basis, you can see in your last commit why this is not always best to see changes in a long line (e.g. a full paragraph of text).
Breaking down the whole text to one sentence per line is a good and established method do deal with that.
But what about the result? No worries. Markdown ignores single line breaks and rebuilds your original paragraph running text.
```

LightIsLife:
No worries, at least I got rid of the run on sentences 😄 😉 
thanks for your explanation, and apologies: markdown and git newbie here and didn't think it through, of course it makes sense giving that git is so line oriented (and awesome)
It will certainly help for the next edits, I started to struggle with further shortening sentences.....

ThomDietrich:
Hey @LightIsLife, I had a few minutes to spare and added my own set of improvements. Please check all of them and feel free to change whatever you don't like! I hope everything is still in line with your idea and you are not mad for messing around in your work.

I've three more things to discuss:

1. I'd move the whole block below "Scripts", maybe in front of "Implicit Variables inside the Execution Block". wdyt?

2. You wrote "Actions can only accept Strings as arguments" and I guess I understand that in the wrong way. `sendCommand(Heating, ON)` is a perfectly fine example without a string argument. @rkoshak could you elaborate on your original suggestion?

3. To me the whole "Details" chapter seems to be overly ...well... detailed and overdone for the documentation. I'd vote to leave it out. Maybe you can post it in the community to be found be interested users?

LightIsLife:
@ThomDietrich thanks for your edits. Make all sense to me and if it helps improving the documentation, I am all for it. 
* re 1) I am rather agnostic and happy to move it if you feel strongly about it. I put it where it made sense for me; but in the end it has to make sense to all of us
* re 2) I would like to believe that ON is of the OnOffType and therefore an Object; which in turn comes with a toString method that is automatically called......now I am no expert here, but that's how it made sense to me
* re 3) Again, if you feel strongly about, OK, but I know from myself and from the forum that type conversion/casting is an ever confusing topic, which I am only able to understand if I have the knowledge that is in the details section; as discussed elsewhere and captured in issue #472 I have been working with @rkoshak on a follow-up chapter for type conversion. Not sure this would make as much sense to everyone if we dropped the details section. In #472 you can find a link to the wiki entry in the forum, where I believe we capture almost all we wanted to capture. The wiki is too long and has too many run on sentences, but I am happy to put more work in it, when we agreed on the general direction.

rkoshak:
> @rkoshak could you elaborate on your original suggestion?

The Actions are both defined as sendCommand(String, String) and postUpdate(String, String). Sometimes the Rules Engine is smart enough to automatically convert SOME objects to String automatically for you. The problem arises in those cases where the Rules Engine fails to do so.

I've yet to find any pattern for when the Rules Engine fails to do the conversion. This admittedly slightly incorrect statement that the Actions only take Strings was a form of defensive documentation. In short, if you always explicitly pass Strings it will always work. If you rely on the Rules engine to call toString on the passed in objects it will sometimes fail.

> To me the whole "Details" chapter seems to be overly ...well... detailed

This section is useful but perhaps a little too low level for this part of the document. I'd hate to lose it as I often link back to the original post that @LightIsLife got it from but maybe it would be possible to post it as a separate page with a link or as a "click the + for more" box or something. 

I agree, it is not necessarily required information for all users. But when users run into type problems, understanding why they are having problems is helpful.

Maybe it should go in the FAQ?

LightIsLife:
Maybe this helps illustrate the point why this whole thing can be confusing for the less initiated:
```
rule "OnOff"
when Item Date received update
then
var OnOffType test_OnOff
test_OnOff = ON

logInfo("OnOff", " OnOff.toString: "+test_OnOff.toString)
logInfo("OnOff", "OnOff Class: "+test_OnOff.class.toString)
if (test_OnOff.equals(1)) {logInfo("test","ON equals 1")} else {logInfo("test","On equals NOT 1")}
if (test_OnOff.equals("ON")) {logInfo("test","ON equals String \"ON\"")} else {logInfo("test","On equals NOT String \"ON\"")}
if (test_OnOff.equals(ON)) {logInfo("test","But ON equals Object ON")}
end
```
the output is as expected:
```
2017-08-24 21:16:53.908 [INFO ] [eclipse.smarthome.model.script.OnOff] -  OnOff.toString: ON
2017-08-24 21:16:53.909 [INFO ] [eclipse.smarthome.model.script.OnOff] - OnOff Class: class org.eclipse.smarthome.core.library.types.OnOffType
2017-08-24 21:16:53.909 [INFO ] [.eclipse.smarthome.model.script.test] - ON equals NOT 1
2017-08-24 21:16:53.910 [INFO ] [.eclipse.smarthome.model.script.test] - ON equals NOT String "ON"
2017-08-24 21:16:53.911 [INFO ] [.eclipse.smarthome.model.script.test] - But ON equals Object ON
```
But from own experience I can only emphasize: it is not immediately intuitive......

rkoshak:
> But from own experience I can only emphasize: it is not immediately intuitive......

I agree for non-programmers this behavior is not intuitive. However, I think Thom's point is does this information belong **here** in the docs, or is it better placed somewhere else? To a large extent, the content doesn't quite fit where we have it placed here and it might be more of a distraction or cause more confusion to have it here than to have it somewhere else.

To some extent, it is a topic at the same level as the Actions versus Methods that this section is addressing rather than a subsection. Perhaps it would go better with the Type Conversion article because all this information is equally relevant, if not more so, to that article and personally, if I were to try to browse to this content I would look there first.

LightIsLife:
Sorry, reopening; was closed by user error 😊

LightIsLife:
@rkoshak @ThomDietrich thanks again.
Makes now all sense to me (Finally :smile: ) 
Moved all under the sections Scripts and took out the whole proposed "Details" section. 
Will keep it and attempt to reintroduce the topic when drafting the type conversion part

LightIsLife:
Actually, would you consider it to be appropriate to add a sentence to clarify the situation re: mandatory string use in actions. I would add it somewhere in parenthesis and it should read: "(The full explanation for experts: Actions can be used with arguments whose types can be auto-converted to a String through Rules-DSL. All other types will make the Action fail)"
I was hoping to clarify the rather limited statements that are currently in the text for those who know a little more. For everybody else, I believe the guidance "just use strings" may suffice.

rkoshak:
I'd be more inclined to leave out the parenthetical primarily because even that statement isn't entirely correct or sufficient. The problem is the Rules Engine is not consistent in what types it does and does not automatically convert to Strings and I don't have the full list of all the types that won't work. Put another way, all Objects SHOULD work just fine but many Object do not.

So rather than muddy the waters while trying to make it clear I'd just leave that out and leave the users with the impression that toString is the only way to go. It may not be correct but it will never go wrong and it requires far less text in the docs to explain.

LightIsLife:
Works for me....I did not add it yet to the PR, but I thought I'd bring it
up; thanks Rich!

ThomDietrich:
For reference, the deleted details can be found here: https://github.com/openhab/openhab-docs/pull/469/commits/23e2a8933b350728750cb77836e2fd26f322ce6a

I agree, that the details can fit nicely in the Type Conversion article, maybe at the very end because of its length? @LightIsLife go ahead and move it there right away, before we forget about that.

LightIsLife:
Thanks to both of you for helping me through my first PR!!!

ThomDietrich:
Ah right! Congrats on that 😉 

Btw. a good read: http://www.firsttimersonly.com (follow the links)

PR #470, opened 2017-08-06
Clarified that item is mandatory for a group

kaikreuzer:


kaikreuzer:
Fixes https://github.com/openhab/openhab2-addons/issues/2523.

PR #486, opened 2017-08-31
Improved the SCALE transformation article

### LINKED ISSUE #484: Fix syntax highlighting for the SCALE transformation article
(Triggered by https://community.openhab.org/t/item-calculation-m-s-to-km-h/33758/7)

http://docs.openhab.org/addons/transformations/scale/readme.html

It would be better to transform this into a list rather than a code block. Also the examples are rather poor and unclear. Please feel free to grab this issue.

Confectrician:
- Replaced the temperature example with the UV index example from OH1 wiki
- Transferred the allowed formats from a codebox to a table
- Added some Explanations, to give the article a common thread

Fixes #484

Signed-off-by:  Jerome Luckenbach <github@luckenba.ch> (github: @Confectrician)


_transformations/scale/readme.md
```diff
@@ -14,17 +14,43 @@ install: auto
...
-]12..=Any temp above 12
+example item:
+
+```java
+Number Uv_Sensor_Level "UV Level [SCALE(uvindex.scale):%s]"
+```
+
+uvindex.scale in transform folder:
+
+```
```
ThomDietrich: Every code block should have a language. See: http://www.cirosantilli.com/markdown-style-guide/#option-code-fenced

Will also look nicer:
```python
[0,3]=1
[3,6]=2
[6,8]=3
[8,10]=4
[10,100]=5
```

_transformations/scale/readme.md
```diff
@@ -14,17 +14,43 @@ install: auto
...
+[3,6]=2
+[6,8]=3
+[8,10]=4
+[10,100]=5
 ```
+
+Each value the item receives, will be categorized to one of the five given ranges.
+The only condition here is that the received value has to be between `0` and `100` in our example, since we haven't defined other cases yet.
+
+## Allowed formats
```
ThomDietrich: *Formats

_transformations/scale/readme.md
```diff
@@ -14,17 +14,43 @@ install: auto
...
+
+uvindex.scale in transform folder:
+
+```
+[0,3]=1
+[3,6]=2
+[6,8]=3
+[8,10]=4
+[10,100]=5
 ```
```
ThomDietrich: Which is it, `[10,100]` or `[10..100]`

Also how about showing the "smaller than 3" and "bigger than 10" logic from below?

_transformations/scale/readme.md
```diff
@@ -14,17 +14,43 @@ install: auto
...
+## Allowed formats
+
+This table shows the 4 allowed formats for creating scale expressions.
+
+| scale expression | result |
+|--------------------|--------|
+| `[12..23]=XYZ` | `Returns XYZ when the given value is between (or equal to) 12 and 23` |
+| `]12..23[=XYZ` | `Returns XYZ when the given value is between 12 and 23. 12 and 23 are excluded in this case.` |
+| `..23]=XYZ` | `Returns XYZ until the given value is lower than or equal to 23` |
+| `]12..=XYZ` | `Returns XYZ until the given value is greater than 12` |
```
ThomDietrich: Shouldn't that table come before the example?

_transformations/scale/readme.md
```diff
@@ -14,17 +14,43 @@ install: auto
...
+
+With those different expression formats you could add some edge cases to our example from above and catch sensor values above 100 with `]100..=Out of UV index range` for example.
```
ThomDietrich: Careful with an example where we mix numbers and strings as return value. This might work on a correctly configured Sitemap element but it's definitely not how we should motivate the usage...

_transformations/scale/readme.md
ThomDietrich: I think this table would be much cleaner if you could remove the redundant "Returns XYZ when the given value is"

_transformations/scale/readme.md
```diff
@@ -14,17 +14,43 @@ install: auto
 
 # Scale Transformation Service
 
-Transform the input by matching it between limits of ranges in a scale file.  The input string must be in numerical format.
+Transform a given input by matching it between limits of ranges in a scale file.
```
ThomDietrich: I feel the sentence would be easier to understand without the ending "in a scale file"

ThomDietrich:
Heyho! Wow this was quick! 🍻

_transformations/scale/readme.md
Confectrician: Same procedure as every time...
I copied that from the wiki and didn't check it. :D

Which language did you choose for that?

_transformations/scale/readme.md
Confectrician: I will check that in my installation and rewrite it then.

For the other logic, i will comment below.

_transformations/scale/readme.md
Confectrician: Maybe this sentence would be better placed as column header.

Confectrician:
Was a bit of luck for that quick start.
I looked at the article and the old wiki page and had an idea/concept in mind within seconds.
So the start was relative easy for that issue.

There still will be something to discuss and it will last some time.
The earlier we start, the earlier we can finish that issue. :)

_transformations/scale/readme.md
ThomDietrich: I chose python, simply because I know that *lists* are defined with square brackets.

Python:
```python
[0,3]=1
[3,6]=2
[6,8]=3
[8,10]=4
[10,100]=5
```

Java:
```java
[0,3]=1
[3,6]=2
[6,8]=3
[8,10]=4
[10,100]=5
```

PHP:
```php
[0,3]=1
[3,6]=2
[6,8]=3
[8,10]=4
[10,100]=5
```

Yeah... I think it doesn't really matter ;)

ThomDietrich:
Sure thing. Just mixed a few comments in myself. No bad intentions 🤹‍♂️

Confectrician:
I have changed to the correct syntax. (The OH1 wiki is definitely outdated here.)
Also i have changed the table header.
I changed the old wiki example a bit to include some of the allowed formats and gave some hints about them.

_transformations/scale/readme.md
```diff
@@ -14,17 +14,44 @@ install: auto
...
+]3..6]=2
+]6..8]=3
+]8..10]=4
+[10..100]=5
 ```
+
+Each value the item receives, will be categorized to one of the five given ranges.
+Values **lower than or equal to 3** are catched with `..3]=1`.
+Greater values are catched in ranges with 2 values as criteria.
+The only condition here is that the received value has to be lower than or equal to `100` in our example, since we haven't defined other cases yet.
```
ThomDietrich: What happens if none of the scale conditions match? What will be shown (on a sitemap for example), if the received value would be 101?

_transformations/scale/readme.md
```diff
@@ -14,17 +14,44 @@ install: auto
...
+..3]=1
+]3..6]=2
+]6..8]=3
+]8..10]=4
+[10..100]=5
 ```
+
+Each value the item receives, will be categorized to one of the five given ranges.
+Values **lower than or equal to 3** are catched with `..3]=1`.
+Greater values are catched in ranges with 2 values as criteria.
```
ThomDietrich: Are the lines evaluated from top to bottom or do they have to mutually exclusive? Would this example for temperatures be allowed?

```python
[15..25]="okay"
[5..35]="soso"
[-10..45]="nope"
[-100..100]="you die"
```

_transformations/scale/readme.md
```diff
@@ -14,17 +14,44 @@ install: auto
 
 # Scale Transformation Service
 
-Transform the input by matching it between limits of ranges in a scale file.  The input string must be in numerical format.
+Transform a given input by matching it between limits of ranges.
+The input string must be in numerical format.
```
ThomDietrich: What about the output?

_transformations/scale/readme.md
```diff
@@ -14,17 +14,44 @@ install: auto
...
+|--------------------|--------|
+| `[12..23]=XYZ` | `between (or equal to) 12 and 23` |
+| `]12..23[=XYZ` | `between 12 and 23 (12 and 23 are excluded in this case.)` |
+| `..23]=XYZ` | `lower than or equal to 23` |
+| `]12..=XYZ` | `greater than 12` |
 
 ## Example
 
-transform/temps.scale:
+The following example shows how to break down numeric UV values into fixed UV index categories.
```
ThomDietrich: Do you think the example is easy to grasp? Maybe a short description or a wikipedia link would clarify things.

_transformations/scale/readme.md
```diff
@@ -14,17 +14,44 @@ install: auto
...
+
+example item:
+
+```java
+Number Uv_Sensor_Level "UV Level [SCALE(uvindex.scale):%s]"
+```
+
+uvindex.scale in transform folder:
```
ThomDietrich: This and "example item" would as sentences? They should at least start with an uppercase letter.

ThomDietrich:
@Confectrician besides my small comments I have a bigger one. You might have seen the comment warning authors to not edit this file directly. That is in fact true :)

I'm afraid we need your PR over to https://github.com/eclipse/smarthome/tree/master/extensions/transform/org.eclipse.smarthome.transform.scale

Confectrician:
No unfortunately I haven't seen that comment.

For me it would be fine to get a merge ready state here and move it afterwards.

I have to test most of the current comments anyway since I have never worked with scale before and can't answer the questions directly.

What would be the best workflow in your opinion?
I can't say much about it since I hadn't this case yet.

ThomDietrich:
> For me it would be fine to get a merge ready state here and move it afterwards.

Sounds good to me.

Yeah most of these questions are also unclear to me. It's good that we finally figure it out. It might also be interesting to check the sourcecode to learn about hidden treasures :) Let me know if you would like my help!

Confectrician:
Sorry for the delay, i was really busy last weeks. 
I am doing some more docs stuff again in the next weeks.

> It might also be interesting to check the sourcecode to learn about hidden treasures :) Let me know if you would like my help!

I agree and wanted to dig into the code myself anyways at least for the output comment above.
I will call for help, if needed.

Corrently there are merge conflicts, but that doesnt target us anyways, since we want to move to the esh repo later. Correct?

_transformations/scale/readme.md
Confectrician: According to my tests it is not necessary to keep a specific order.
If the given value fits in one of the conditions, it will get scaled.

So my result for this question would be to not mention this.

If ther would have been a specific order we need to follow i would have mentioned it, but since it is not relevant, i see no reason in writing something about it.

_transformations/scale/readme.md
Confectrician: The output is a simple string according to https://github.com/eclipse/smarthome/blob/master/extensions/transform/org.eclipse.smarthome.transform.scale/src/main/java/org/eclipse/smarthome/transform/scale/internal/ScaleTransformationService.java

Should we mention that?

Confectrician:
I have looked at
https://github.com/eclipse/smarthome/blob/master/extensions/transform/org.eclipse.smarthome.transform.scale/src/main/java/org/eclipse/smarthome/transform/scale/internal/ScaleTransformationService.java

and didnt found any "hidden treasures" on the first look.
Maybe you can have a look too @ThomDietrich.
I think we can move those changes to the ESH Repo in the nearer future then.

_addons_transformations/scale/readme.md
```diff
@@ -14,17 +14,47 @@ install: auto
 
 # Scale Transformation Service
 
-Transform the input by matching it between limits of ranges in a scale file.  The input string must be in numerical format.
+Transform a given input by matching it between limits of ranges.
+The input string must be in numerical format.
 
-The file is expected to exist in the `transform` directory, and should follow the format given in the example below.
+The file is expected to exist in the `transform` directory, and should follow the format given in the table below.
```
ThomDietrich: Please mention, that the file ending has to be scale and that the directory can be found under the configuration path

_addons_transformations/scale/readme.md
```diff
@@ -14,17 +14,47 @@ install: auto
...
+
+| scale expression | returns XYZ when the given value is |
```
ThomDietrich: Please capitalize

_addons_transformations/scale/readme.md
ThomDietrich: Also please add one sentence describing the left and right part of the expression

_addons_transformations/scale/readme.md
```diff
@@ -14,17 +14,47 @@ install: auto
...
+|--------------------|--------|
+| `[12..23]=XYZ` | `between (or equal to) 12 and 23` |
+| `]12..23[=XYZ` | `between 12 and 23 (12 and 23 are excluded in this case.)` |
+| `..23]=XYZ` | `lower than or equal to 23` |
+| `]12..=XYZ` | `greater than 12` |
```
ThomDietrich: Good examples :)

ThomDietrich: We should add a sentence like "Lines are evaluated from top to bottom" - I did not test this! That's how I understood it from sourcecode.

_addons_transformations/scale/readme.md
```diff
@@ -14,17 +14,47 @@ install: auto
...
+Each value the item receives, will be categorized to one of the five given ranges.
+Values **lower than or equal to 3** are catched with `..3]=1`.
+Greater values are catched in ranges with 2 values as criteria.
+The only condition here is that the received value has to be lower than or equal to `100` in our example, since we haven't defined other cases yet.
+If **none** of the configured conditions matches the given value, the response will be empty.
+
+Please note that all ranges for values above **3** are opened with a `]`.
+So the border values (3, 6, 8 and 10) are always transformed to the lower range, since the `]`excludes the given critera.
+
+With the different expression formats you could add an additional range to our example from above and catch sensor values above 100 with `]100..=6` for example.
```
ThomDietrich: You could remove this line. It doesn't add new information and per the example description values above 100 are not possible.

_addons_transformations/scale/readme.md
ThomDietrich: That would also answer my question regarding overlapping ranges

ThomDietrich:
Hey Jerome,
a few more ideas from my side. Shouldn't take long. After that feel free to move to ESH and close here. Thanks!

_addons_transformations/scale/readme.md
Confectrician: >We should add a sentence like "Lines are evaluated from top to bottom" - I did not test this! 

Hm i don't se an effort here, since the order is irrelevant. (See last review comments.)

_addons_transformations/scale/readme.md
```diff
@@ -14,17 +14,47 @@ install: auto
...
 ```
+
+Each value the item receives, will be categorized to one of the five given ranges.
+Values **lower than or equal to 3** are catched with `..3]=1`.
+Greater values are catched in ranges with 2 values as criteria.
+The only condition here is that the received value has to be lower than or equal to `100` in our example, since we haven't defined other cases yet.
+If **none** of the configured conditions matches the given value, the response will be empty.
+
+Please note that all ranges for values above **3** are opened with a `]`.
+So the border values (3, 6, 8 and 10) are always transformed to the lower range, since the `]`excludes the given critera.
```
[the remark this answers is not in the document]
Confectrician: That's not correct.

I have tested it with 
`[0..3]=1`
`[3..6]=2`
and
`[0..3]=1`
`]3..6]=2`
with an item value of `3`.

In the first constellation the item gets scaled to `2`.
Only with the *excluded* bracket, it will get scaled to `1`.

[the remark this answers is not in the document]
ThomDietrich: Really? In this case I'd be interested in the result of

```
[3..6]=2
[0..3]=1
```

also

```

```
[0..7]=3
[3..6]=2
[0..3]=1
[0..7]=4
```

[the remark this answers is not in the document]
Comment (Confectrician) [reply _addons_transformations/scale/readme.md]: Its `1` for your first example and `4` for the last one.

So it literally seems to walk through from top to bottom and uses the last matching range?!

File: _addons_transformations/scale/readme.md [hunk 17]
```diff
@@ -55,6 +60,4 @@ The only condition here is that the received value has to be lower than or equal
 If **none** of the configured conditions matches the given value, the response will be empty.
 
 Please note that all ranges for values above **3** are opened with a `]`.
-So the border values (3, 6, 8 and 10) are always transformed to the lower range, since the `]`excludes the given critera.
-
-With the different expression formats you could add an additional range to our example from above and catch sensor values above 100 with `]100..=6` for example.
\ No newline at end of file
+So the border values (3, 6, 8 and 10) are always transformed to the lower range, since the `]`excludes the given critera.
```
ThomDietrich: space missing

_addons_transformations/scale/readme.md
```diff
@@ -17,9 +17,14 @@ install: auto
 Transform a given input by matching it between limits of ranges.
 The input string must be in numerical format.
 
-The file is expected to exist in the `transform` directory, and should follow the format given in the table below.
+The file is expected to exist in the `transform` directory and its ending has to be `.scale`.
+It should follow the format given in the table below.
 
-| scale expression | returns XYZ when the given value is |
+Range expressions always contain two parts.
+The range to scale on, which is located *left* from the equality sign and the corresponding output string on the right of it.
```
ThomDietrich: why italic?

_addons_transformations/scale/readme.md
```diff
@@ -17,9 +17,14 @@ install: auto
 Transform a given input by matching it between limits of ranges.
 The input string must be in numerical format.
 
-The file is expected to exist in the `transform` directory, and should follow the format given in the table below.
+The file is expected to exist in the `transform` directory and its ending has to be `.scale`.
```
ThomDietrich: "`transform` configuration directory"

_addons_transformations/scale/readme.md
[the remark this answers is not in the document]
ThomDietrich: It seems like it but it is actually not that easy. I've taken a closer look at the source code and found that HashMap is used to store the parsed ranges. A HashMap does not have a guaranteed order. 

>  This class makes no guarantees as to the order of the map; in particular, it does not guarantee that the order will remain constant over time.

Can you confirm that it works multiple times? I'd very much like to guarantee a certain evaluation order in documentation. 😞 

@clinique you are the original developer of the SCALE transformation. Is the above a bug or am I missing something?

ThomDietrich:
Wonderful. Thanks! I think the article is almost finished, the last open issue is https://github.com/openhab/openhab-docs/pull/486#discussion_r145267461
@martinvw could you please add your opinion? Did I miss something or is the evaluation order per definition not guaranteed?

martinvw:
With the evaluation order you are referring whether it takes the first matching item or does not do that? Let me check.

_addons_transformations/scale/readme.md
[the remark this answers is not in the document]
clinique: @ThomDietrich :+1:  yes, we currently can not predict what range will be the first to intercept a given value. I would not say a bug. In my own usage, when I define scale file, I write them in order to have no overlap between ranges so I did not face this problem but in fact predictability in this case seems important, then it would be worth improving the scale transformation.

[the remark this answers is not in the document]
Confectrician: Of course we could mention that the ranges have to be defined in order, but i think an improvement of the scale transformation would be the preferable approach.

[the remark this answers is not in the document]
clinique: Ok, I can take care of it. Will had a specific testing to ensure we have a predictible evaluation order.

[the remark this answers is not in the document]
ThomDietrich: @clinique Thanks for the quick response! I believe using LinkedHashMap would be a quick and painless solution. Additionally could you please reverse the Map to gain a top-to-bottom evaluation order? That would seem the most intuitive to me. Thanks for your consideration!

@martinvw all-clear 😄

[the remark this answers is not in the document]
martinvw: > Ok, I can take care of it. Will had a specific testing to ensure we have a predictible evaluation order.

@clinique did you already start, I was also on the way preparing something :-)

Confectrician:
Hey @ThomDietrich,

Since https://github.com/openhab/openhab-docs/pull/486#discussion_r145267461 is already on a good way, we could already start moving this to a ESH PR. 

Wdyt?

_addons_transformations/scale/readme.md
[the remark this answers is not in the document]
clinique: @martinvw : I was about to go :)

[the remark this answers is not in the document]
martinvw: I also updated some doc-blocks and followed some suggestions of my idea and of the more recent coding guidelines. But I'm busy the rest of the day, so maybe you do the quick fix and I contribute my small changes later?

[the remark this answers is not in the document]
ThomDietrich: Love the enthusiasm! @clinique While we are at it I had another look at the regex expression and found a few more things to fix or improve:

1. Other than claimed in the readme, bound symbols are not optional. The regex asks for either `[` or `]`. The expression `..23]` is not valid, `[..23]` would be valid. We should agree on one version or support both.
2. While we are at it, let's also support round brackets? https://en.wikipedia.org/wiki/Interval_(mathematics)#Classification_of_intervals
3. To catch another potential issue, do you think we should support and trim spaces in between elements?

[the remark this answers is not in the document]
clinique: 1. Yes, the contract as always be to defined range with '[' and ']'. They can not be omitted. The scale transformation relied on guava scale library when I wrote it. So was it.
2. Guava did not '('. From what I read in wikipedia, I think that adding parenthesis can be misleading for a poor value added as '(' <=> ']'. I do not think this is error prone.
3. Yes spaces should be trimed. I'll include some tests on this.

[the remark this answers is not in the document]
clinique: @martinvw : the main problem is not generated by Hashmap but by Properties that is not ordered neither.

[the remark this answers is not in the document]
ThomDietrich: Agreed, Accepted, Appreciated 😉 

@Confectrician you'll need to correct the README regarding 1.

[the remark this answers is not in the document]
martinvw: @clinique that makes it way harder, I might have time tonight, so please mention me on any pull request or if you will not be able to complete it.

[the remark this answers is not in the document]
clinique: @ThomDietrich : regarding point 3, spaces can not be trimmed if we keep on using Properties to read scale definition file because they are not accepted in property key name and are then considered as a delimiter.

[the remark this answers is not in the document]
ThomDietrich: I see. Weird... well then don't worry, we can do without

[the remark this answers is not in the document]
clinique: @martinvw : I had to subclass Properties and use LinkedHashMap.

_addons_transformations/scale/readme.md
```diff
@@ -14,17 +14,50 @@ install: auto
...
+Range expressions always contain two parts.
+The range to scale on, which is located left from the equality sign and the corresponding output string on the right of it.
+A range consists of two bounds. Both are optional, the range is then open. Both bounds can be inclusive or exclusive.
+
+| Scale Expression | Returns XYZ when the given Value is                        |
+|------------------|------------------------------------------------------------|
+| `[12..23]=XYZ`   | `between (or equal to) 12 and 23`                          |
+| `]12..23[=XYZ`   | `between 12 and 23 (12 and 23 are excluded in this case.)` |
+| `[..23]=XYZ`     | `lower than or equal to 23`                                |
+| `]12..]=XYZ`     | `greater than 12`                                          |
```
ThomDietrich: Expressions are evaluated from from top to bottom. The first range that includes the value is selected.

_addons_transformations/scale/readme.md
```diff
@@ -14,17 +14,50 @@ install: auto
...
 
 ## Example
 
-transform/temps.scale:
+The following example shows how to break down numeric UV values into fixed UV index categories.
+We have an example UV sensor that sends us numeric values from `0` to `100`, which we then want to scale into the *UV Index* range.
```
ThomDietrich: remove "us"

_addons_transformations/scale/readme.md
```diff
@@ -14,17 +14,50 @@ install: auto
...
+For further Information on the UV Index take a look at [Wikipedia](https://en.wikipedia.org/wiki/Ultraviolet_index).
```
ThomDietrich: You could remove this sentence and link "UV index" above directly

_addons_transformations/scale/readme.md
```diff
@@ -14,17 +14,50 @@ install: auto
...
+
+Example item:
+
+```java
+Number Uv_Sensor_Level "UV Level [SCALE(uvindex.scale):%s]"
+```
+
+Uvindex.scale in transform folder:
```
ThomDietrich: U in lowercase?

Consider "Referenced scale file `uvindex.scale` in the `transform` folder:"

_addons_transformations/scale/readme.md
```diff
@@ -14,17 +14,50 @@ install: auto
...
-[12..23]=Temp between 12 or higher and 23 or lower
-]12..23[=Temp over 12 and below 23
-..23]=Any temp up to and including 23
-]12..=Any temp above 12
+```python
+[..3]=1
+]3..6]=2
+]6..8]=3
+]8..10]=4
+[10..100]=5
```
ThomDietrich: Leave right-unbounded? (remove "100")

_addons_transformations/scale/readme.md
```diff
@@ -14,17 +14,50 @@ install: auto
...
 ```
+
+Each value the item receives, will be categorized to one of the five given ranges.
+Values **lower than or equal to 3** are catched with `..3]=1`.
```
ThomDietrich: Wrong example

_addons_transformations/scale/readme.md
Confectrician: Hm to be honest. Some reviews (on of yours) above i had to change it from lower to upper case.
Maybe we should talk for those examples about a general handling in the future. :D

I like the suggestion and will change it anyways.
Its just something to think about.

_addons_transformations/scale/readme.md
Confectrician: In fact i would leave it this way.
With this setup we have an example on which we can explain what happens with values which are not recognized by any of the matching ranges.

Confectrician:
> The last thing: I don't feel the UV index example is a good fit. It does currently not show all cases and the origin of the 0 to 100 item is not clear. If you want to spend the extra time, I think another example could be better.

1. I want, but not now
2. Yes i feel the same about that example, but i didn't find a better one yet, since i am using scale not that long yet.

Maybe  @clinique has a live example, we could hang on for the future?

I had already thought about something temperature like.

- -10..0 -> "Double your underwear, if you wana go outside"
- 0..10 -> "Still really cold outside"
- 10..20 -> "Spring is on its way"
- 20..30 -> "Don't forget your sunscreen"

Long story short:
I am on the topic, but i have not thought about details yet.


### Fun-Fact about the esh repo:
I tried to create an eclipse account yesterday already.
Automatic account activation mail arrived after impressive 11 hours^^
Now i can go on preparing my contribution there.

_addons_transformations/scale/readme.md
ThomDietrich: To be honest, you can do it either ways - it just has to be consistent between this line and the item definition 😉

ThomDietrich:
Okay don't worry about the example for now. Go for the PR! 🎉 😀

ThomDietrich:
https://github.com/eclipse/smarthome/pull/4443 got merged. Congrats @Confectrician !

PR #488, opened 2017-09-02
Add general usage description to Transformations

ThomDietrich:

### DIFF
--- a/addons/transformations.md
+++ b/addons/transformations.md
@@ -1,51 +1,67 @@
 ---
 layout: documentation
-title: Transformations
+title: Transformation Services
 ---
 
 {% include base.html %}
 
-# Transformations
+# Transformation Services
 
 Transformations are used to translate data from a cluttered or technical raw value to a processed or human-readable representation.
-They are often useful, to **interpret received Item values**, like sensor readings or state variables, and translate them into a human-readable or better processible format.
+They are often useful, to **interpret received Item values**, like sensor readings or state variables, and to translate them into a human-readable or better processible format.
 
 **Examples:**
 
-* Translation of a technical Binding output, e.g., "CLOSED" can be translated to the spanish translation "cerrado"
-* Processing of a raw Item value, e.g., Parsing a number from a JSON string, like `{ "temperature": 23.2 }`
-* Conversion of sensor readings, e.g., temperature in degree Celsius can be converted to degree Fahrenheit
+- Translation of a technical Binding output, e.g., "CLOSED" can be translated to the Spanish translation "cerrado"
+- Processing of a raw Item value, e.g., Parsing a number from a JSON string, like `{ "temperature": 23.2 }`
+- Conversion of sensor readings, e.g., temperature in degree Celsius can be converted to degree Fahrenheit
 
-**Transformations are applicable in:**
+## Usage
 
-1. Item and Sitemap Label
+Transformations are applicable in Item and Sitemap element labels and inside DSL rules.
+Be aware, that some Transformation services rely on transformation files, while others work by directly providing the transformation logic.
+Transformation files need to be placed in the directory `$OPENHAB_CONF/transform`.
 
-   Transformations used in the [state/value part]({{base}}/configuration/items.html#state-transformations) of Labels are applied **on the fly**.
-   While the **transformed value** will (for example) be visible on a Sitemap, the **original value** is however stored in the Item.
-   
-   The following example shows a Map transformation (see below) used in the State part of an Items label to transform the technical state of a Contact Item (e.g. "CLOSED") into a human readable representation ("cerrado"):
-   ```java
-   Contact LR_Window "Livingroom Window [MAP(window_esp.map):%s]"  { someBinding:... }
-   ```
-   
-   The same method can be used in the [Label parameter of Sitemap Elements]({{base}}/configuration/sitemaps.html).
+1.  Item and Sitemap Labels
 
-2. Rules
+     Transformations used in the [state/value part]({{base}}/configuration/items.html#state-transformations) of labels are applied **on the fly**.
+     While the **transformed value** will (for example) be visible on a Sitemap, the **original value** is stored in the Item.
 
-   Transformations can also be used in Rules to **transform/translate/convert data**.
-   The general syntax is as follows:
-   ```java
-   transform("<transformation-identifier>", "<transf. expression or transf. file name>", <input-data or variable>)
-   ```
-   
-   The following shows two examples. For more details regarding the Map and JsonPath transformations, check the individual transformation articles below.
-   ```java
-   var condition = transform("MAP", "window_esp.map", "CLOSED")
-   var temperature = transform("JSONPATH", "$.temperature", jsonstring)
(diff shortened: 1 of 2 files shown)


ThomDietrich:
A first attempt to introduce transformations with everything that came to mind.
Comments? @rkoshak @Confectrician

LightIsLife:
Honestly, you are moving way too fast here and it may be difficult to follow along. There are so many ways to do things in OpenHAB, a little slower pace may help the non-experts. 

For example, any of the examples below are too abstract (at least for me)
 ```java
 +   transform("<transformation-name>","<logic or transformation file name>", <input-data or variable>)
 +   
 +   var weather = transform("MAP", "weather_en.map", rawvalue)
 +   var author = transform("JSONPATH", "$.store.book[0].author", jsonstring)
 +   ```

Most of the time I still struggle to understand when which transform needs or can be used when, why and for what purpose. (I get the basics: it transforms.....but how exactly does it work with json, regex is a perpetual mystery and what about the map transform? when do can I use what again? etc, etc)
Just my thoughts, take it or leave it. It all depends on the desired target audience. You and others are so knowledgeable, it is just awesome, but I am not always clear at what level of understanding the docs are aiming.

ThomDietrich:
Hey @LightIsLife,
thanks for your honest comment! I'm writing this kind of content especially for the new and unexperienced user. I think it is very important to make the documentation a usable source for new users and experienced users alike. I think adding clear details to a specific topic (like #486 by @Confectrician) is important to both but an introduction like this one will go a long way picking up the new user.

"Make everything creamy" is a quote by a consultant I like to remember :)

So back to your comment: I've put some more thought into the article and created a second commit. Could you please go through and let me know where exactly you feel further clarification is needed? Thanks!

LightIsLife:
Thanks much @ThomDietrich appreciate your comments and feedback. I think your edits helped quite a bit in stating in much more obvious what transforms do.

Here some more thoughts:
- it might help to use the same examples throughout; in the beginning translation into Spanish is used, the MAP example maps CLOSED though; 
- Fahrenheit and Celsius conversion is mentioned as a possibility, but no examples are given

Regarding the MAP transform, I think one or two additional statements connecting the dots might help. In particular, the fact that the mapping pairs are always in a separate file, which needs a unique filename and needs to be stored in the transform folder is missing. The example for mapping CLOSED mentions the file "window.map", but no info is giving on the format of this file or its location. I remember that my mind totally overcomplicated things in the beginning. The text already addresses a lot of this (on the fly, only for sitemap, items still stores the original value, etc) this part works great; if it is possible to add a sentence that MAP transforms are based on one-to-one mappings in a separate file (here windows.map) that needs to be placed in transforms folder. Example for the content of the file 
CLOSED:currently closed
OPEN:currently open
This would also cover the questions on whether quotation marks for strings longer than one word are needed (or whether space needs to be escaped) and whether there are any line separators such as commas needed if there is a list of mapping. 
Maybe part of all this belongs to the MAP readme, which as of right now is utterly barebone.

The JSON transform is a little more straightforward and more links to read up on are given in the readme file.

LightIsLife:
Just an additional thought, as this is the overarching section on transforms, would it be possible to give a list, table or similar that as an overview when to apply which transforms.At first glance the variety of transforms is either unclear or confusing. Not sure though, if what I have in mind works; maybe the content of this list (w/o the logos): http://docs.openhab.org/addons/transformations.html could be pulled in

rkoshak:
A table is absolutely possible and could be a good fit for presenting this content. See the Things,  Items, Installation Overview, and elsewhere for places where tables are used as an example for how to do it with the markdown syntax.

ThomDietrich:
Not sure if we are on the same page here. The file I've changed here is the one behind http://docs.openhab.org/addons/transformations.html
The text added in this PR **will** be above this table :)

rkoshak:
> Not sure if we are on the same page here. 

My interpretation of what Lipp was saying was he would like to format the transforms into a table rather than the existing bullet points. I could be wrong though.

ThomDietrich:
Hey guys, I've synced the examples. It's not finished but I have to leave now.

@rkoshak I was not able to get the following to work. Do you have any experience with the JS transformation? Otherwise I'll have to do a few more tests next week.

```java
var fahrenheit = transform("JS", "convertTempToF.js", temperature)
```

> 2017-09-06 17:06:58.672 [ERROR] [ore.transform.actions.Transformation] - Error executing the transformation 'JS': An error occurred while loading script.

It seems like he can't find the script file, it's under the $OPENHAB_CONF/scripts folder.

rkoshak:
It doesn't go in the scripts folder. It goes in the transform folder.

It is confusing I know. But the scripts folder is strictly intended to store OH Rules Scripts. Shell scripts for the exec binding, JavaScript transformations, and JSR223 rules scripts do not belong there. It really is an unfortunately named folder.

ThomDietrich:
For myself.
Compare with the MAP README:

To do this you have to create a .map file in your `transform` folder.
These files are typical key/value pair files.

```text
key1=value1
key2=value2
...
```

Let's make a small example to illustrate this function.
If you have a sensor which returns the number 0 for a closed window and 1 for an open window, you can transform these values into the words "opened" or "closed".
Create a map file named `window.map` for example and add the desired keys and values.

```text
0=closed
1=opened
NULL=unknown
-=unknown
```

Next we define two Items.
One showing the raw value as it is provided from our sensor and one with transformed value.

```java
Number WindowRaw          "Window is [%d]"                  { someBinding:somevalue }
Number WindowTransformed  "Window is [MAP(window.map):%s]"  { someBinding:somevalue }
```

The output will be:

```text
Window is 1
Window is opened
```

Transform files use UTF-8 encoding, so Unicode symbols will also work.

```text
ARIES=♈ Aries
TAURUS=♉ Taurus
WAXING_CRESCENT=🌑→🌓 Waxing Crescent
FIRST_QUARTER=🌓 First Quarter
```

LightIsLife:
> My interpretation of what Lipp was saying was he would like to format the transforms into a table rather than the existing bullet points. I could be wrong though.

Sorry, saw the discussion only now, what I meant is a simple table that would clarify when to use what transform, or the other way round, what could each transform offer. Such table exists here already: http://docs.openhab.org/addons/transformations.html, but not in the configurations chapter. Also, personally, I would hope that the table could be much more compressed, the one from the addons looks a little oversized for its content and for the life of me, I cannot make sense out of the symbols (install/auto).

The descriptions from the addons may also need some improvement. May 'favorite' one is the definition for MAP which is a circular definition adding only one word of new info: "Transforms the input by mapping it to another string. " the transform is called MAP and it is under the folder transformations, the only new information is "string".
 
Something like this may be more helpful (rough draft): A MAP transformation expects as input a string and returns another string. The matching pairs of string are defined in and retrieved from a text file. This file is stored in the transform folder.

ThomDietrich:
I've added final changes and thing that should be it. @LightIsLife I've read all your concerns. These descriptions need to be edited in the respective README files over at https://github.com/eclipse/smarthome/tree/master/extensions/transform
If you open a PR, please mention me ;)

PR #489, opened 2017-09-02
[Items] Remove invalid mapping info for dyn. icons

ThomDietrich:
This info is wrong. If anyone can make sense of it and knows how to improve this part, please let me know.
@rkoshak 

### DIFF
--- a/configuration/items.md
+++ b/configuration/items.md
@@ -82,7 +82,7 @@ The last example above defines an Item with the following parts:
 
 The remainder of this article describes the Item definition parts in more detail.
 
-### Types
+### Type
 
 The Item type defines which kind of state can be stored in that Item and which commands can be sent to it, e.g. String, Number or binary Switch.
 They are comparable with basic variable data types in programming languages.
@@ -256,9 +256,6 @@ Dynamic icon families must meet the following criteria:
 
 -   the state part of the icon must consist of all lower case letters
     (even if the state of the Item includes uppercase letters)
-
--   the icon selected is based on the mapped value used in the label
-    (i.e. what gets returned by `[MAP(file.map):%s]`, not the Item's raw state.
 
 The name of dynamic icons must meet the following format:
 


ThomDietrich:
I've just tested all combinations I could think of, there is no relation between mappings and dynamic icons.

@rkoshak can you make any sense of it? https://github.com/openhab/openhab-docs/commit/40c8a07e0ef4b15c20b714de086eb04486ce4420#diff-7e7b8936c68aca319804243a873efe83R277

ThomDietrich:
Additional Commit applied directly by accident: https://github.com/openhab/openhab-docs/commit/bbcd167f0b42e0ca9f15248e17ee82f293a9587c

rkoshak:
Perhaps this has changed again...

Sometime after 2.0 people started to have problems where they would use a MAP or some other transform in their label and the dynamic icons would stop working.  For example, if you mapped ON to "present" and OFF to "absent" in a MAP in the label you would have to have presence.png, presence-present.png, and presence-absent.png instead of the default ON/OFF versions or else the correct icon wouldn't be chosen. 

Dim Angelos (@dim on the forum, don't know his Github handle) just recently helped someone get dynamic icons working because of this problem. I've not done an exhaustive study of this behavior, perhaps it only applies to certain types of Items but not all, or the behavior has changed. All I can say for certain is that several people have reported problems with dynamic icons and the fix was to make the icons match the transformed version of the state in the label, rather than the Item's raw state.

ThomDietrich:
I'd be thrilled to see an example :)

ThomDietrich:
@rkoshak I realized that there is one constellation where this is true. 

```java
String FK_Senseo_OpState "Senseo Wifi Zustand [%s]" <senseo_coffeesign> (gSenseo) {mqtt="<[broker:senseo-w66a/machine/opState:state:MAP(senseo-wifi.map)]"}
```
A transformation used by the mqtt Binding will indeed result in translated item states. That is clearly a totally different case but I wonder if the "myth" originated out of smth like it.

rkoshak:
The myth actually came from a pretty rigorous testing of the behavior. I'm trying to find the original posting where we tested all sorts of different combinations to discover what was going on. I think the post was in the old 2.0 Beta 1 days but am not sure. My Google Foo is failing me on the forum. I'm exporting all my postings to see if I can find it in Excel.

I'll post the link when/if I find it.

But I can say from memory that I think it was a Switch Item that was being mapped to something different from ON and OFF and the icons wouldn't work without creating copies of the icons with names mapped to what the value was in the label, not the Item's State. I remember thinking and saying this feels really wrong at the time but can't remember if someone filed an issue on it (I know I didn't).

I can say that at least for Switches and Contact Items in my sitemap right now, I am mapping the states and it is working as expected (i.e. using the Item's state. not the transformed label state) so perhaps this problem was fixed somewhere along the way and I didn't notice it.

Issue #492, opened 2017-09-03
Sitemaps: Add MapView Sitemap element


ThomDietrich:
(Derived from https://github.com/openhab/openhab-docs/issues/53)

All details needed can be found in this diff: https://github.com/eclipse/smarthome/pull/277/files#diff-10aa8fe84e9119cfe8cac6ea022ff706

The probably most time consuming task here is probably to create a screenshot :)

ThomDietrich:
Addition: While on it, please create all mismatching images (e.g. the temperature Text element should read 21.5°C)

Confectrician:
@ThomDietrich 

I have read your comment above 3 times now and i simply don't get it.
Could you explain this for me again?

Does ist simply mean that all screenshots should be based on the same sitemap and the same current values?

ThomDietrich:
Hey @Confectrician,
yes I should have made myself more clear :) The sitemap used to create the screenshots is part of the sitemaps article. It is appended at the end of the article as a comment (It ain't stupid if it works 😄). See: https://github.com/openhab/openhab-docs/blame/gh-pages/configuration/sitemaps.md#L653

Regarding the temperature: As you'll see, the temperature is hardcoded with "21.3 °C". However, [later on](http://docs.openhab.org/configuration/sitemaps.html#element-type-setpoint) I used the temperature item to show an example for the setpoint sitemap element. The example uses a step resolution of 0.5 degree, which is meaningful but one might wonder how 21.3°C can be achieved with such an element. Long story short: If we change the screenshot to show 21.5°C, the problem will not come to mind.

Confectrician:
You can assign me.
PR is nearly incoming^^

I would suggest to edit all temperature related screenshots too, to keep a common thread during the complete article.


Additionally:
Is there a reason for sorting the element types in an alphabetical order within the table and not sorting it in the later content?!

Should i just append the MapView as last element type?

PR #493, opened 2017-09-03
[Sitemaps] Revise article, add element examples

ThomDietrich:
Includes all fixes for what was mentioned in https://github.com/openhab/openhab-docs/issues/53

I've also added the actual code used to create the sitemap elements in the screenshots for more clarity.

A typical section now looks like:

grafik

### DIFF
--- a/configuration/sitemaps.md
+++ b/configuration/sitemaps.md
@@ -20,12 +20,12 @@ This page is structured as follows:
 {:toc}
 
 The definition of Sitemaps happens declaratively in a file with a clear syntax, described below.
-A Sitemap definition file is stored in the folder `<openhab_config>/sitemaps` and has to have the `.sitemap` filename extension.
-For easy editing, the [Eclipse SmartHome Designer]({{base}}/configuration/editors.html#esh-designer) brings full IDE support for these files.
+A Sitemap definition file is stored in the folder `$OPENHAB_CONF/sitemaps` and has to have the `.sitemap` filename extension.
+For easy editing, profit from the IDE support of the [openHAB supporting editors]({{base}}/configuration/editors.html).
 
 The openHAB runtime comes with a demo configuration package containing a [`demo.sitemap`](https://github.com/openhab/openhab-distro/blob/master/features/distro-resources/src/main/resources/sitemaps/demo.sitemap),
 which should let you easily understand possible elements and structures.
-It is recommended to use the `demo.sitemap` or another example Sitemap as a starting point towards building a customized Sitemap that fits your personal home setup.
+It is recommended to use this demo file or another example Sitemap as a starting point towards building a customized Sitemap that fits your personal home setup.
 
 The following example illustrates what a typical Sitemap definition might look like:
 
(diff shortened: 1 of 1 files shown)


bgilmer77:
@ThomDietrich, Thanks for asking me to take a look.  I believe the text below conveys the same information, and also answers some other questions implementers may have:

The 'Selection' element type renders a push-button menu where the user must select one, and only one choice.  Making a selection de-selects another choice that was previously selected.  [Question for @ThomDietrich - is this behavior correct?]  

When the user makes a menu selection, openHAB sets the value of the 'Selection item' to the value of the selected entry in the mapping.  In the example below, if the user chooses 'Cartoon Network', then LR_TV_Channel is set to the value '2'.  This is further described in the mappings section below.

When openHAB initializes, <itemname> is set to the value associated with the first mapping entry.  In the example below, the initial value would be set to zero.   [Question for @ThomDietrich, is this correct?]  Note that zero does not have to be the value associated with the first mapping entry; any legal value may be used.

Confectrician:
I am still reading but wanted to add one comment already:
Maybe it is necessary to change some of the screenshots now, since they probably don't match the example.
This could be confusing for beginners.

ThomDietrich:
@bgilmer77 you are referring to the screenshot but the Pull Request changes a lot more than that. See the "Files changed" tab on the top of this page: https://github.com/openhab/openhab-docs/pull/493/files

Thanks for your corrections, I'll add them to the PR.

@Confectrician Which do not match? At least one screenshot needs to be created for #492, so let's move that over there.
Let me know if you find any other problem!

bgilmer77:
@ThomDietrich I understand your point.  But it is not clear to me where in openHAB the behavior of the Selection element type is established.  Two modes that are common are 'radio button' where you may choose one and only one selection from a list of choices (you do not have the option not to make a choice), and 'check box' where you may select zero or more choices from a list.  Where is this behavior determined?  
Also, when the system is initialized, is there a default state, or should we caution implementers that the system may come up in a random state so they should ensure that they use persistence, and that they set an initial power-up value for these parameters?

ThomDietrich:
@bgilmer77 forgive me but I'm not sure what to extend.

Isn't it obvious that the selection offers the option to select *something*? The terms "radio button" and "check box" are known from other UI worlds, why are they important here? The selection is similar to a radio button but who would benefit from that analogy?

A multi-selection is not explicitly mentioned, which I feel would be mentioned if that was indeed possible - which it isn't.

Regarding the initialization of Items: That topic is unrelated to Sitemap elements. Of course that should be discussed somewhere and I think a chapter in the Items article would be the right place. How do you feel about that?

> should we caution implementers that the system may come up in a random state so they should ensure that they use persistence, and that they set an initial power-up value for these parameters?

In fact that's already covered partly: http://docs.openhab.org/configuration/rules-dsl.html#system-based-triggers

ThomDietrich:
Please open a new issue if anything is still wort further discussion! ;) Thanks

PR #494, opened 2017-09-03
[Items] Add naming recommendations, revise article

ThomDietrich:
I always thought we should give some recommendations while not forcing anyone. Please don't see these paragraphs as a complete Style Guide but rather as some suggestions for users to follow or to orientate on while defining their own scheme.

@kubawolanin wdyt?

### DIFF
--- a/configuration/items.md
+++ b/configuration/items.md
@@ -428,10 +428,18 @@ The third example groups combines all room temperatures and the group state repr
 {: #tags}
 ### Tags
 
-Tagging is a new feature under development.
+Tags added to an Item definition characterize a specific nature of the Item beyond it's basic Item type.
+Tags can then be used by add-ons to interact with Items in context-sensitive ways.
 
-Tags are used by some I/O add-ons, see [Hue Emulation]({{base}}/addons/io/hueemulation/readme.html) for more details.
-Tags are only of interest if the documentation for a specific add-on or integration explicitly discusses their usage.
+To give an example:
+A Light in a typical home setup can be represented either by a Switch, Dimmer or Color Item.
+To be able to specifically interact with the light device, e.g. via a natural voice command, the nature of the Item as a Light needs to be known.
+This can be achieved by adding the "Lighting" tag.
+
+Tagging is still a new feature and only some specific I/O add-ons are using it.
+Tags are only of interest, if the documentation for a specific add-on or integration explicitly discusses their usage.
+
+See the [Hue Emulation]({{base}}/addons/io/hueemulation/readme.html) or [HomeKit Add-on](http://docs.openhab.org/addons/io/homekit/readme.html) documentation for more details.
 
 {: #binding}
 ### Binding Configuration
--- a/configuration/items.md
+++ b/configuration/items.md
@@ -33,8 +33,8 @@ Read the [docs page for the respective Binding]({{base}}/addons/bindings.html) t
 
 There are two methods for defining Items:
 
-1.  The first is through PaperUI.
-    Generally all 2.x version Bindings can be configured through PaperUI.
+1.  The first is through [Paper UI]({{base}}/addons/uis/paper/readme.html).
+    Generally all 2.x version Bindings can be configured through Paper UI.
     Other 1.x and legacy Bindings do not offer this path.
 
 2.  The second method is through text `.items` files in the `items` folder.
@@ -43,7 +43,7 @@ There are two methods for defining Items:
     Generally 1.x version Bindings can only be bound to Items through `.items` files.
     2.x Bindings can also be used with this method
 
-**Assumptions for PaperUI:**
+**Assumptions for Paper UI:**
 The following content will discuss details of item definition on the example of `.items` files.
 
 **Editor Recommendation:**
@@ -57,31 +57,33 @@ The parts of an Item must be defined in this order.
 Besides the `itemtype` and `itemname` all parts are optional.
 
 ```java
-itemtype itemname "labeltext" <iconname> (group1, group2, ...) ["tag1", "tag2", ...] {bindingconfig}
+itemtype itemname "labeltext [stateformat]" <iconname> (group1, group2, ...) ["tag1", "tag2", ...] {bindingconfig}
 ```
 
 **Examples:**
 
 ```java
-Number BathRoom_WaschingMachine_Power "Power [%.0f W]" <energy> (gPower) {channel="homematic:HG-HM-ES-PMSw1-Pl:ccu:LEQ1275872:2#POWER"}
+Number Bathroom_WaschingMachine_Power "Power [%.0f W]" <energy> (gPower) {channel="homematic:HG-HM-ES-PMSw1-Pl:ccu:LEQ1275872:2#POWER"}
 Switch Kitchen_Light "Kitchen Light OnOff" {mqtt="<[...], >[...]" }
-String BedRoom_Sonos_CurrentTitle "Title [%s]" (gBedRoom) {channel="sonos:PLAY3:RINCON_C5E93734496A0A400:currenttitle"}
+String Bedroom_Sonos_CurrentTitle "Title [%s]" (gBedRoom) {channel="sonos:PLAY3:RINCON_C5E93734496A0A400:currenttitle"}
 
-Number LivingRoom_Temperature "Temperature [%.1f °C]" <temperature> (gTemperature, gLivingRoom) ["TargetTemperature"] {knx="1/0/15+0/0/15"}
+Number Livingroom_Temperature "Temperature [%.1f °C]" <temperature> (gTemperature, gLivingroom) ["TargetTemperature"] {knx="1/0/15+0/0/15"}
 ```
 
 The last example above defines an Item with the following parts:
 
-- Item type `Number`
-- Item name `LivingRoom_Temperature`
-- Item state formatted in a way which will produce for example "21.5 °C" as its output
-- Item displaying icon with the name `temperature`
-- Item belongs to groups `gTemperature` and `gLivingRoom`
-- Item is tagged as a thermostat ("TargetTemperature")
-- Item is bound to the openHAB Binding `knx` with binding specific settings
+- Item [type](#type) `Number`
+- Item [name](#name) `Livingroom_Temperature`
+- Item [label](#label) "Temperature"
+- Item [state formatted](#state-presentation) in a way which will produce for example "21.5 °C" as its output
+- Item [icon](#icons) with the name `temperature`
+- Item belongs to [groups](#groups) `gTemperature` and `gLivingroom` (definition not shown in the example)
+- Item is [tagged](#tags) as a thermostat ("TargetTemperature")
+- Item is [bound to](#binding) the openHAB Binding `knx` with binding specific settings
 
 The remainder of this article describes the Item definition parts in more detail.
 
+{: #type}
 ### Type
 
 The Item type defines which kind of state can be stored in that Item and which commands can be sent to it, e.g. String, Number or binary Switch.
@@ -96,15 +98,13 @@ All available openHAB2 Item types and their relevant commands can be viewed here
 While a Dimmer Item can accept either On/Off, Increase/Decrease, or Percent updates or command, a Dimmer Item stores its state as a Percent value.
 See the following example:
 
-Item:
-
 ```java
+//demo.items
 Dimmer Office_Light "Dimmer [%d %%]" {milight="bridge01;3;brightness"}
 ```
 
-Sitemap:
-
-```perl
+```javascript
+//demo.sitemap
 Switch item=Office_Light
 Slider item=Office_Light
 ```
@@ -113,6 +113,7 @@ When the Switch widget is used, it sends ON or OFF commands to the Item which ar
 When the Slider widget is used, it sends Percent commands (values between 0 and 100) to the Item, which are used as the Item's state.
 In the example above, if you move the Slider widget to 60%, move the Switch to OFF, and finally move the switch to ON, the Item's state will be 100%.
 
+{: #name}
 ### Name
 
 The Item name is the unique identifier of the Item.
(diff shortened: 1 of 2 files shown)


configuration/items.md
```diff
@@ -118,7 +118,20 @@ In the example above, if you move the Slider widget to 60%, move the Switch to O
...
+It is advised to follow a predefined naming scheme.
+The officially recommended scheme is used throughout this documentation and builds on a hierarchy idea.
+Floor or room names can be abbreviated to reduce overall name length.
+
+Every word of the Item name starts with a uppercase letter.
+Words are separated by an underscore character, except for logically belonging words.
+
+Examples:
+
+- "`BathRoom_WaschingMachine_Power`" for the used power of a washing machine in a single floor apartment bathroom
```
kubawolanin: I'd go with `Bathroom_WashingMachine_Power`

configuration/items.md
```diff
@@ -118,7 +118,20 @@ In the example above, if you move the Slider widget to 60%, move the Switch to O
...
 The name should only consist of letters, numbers and the underscore character.
 Spaces and special characters cannot be used.
-A good naming schema can be advised.
+
+**Style Guide:**
+It is advised to follow a predefined naming scheme.
+The officially recommended scheme is used throughout this documentation and builds on a hierarchy idea.
+Floor or room names can be abbreviated to reduce overall name length.
+
+Every word of the Item name starts with a uppercase letter.
```
kubawolanin: Except for the device Group items apparently :-)
```
Every word of the Item name starts with a uppercase letter, unless it's a `Group` item.
```

configuration/items.md
```diff
@@ -118,7 +118,20 @@ In the example above, if you move the Slider widget to 60%, move the Switch to O
...
+Words are separated by an underscore character, except for logically belonging words.
+
+Examples:
+
+- "`BathRoom_WaschingMachine_Power`" for the used power of a washing machine in a single floor apartment bathroom
+- "`GF_LR_Heating_Setpoint`" for the Groundfloor Livingroom set temperature
```
kubawolanin: How about
"`GF_Living_Heating_Setpoint`" for the Ground floor Living room set temperature
?

I would like to avoid abbreviating specific rooms.

configuration/items.md
```diff
@@ -311,9 +324,13 @@ Groups are supported in Sitemaps, Automation Rules and other areas of openHAB.
 A simple example Group definition is:
 
 ```java
-Group Temperatures
+Group gGroundFloorTemperatures
```
kubawolanin: My suggestion would be to stick with `gTemperatures`.

ThomDietrich:
Hey Kuba, I should have mentioned that my addition should be seen as a first draft and I wanted to start an active discussion about the final content.

Here's my first question for you: You write

> 4. Names of Groups of devices are g[device]

However for the floor groups you are going a different direction:

```java
Group:Number:AVG                 gTemperature   "Temperature"   <temperature>   (Home)
//but
Group     GF                      "GroundFloor"   <groundfloor>      (Home)
```

In fact I had my setup in the exact same way and switched recently. See the [this commit](https://github.com/ThomDietrich/openhab-config/commit/1ffb3808ce2a3b9a498da0a0466fbc589ee34341#diff-72d42aa6df16b9877247c614de6bf760).

There are two reasons to do so:

* If we agree that groups should start with a prepended "g" then we should do that all the way
* More important for me: a name like "GF" will be identified as a CONSTANT by most highlightings:

```java
Group   BD  "Bad"               <bath>      (Whg)
// vs
Group   gBD  "Bad"              <bath>      (gAll)

Number WZ_Heizung_ActTemp "WZ_Heizung Ist-Temperatur [%.1f °C]" <temperature> (gWZHeat,WZ,gHeatAct) { channel="homematic:HG-HM-CC-RT-DN:ccu:LEQ0785556:4#ACTUAL_TEMPERATURE" }
// vs
Number WZ_Heizung_ActTemp "WZ_Heizung Ist-Temperatur [%.1f °C]" <temperature> (gWZHeat,gWZ,gHeatAct) { channel="homematic:HG-HM-CC-RT-DN:ccu:LEQ0785556:4#ACTUAL_TEMPERATURE" }
```

configuration/items.md
ThomDietrich: Reminder to self: "an uppercase"

ThomDietrich: I don't understand. Besides the g everything still starts with an uppercase letter!?

configuration/items.md
ThomDietrich: Agreed!

configuration/items.md
ThomDietrich: Not sure, the example is there to **showcase** abbreviations and the user doesn't have to do it the same way.

configuration/items.md
ThomDietrich: Again, this is an example. An example to show how to name a multi-word Group.

However I should probably give a list of Group name examples, just like above.

configuration/items.md
BClark09: I agree with @ThomDietrich, using "unless" can get confusing, although if you want to highlight the difference adding "Group names are prepended with a lowercase 'g'" won't hurt right?

kubawolanin:
Hi @ThomDietrich,

You're right - showcasing the examples is a good thing! Didn't mean to point right/wrong way :innocent: 

> You write
> > Names of Groups of devices are g[device]
> 
> However for the floor groups you are going a different direction:

Yep, because floors are not devices :-) I agree it sounds odd though :smile: 
What I understand as "device" is a single physical sensor/actuator that:
* can (and should) be assigned to an item
* belongs to a specific room
* belongs to a group of those devices

So given that we have a Window sensor in the bedroom on the first floor, I'd assign it to `FF_Bedroom_Window` or something like that :-)
Note that each *room* is also a Group.
Anyways, all this is just an idea I've built my generator around.

> Basically my scheme already covers this... in the sense, that it suggests a hierarchy but not defines everything explicitly. Do you feel that's needed?

Nope. Just throwing my 3 cents. Most likely I'll cover that scheme in HomeBuilder README, as it's more specific for the tool, rather than THE style guide to follow along :grin: 

Thank you for tackling this subject! I'm sure it'll be helpful. :beer: 
Cheers

ThomDietrich:
Heyho Kuba,
we are pretty much on the same page 👌 

A few more words regarding groups: In my understanding there are two types of groups: "horizontal and vertical":

1. physical hierarchy: floors, rooms, appliances/devices, ...
2. logical hierarchy: similar Items, maintenance groups, persistence groups, ...

So just as your homebuilder generates my system consists of a common "gAll" group, several room groups (e.g. "gLR"), groups for specific parts (e.g. gLRHeating) and this could go on. This is the physical perspective.
The second level is a vertical cut through the physical hierarchy and combines logically similar Items. Examples are "gBattery, gLight, gError, gUnreachable, gRSSI". Of course these can also be nested. In my setup there is for example a group "gMaint" combining all of the above maintenance groups.

Back to the issue at hand: I'm unsure if it is intuitive to name *some* groups with a prepended "g". There's also the CONSTANT argument that will create confusion among non-developers who will try to make sense of the coloring.

I think we should agree on a common style between the documentation and other important core parts of openHAB. The Home Builder would of course be one of these. My opinion and scheme are not fixed, let's find a compromise 🤝

ThomDietrich:
Out of coincidence this thread just appeared in the community: https://community.openhab.org/t/yagsa-yet-another-group-structure-approach/33939

I think this only shows the importance of such details to be documented and available for users...

kaikreuzer:
>  Also wanted to invite @kaikreuzer to this.

Thanks - although I am not sure, if I can add much value... I like the general idea&approach and my main wish is that in the end the HomeBuilder is in sync with the guidelines ;-)

> I'm unsure if it is intuitive to name some groups with a prepended "g". 

Right, so besides deciding that all groups need a "g", the other option would be to completely drop such a prefix. For the functional groups, I would actually prefer this - the plural could be used instead: Lights, Windows, etc. Would imho be more intuative than "gLight". For room names, it would also be quite natural to already know that those are group items without requiring such a marker.

ThomDietrich:
Good of you to mention that.
I was also thinking about the combination of no g + plural. This is for sure more intuitive and not a weird mixture with the Hungarian notation idea which (HN) I pretty much disprize 😄 

Issues:
* Not everything can easily be given in plural. While "gLight" could be "Lights", my group "gMaintenance" doesn't have such an equivalent. On the other side "Maintenance" or a group for a room called "Livingroom" would already transport the idea of a nested element...
* The small issue of abbreviations I've mentioned before will be back without the "g".

BClark09:
> I was also thinking about the combination of no g + plural

This is what I use at the moment, the only other difference between that and @kubawolanin's first post is that I don't use floor notations now. I don't think I'll be fortunate enough to own a house with two of the same room name :derelict_house:. so didn't see the point in adding it to the name of each item.

e.g.

```
// Individual Items
Kitchen_MainLight_Brightness
Kitchen_MainLight_Switch
Livingroom_MainLight_Colour
Livingroom_MainLight_Switch
Livingroom_CornerLight_Colour
Livingroom_CornerLight_Switch

// Livingroom Groups
Livingroom_Lights_Colour
Livingroom_Lights_Switch

//Floor Groups
GroundFloor_Lights_Switch

// House Groups
Lights_Switch
```

rkoshak:
Just noticed this little thread. Everything looks great and I love the discussion.

My only real comment is while I agree the the "g" prefix is not terribly intuitive, I have a problem with using plural to denote groups as it makes one use case very difficult, my Associated Items design pattern. And in fact, it is only a subset of that one use case, mainly the ability to easily construct a Group's name to pull it out of its parent Group by name.

I honestly don't think this is a use case that is worth accommodating by any means. But I bring it up to make sure that we also think about how Item and Group names are used in Rules in addition to Sitemap and by the human user.

tbail:
I do agree with Rich using the plural to denote groups is not "Eye catching" i am more with the g

kaikreuzer:
Well, if plural isn't eye-catching, there'd always be the chance to add "Group" to its name like e.g. "Maintenance_Group" or "Group_Maintenance".

kubawolanin:
It may not be eye-catching but it's definitely more human readable :-)
My use case with plural groups - I use a voice recognition with openHAB and Wit.ai. I can say "Wyłącz Światła" ("Turn the Lights off") and it's immediately recognized by openHAB as an Item to switch. No hassle :-)

tbail:
From my point of view i like to identify groups very quickly. So the use of a lower g is simple and straight forward. Based on this a group name ist maybe build like this: <Identifier><Qualifier>. Where identifier is the g or Group or anything else. The qualifier is the human readable part with or with out plurals. If the identifier is more than just a g i would prefer to separate identifier und qualifier with a "_".

Having the voice recognition in mind (In my case Amazon Echo) it is maybe a good idea to give some hints how to design labels so that they are recognized properly. 

At this point i think that we are very deep in details. Is it maybe a good idea to go up a level and define some basis design principles like "Use a consistent naming scheme", "Experience has shown that it is agood ides to combine the following Information in item names: Location; Prupose, Channel, DayLightsavingTime or whatever" after we have these basics we should discuss a concrete scheme ...

ugh, so many things in my head ....

rkoshak:
Overall I think this thread is confirming something that I've thought for a long time. We have incompatible use cases which will make coming up with a consistent strategy very challenging.

It is far easier to use plural names for Groups for the Voice Command use case but it is all but impossible to programmatically construct a Group name in your rules if using the plural, at least in English (the mongrel language that it is), because of the inconsistent rules for pluralization. 

Personally, even though I construct the names of things in Rules all over the place, given the popularity of Alexa and the hopefully soon to be integration with Google Assistant, not to mention all the other voice command systems people have set up, I would error on the side of a naming scheme that supports a more natural voice commanding.

So maybe have footnotes or an appendix to explain some of the limitations of the proposed naming scheme and alternatives. 

So how about something like:

"Items and Groups should be named with a noun or noun clause. For Groups, use the plural of the noun. Using this approach is more human readable and supports voice commands. If the noun has an awkward pluralization (e.g. maintenances) that would not be used in voice commanding, use the singular. If one needs to be able to programmatically construct the name of the Group or Item, use a standard scheme (e.g. gGroupName, GroupName_Group, etc.)."

That covers all the bases. We push the human centered approach first (which I kind of like). We cover awkward nouns like maintenance (though despite spell checkers, it does have a plural). And we provide an out for those who need the other use case (i.e. rules name construction).

configuration/items.md
skatun: I always do  Group_LivingRoom etc instead of GF_Living , when i scroll through logs/paperui its then easy to see whats Groups and whats not.. And simimliar Temperature_MasterBedRoom, Light_Kitchen_Island,  Light_Spot1

configuration/items.md
ThomDietrich: ✔️

ThomDietrich:
All, thanks for your comments!

I've tried to respect all ideas and concerns. The naming scheme recommendations are now longer, while the style guide recommendation gives all freedom to the user to still do everything that we have discussed in this thread. The result can be found here: https://github.com/openhab/openhab-docs/blob/1a24b9eb50308ec17ef01107cf9c6496353dc22b/configuration/items.md#name

The question of groups names is still undecided, I struggle to selfishly select one (not that I have a true favorite). Maybe my next comment will bring some clarity... 

While working on the article I did quite a few other changes. It would be great if one-two could do a **full review** again!

ThomDietrich:
## Public Poll

[Here](https://github.com/openhab/openhab-docs/blob/1a24b9eb50308ec17ef01107cf9c6496353dc22b/configuration/items.md#name) you'll find the current group naming recommendations. Please vote on this comment with:

* 👍 - If you prefer the plural / "Group" approach
* ❤️ - If you prefer the "g" approach
* 🎉 - If you believe giving both options is acceptable
* 👎 - If you are not satisfied (please comment!)

Thanks!

configuration/items.md
```diff
@@ -304,34 +290,33 @@ and the icons for `Number WindowTransformed` would be:
...
+-   Vertical/Physical/Location perspective
+    - e.g. Floors of a house → Rooms on that floor → An appliance in that room...
+
+-   Horizontal/Logical/Context/Function perspective
+    - e.g. All lights, All room temperatures, Combined power consumption, ....
+    - Maintenance Group → All battery states → All battery states in percentage / volt
+
+-   and so on ...
+
+These relation can be exploited in sitemaps or in automation rules to browse Items or to apply computations and updates on Items.
```
kubawolanin: `These relations`

configuration/items.md
```diff
@@ -340,62 +325,54 @@ Group gLivingRoom  (gRoom)
...
 ```java
-Group:Switch:OR(ON,OFF)  mySwitches
+Group[:itemtype[:function]]  groupname  ["labeltext"]  [<iconname>]  [(group1, group2, ...)]
 ```
 
-The state of the `mySwitches` group will be ON if any of the members states are ON.
-But this means that once one Item in the group has its state changed to ON, the group's state gets set.
-Each subsequent Item that changes state to ON will not trigger "mySwitches changed" because the group isn't changing.
+- If the aggregation function is omitted, the function `EQUAL` will be used.
+- If the aggregation function and itemtype are omitted, no group state will be aggregated from member Items.
```
kubawolanin: I'd wrap function and itemtype with backticks `

configuration/items.md
```diff
@@ -115,97 +115,83 @@ In the example above, if you move the Slider widget to 60%, move the Switch to O
...
 
-```text
-Window is 1
-Window is opened
-```
+Transformations can be used in the state part of an Item, to translate the raw state of an Item into another language or convert technical values into human readable information.
 
-Transform files use UTF-8 encoding, so Unicode symbols will also work.
+**Example:**
+Translation of a technical Binding output, e.g., "CLOSED" can be translated to the Spanish translation "cerrado"
```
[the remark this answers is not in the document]
ThomDietrich: Yes why not. However just to be clear, this should eventually point to https://github.com/openhab/openhab-docs/pull/488

configuration/items.md
```diff
@@ -115,97 +115,83 @@ In the example above, if you move the Slider widget to 60%, move the Switch to O
...
-One showing the raw value as it is provided from our sensor and one with transformed value.
+Only the square brackets, the leading '%' and the trailing 'conversion' are mandatory.
+A few examples are given below:
 
 ```java
-Number WindowRaw          "Window is [%d]"                  { someBinding:somevalue }
-Number WindowTransformed  "Window is [MAP(window.map):%s]"  { someBinding:somevalue }
+Number    MyTemperature  "Temperature [%.1f °C]"     {someBinding:somevalue} // e.g. "23.5 °C"
+String    MyString       "Value [%s]"                {someBinding:somevalue} // e.g. "Lorem ipsum"
+DateTime  MyLastUpdate   "Last Update [%1$ta %1$tR]" {someBinding:somevalue} // e.g. "Sun 15:26"
```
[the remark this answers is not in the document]
ThomDietrich: Inside the "State" chapter or the "Groups -> Group State" chapter?

Also which of those do you think is the better example?
```
Group:Switch:OR(ON, OFF)    Lights         "All Lights [(%d)]"  // e.g. "(2)"
Group:Number:SUM            Lights         "All Lights [(%d)]"  // e.g. "(2)"
Group:Number                Lights         "All Lights [(%d)]"  // e.g. "(2)"
```

(Yes all of them are equal for a combination of simple Switch items.)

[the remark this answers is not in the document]
kubawolanin: Let's maybe go with `Group:Number:SUM` here  and rename it to `gLight` then :-)

ThomDietrich:
@kubawolanin I'm finished for today. All your comments are included plus a bunch of other improvements. The Items article is finally reaching a presentable maturity...
Could I bother you once again to proof read everything? Thanks mate!

kubawolanin:
LGTM! :-)

configuration/items.md
```diff
@@ -13,7 +13,7 @@ While a device or service might be quite specific, Items are unified substitutio
 Items can be Strings, Numbers, Switches or one of a few other basic Item types, a programmer can rightly compare Item Types with base variable types of a programming language.
```
ThomDietrich: @bgilmer77 this would be the best way to comment on a part of this PR. You simply click on "Files changed", then click the + symbol left to a line changed and add your comment. see:  https://user-images.githubusercontent.com/2870104/30526250-c4ec64b6-9c17-11e7-8bf7-ea0b0af234e6.png

ThomDietrich:
@kubawolanin @BClark09 the last thing I'm not happy about is the Nested Groups hierarchy/perspective itemization. Any suggestions are welcomed!

BClark09:
I like it and think the general gist of it is fine, but it might be easier to explain in with a non-technical paragraph which describes the situation for each perspective. I'd also consider changing the example of using `house` to `GroundFloor`, so there is an immediate step in the hierarchy:

>Say you have an item which stores the temperature of the living room called `Livingroom_Temperature`. From a location perspective, you may have a group called `Livingroom` which is in the group `GroundFloor`. This implies that `Livingroom_Temperature` is in the group `GroundFloor`.

> From a functional perspective, it can **also** belong to a group called `Temperatures`, which is in the group `Sensors`. This implies that `Livingroom_Temperature` is in the group `Sensors`.

> Using nested group hierarchies such as these allows a rule to iterate through all the sensors on the ground floor, therefore the rule does not need to be updated if a new item were to be added to any of these groups.

ThomDietrich:
Hey friends,
thanks to all for commenting and contributing to this PR. It started as a simple addition and ended up to be a complete overhaul of the original Items article. I'm very happy to finally have a presentable mature article on Items in documentation.

Thanks to everyone taking part in the Poll. I think we can agree, that the result is clear and both options seem to have their share of supporters. With that result I went the middle path and present both naming scheme recommendations to the article.

PR #502, opened 2017-09-16
[WIP] Fix duplicate binding names in user menu

### LINKED ISSUE #401: Binding menu has many duplicates
The left menu comes with many duplicate entries for bindings (whenever there's a 1.x and 2.x version):

screen shot 2017-06-16 at 23 45 53

This isn't really helpful and rather confusing (looks like a bug to the user).
I wonder if we should keep this long list at all - I would actually say no.
@ThomDietrich wdyt?

ThomDietrich:
Resolves #401 

TODO: Add similar changes to other addons

### DIFF
--- a/_includes/user-menu.html
+++ b/_includes/user-menu.html
@@ -68,7 +68,7 @@
           <li><a href="{{docu}}/addons/bindings.html">Overview</a></li>
           <hr />
           {% for binding in site.bindings %}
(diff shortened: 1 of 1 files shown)


kaikreuzer:
Ooops, just noticed your [WIP] now, sorry for merging early - but it already looks sooo much better now 👍

PR #505, opened 2017-09-17
install doc: openHABian uses openHAB version 2.1.0 (minor change)

hreichert:
The installation documentation for openHABian says:
> * openHAB 2 in the latest recommended version (2.0.0 stable)

But openHABian uses OpenHAB version 2.1.0; so changed to this:
> * openHAB 2 in the latest recommended version (2.1.0 stable)

### DIFF
--- a/installation/openhabian.md
+++ b/installation/openhabian.md
@@ -34,7 +34,7 @@ To that end, the project provides two things:
 The following features are provided by the openHABian images out of the box:
 
 * Hassle-free setup without a display or keyboard, connected via [Ethernet or Wi-Fi](#wifi-setup)
-* openHAB 2 in the latest recommended version (2.0.0 stable)
+* openHAB 2 in the latest recommended version (2.1.0 stable)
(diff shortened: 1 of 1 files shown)


ThomDietrich:
Thanks for the hint @hreichert I've fixed the issue in the openHABian repository and the newest version of the document will soon be visible in docs.

bgilmer77:
Consider eliminating  reference to a specific version/subversion of OH.  The statement creates an ongoing document maintenance issue.  Unless OHabian will not track the latest version, then isn't it sufficient to simply say, "openHAB 2 in the latest recommended version"?

Confectrician:
@bgilmer77 is totally right.
We have enough issues with different versions of bindings, snapshots, etc.

The concrete Version-Number doesn't matter here and we could save time and unnecessary work here.

ThomDietrich:
That is definitely correct. The line was originally introduced when openHAB 2.0 was in its non-stable state. Nowadays openHABian simply follows the latest stable. I'll remove the exact version number.

PR #518, opened 2017-09-27
[Sitemaps] Charts: Documentation for legend parameter

hreichert:
Regarding this ESH PR: https://github.com/eclipse/smarthome/pull/4291

Add documentation for the new `legend` parameter to the `Charts` section of sitemap.md

Signed-off-by: Holger Reichert <mail@h0lger.de>

### DIFF
--- a/configuration/sitemaps.md
+++ b/configuration/sitemaps.md
@@ -397,7 +397,7 @@ Video url="http://demo.openhab.org/Hue.m4v"
 
 ```perl
 Chart [item=<itemname>] [icon="<iconname>"] [label="<labelname>"] [refresh=xxxx]
-[period=xxxx] [service="<service>"] [begin=yyyyMMddHHmm] [end=yyyyMMddHHmm]
+[period=xxxx] [service="<service>"] [begin=yyyyMMddHHmm] [end=yyyyMMddHHmm] [legend=true/false]
 ```
 
 Adds a time-series chart object for displaying logged data.
@@ -412,6 +412,10 @@ Therefore, for an installation with only a single persistence service, this is n
 
 -   `begin` / `end` represent the beginning and end of the time axis of the Chart.
 Valid values are in the format: "yyyyMMddHHmm" (yyyy = year, MM = month, dd = day, HH = hour (0-23), mm = minutes).
+
+-   `legend` is used to show or to hide the chart legend.
+    Valid values are `true` (always show the legend) and `false` (never show the legend).
+    If this parameter is not set, the legend is hidden if there is only one chart series.
 
 <!-- TODO: This paragraph needs an update -->
 


ThomDietrich:
Hello @hreichert,
the parameter is only available in openHAB 2.2 builds. Could you please add a short note to the end of the description?

kaikreuzer:
> Could you please add a short note to the end of the description?

No, he shouldn't!
I thought we all agreed that the master branch of the docs should always contain the latest version and we won't start making differences in the content.
We should rather soon manage to publish a version of the 2.1.0 tag as the docs for the latest official release.

hreichert:
OK, so this is ready for merge?

kaikreuzer:
I would say so - @ThomDietrich do you agree on the versioning topic?

ThomDietrich:
Hey @kaikreuzer,
Yes we've agreed on that and this is still something of high importance. In an ideal world we would have it now and wouldn't have this problem. However I told you just recently that my time is currently limited and I've tried to attend to more pressing issues lately.
Looking at the general problem at hand - a PR that is only valid for the unstable openHAB version - I see a few options for here and now:

1. Merge unchanged and confuse users who have the latest stable installed
2. Wait with the PR for the versioning feature. Without a clear deadline not reasonable.
3. Merge with a small version hint (known from many documentations) that is easily removed at a later point, e.g. via search&replace

When writing the above comment I thought (3) is a good middle path without any drawbacks. But I don't want to overcomplicated it (all binding docs have the same issue, hence this is just a drop in the ocean) and rather finish the versioning feature in the next 2-3 weeks. Not sure if I can implement it before the Smart Home Day, otherwise we can discuss details there.

Small Side Note: I don't want to stress this too much but I feel like I should at least mention it: It happens that I do not react to messages including exclamation marks... No hard feelings, best wishes, Thomas ;)

ThomDietrich:
@hreichert, I didn't say it before but thanks for your contribution! Looking forward to your next improvement idea ;)

kaikreuzer:
> When writing the above comment I thought (3) is a good middle path without any drawbacks

Imho the drawback is that we won't easily find all the places where we "temporarily allowed" that - and  specifically, it encourages others to do it the same way and that's what we should avoid as it defies the versioning approach.

> It happens that I do not react to messages including exclamation marks

You mean the small single one from me [above](https://github.com/openhab/openhab-docs/pull/518#issuecomment-332848645)? Then never mind, because it was not targeted at you at all - its only purpose was to get @hreichert's attention in time before he puts efforts into changing the PR in a way that we eventually do not want. So I hope this use of an exclamation mark finds your acceptance ;-)

ThomDietrich:
No problem. Messages can easily misinterpreted and I think everything is settled. Let's continue over in #520 !

skatun:
I spent 20 minutes figuring out why legend did not work on 2.0 :)

ThomDietrich:
@skatun ouch :) Two questions though: Why are you not using 2.1 and Do you know that by now actually have a versioning system on the documentation page?

skatun:
Why are you not using 2.1 : Because you said not to upgrade because of some error in the old version:
https://github.com/openhab/openhabian/issues/214

So I am waiting for the DMX to be implemented before I start a fresh start, is the GPIO also fixed now?(So that we do not run as root anymore)

I am renting out one room, so I can not afford the system to go down on upgrade!

ThomDietrich:
This is an old comment and 100% directed at openHABian, not openHAB 😃 
You should be safe to upgrade openHAB as long as you remember to [this](https://github.com/openhab/openhabian/issues/147#issuecomment-312449656) first. You should consider to reinstall your openHABian system and if you wait a few days you can even profit from the latest step forward, see https://github.com/openhab/openhabian/issues/203

Sorry for the DMX thing but you know that you can always add a Pull Request, right? I've btw just recently changed the openhabian structure, making everything easier to understand and use: https://github.com/openhab/openhabian/pull/241
GPIO is fixed since the switch to Raspbian in March.

Remember: Additions and improvements to the openhabian-config menu can also be used by older installations. You just have to "Update" to receive the latest version.

If you can't afford the downtime, how about a second RPi which you can exchange after everything is setup? Just an idea ;)

If you have further questions, please contact me privately rather than in this Issue. Best! Thomas

PR #522, opened 2017-10-10
Add naming convention to bundles

### LINKED ISSUE #165: [Admin/Bundles] naming convention of an addon
Add the following description to "Advanced Functionality", wherever fitting.
https://community.openhab.org/t/addons-cfg-proper-way-to-discover-the-name-of-an-addon/17107/3

Confectrician:
- Added a specific Language (due to remark/lint rules) to all code blocks, since I edited the page anyway.
- Added the naming convention paragraph

Closes #165

### DIFF
--- a/administration/bundles.md
+++ b/administration/bundles.md
@@ -13,7 +13,7 @@ The [Karaf console](#console) offers various commands to manage bundles. Most of
 
 The _bundle:list_ command returns a list of all currently installed bundles including their version.
 
-```python
+```text
 openhab> bundle:list
 START LEVEL 100 , List Threshold: 50
  ID | State    | Lvl | Version               | Name
@@ -31,7 +31,7 @@ START LEVEL 100 , List Threshold: 50
 
 Stoping a bundle is done with the command _stop_ and the ID of the bundle:
 
-```python
+```text
 openhab> bundle:stop 209
 openhab> bundle:list
 ...
@@ -40,7 +40,7 @@ openhab> bundle:list
 
 The _start_ command works accordingly:
 
-```python
+```text
 openhab> bundle:start 209
 openhab> bundle:list
 ...
@@ -51,7 +51,7 @@ openhab> bundle:list
 
 Bundles are named according to the following convention.
 
-```java
+```text
 <prefix>-<type>-<id>
 ```
 
--- a/administration/bundles.md
+++ b/administration/bundles.md
@@ -49,7 +49,7 @@ openhab> bundle:list
 
 ## Naming Convention For Bundles
 
-Bundles are named according to the following convention.
+Bundles are named according to the following convention:
 
 ```text
 <prefix>-<type>-<id>
@@ -58,11 +58,11 @@ Bundles are named according to the following convention.
 where
 
 - **prefix** is the first element to categorize the bundle.
-For addons this is often `openhab` or `esh`. 
+  For addons this is often `openhab` or `esh`. 
 - **type** is the add-on type, e.g. "binding" or "ui"
 - **id** is the identifier for this bundle
 
-The naming convention has many effects you may haven't noticed before, but used already.
+The naming convention has many effects you may not have noticed before, but used already.
 For example in the Paper UI the *sorted tabs* for all bindings are generated automatically based on the bundle naming.
 
 In many places (e.g. logging) you will also have the **package namespace** as an identifier.
--- a/administration/bundles.md
+++ b/administration/bundles.md
@@ -13,7 +13,7 @@ The [Karaf console](#console) offers various commands to manage bundles. Most of
 
 The _bundle:list_ command returns a list of all currently installed bundles including their version.
 
-```
+```python
 openhab> bundle:list
 START LEVEL 100 , List Threshold: 50
  ID | State    | Lvl | Version               | Name
@@ -31,7 +31,7 @@ START LEVEL 100 , List Threshold: 50
 
 Stoping a bundle is done with the command _stop_ and the ID of the bundle:
 
-```
+```python
 openhab> bundle:stop 209
 openhab> bundle:list
 ...
@@ -40,7 +40,7 @@ openhab> bundle:list
 
 The _start_ command works accordingly:
 
-```
+```python
 openhab> bundle:start 209
 openhab> bundle:list
 ...
--- a/administration/bundles.md
+++ b/administration/bundles.md
@@ -46,3 +46,20 @@ openhab> bundle:list
 ...
 209 | Active   |  80 | 2.0.0.b3              | Network Binding
 ```
+
+## Naming Convention For Bundles
+
+Bundles are named according to the following convention.
+
+```java
+<prefix>-<type>-<id>
+```
+
+where
+
+- **prefix** is ALWAYS "openhab" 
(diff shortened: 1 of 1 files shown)


administration/bundles.md
```diff
@@ -40,9 +40,26 @@ openhab> bundle:list
...
+```java
+<prefix>-<type>-<id>
+```
+
+where
+
+- **prefix** is ALWAYS "openhab" 
+- **type** is the add-on type, e.g. "binding" or "ui"
+- **id** is the identifier for this bundle
+  (It is also the id that you would have to use openHAB 1.x in addons.cfg, e.g. "hue", "map", etc.)
```
bgilmer77: Consider, "This is the id you would have used in the openHAB 1.x addons.cfg file (e.g. "hue", "map", etc.)"

administration/bundles.md
```diff
@@ -40,9 +40,26 @@ openhab> bundle:list
...
+
+  This is used for example to assign them the correct features for the karaf console or to sort them in different tabs within the Paper UI.
```
bgilmer77: Consider, "The bundle identifier may be useful to assign the correct features in the karaf console, or to sort the bundle into different tabs within the PaperUI."

If this is not what is meant,  consider revising the text.  The example as written is not clear.

administration/bundles.md
Confectrician: I think the id is used for many other things internally, so putting the focus only on the 1.x usage could be misunderstood.

@ThomDietrich  wdyt?

administration/bundles.md
Confectrician: Hm.

The intention was to give at least some examples why the naming convention is existing.
I think automatic sorting in Paper UI is easier to grasp than the Karaf thing.
(Especially for beginners, who are browsing through the docs.)

Maybe i should change it to one Paper UI example and clarify that there are many other usecases.

Wdyt?

bgilmer77: Yep - one solid example is enough, and the one about sorting stuff in a UI is easily understood.  Beginners may not yet even be aware... "What is this karaf console of which you speak?".  Boom - gone for an hour of Google searching, only to figure out that, in the early stages, they don't need to know anything about it.

administration/bundles.md
bgilmer77: I know you ask @ThomDietrich , but my two cents are as follows... "OH1.x??  What's that?".  OH1 is in the rear view mirror, and getting farther away all the time.  At some point we would remove the reference anyway.  I don't see this adds value, and it potentially adds confusion for someone who is just coming on board.

Confectrician: I am asking him for a third opinion. :)

After reading your 2 cents my suggestion would be to remove that sentence completely from the beginning.

Confectrician:
Changes done, here we go. :)
Sorry for the delay.

administration/bundles.md
ThomDietrich: I agree to remove that additional detail all together.

Confectrician: Already done in the last commit. :)

ThomDietrich:
Heyho!

Why python? It adds some color but the safe alternative is always `text`.

Regarding the bundle names: The relation doesn't get clear as you didn't mention `feature:list`. Also I believe the first part is not ALWAYS openhab, it's just always openhab for openHAB bundles 😊 Besides that the convention is only half the solution as in many places (e.g. logging) you'll find the package namespace of the bundle. The following command could be mentioned:

```text
openhab> bundle:list -s
START LEVEL 100 , List Threshold: 50
 ID | State    | Lvl | Version                | Name                                                   | Symbolic name
-----------------------------------------------------------------------------------------------------------------------------------------------------------------
206 | Active   |  80 | 2.1.0                  | Exec Binding                                           | org.openhab.binding.exec
207 | Active   |  80 | 1.10.0                 | openHAB Expire Binding                                 | org.openhab.binding.expire
208 | Active   |  80 | 2.1.0                  | Homematic Binding                                      | org.openhab.binding.homematic
209 | Active   |  80 | 1.10.0                 | openHAB MQTT Binding                                   | org.openhab.binding.mqtt
210 | Active   |  80 | 1.10.0                 | openHAB Mqttitude Binding                              | org.openhab.binding.mqttitude
211 | Active   |  80 | 2.1.0                  | Network Binding                                        | org.openhab.binding.network
```

Confectrician:
> Why python?

Because java and evrything else, are doing nearly the same highlighting.
`text` is fone for me too.

> Also I believe the first part is not ALWAYS openhab, it's just always openhab for openHAB bundles 😊

Then my question would be:
Are we going to mention that always is openhab for openHAB bundles or are we going to mention something generic for all kind of bundles?

> The relation doesn't get clear as you didn't mention `feature:list`.
What do you have in mind here?
Just a short sentence that delimits bundles from features?

> Besides that the convention is only half the solution as in many places (e.g. logging) you'll find the package namespace of the bundle. The following command could be mentioned:

Good point, i will add this with your example table.

ThomDietrich:
I'd say let's go with something generic and a second part like "for addons this is often openhab or esh"

Do that, just know that the table is a random grab. maybe use one that matches the other examples?

Confectrician:
Any suggestions for the `feature:list`.
I am not getting something useful in my mind. (Maybe tomorrow. :D)

bgilmer77:
Okay - so question.  I am seeing ```java, ```python, ```text, ```elephant (okay - maybe I made up the elephant part).  Are we going to pick one and normalize everything to that?

Confectrician:
I think we could at least centralize **logging parts** to `text´, that would be an effort.
I use all the other languages on a personal impression of how it looks, when I generate the page.

Confectrician:
> The relation doesn't get clear as you didn't mention `feature:list`.

> What do you have in mind here?
Just a short sentence that delimits bundles from features?

> Any suggestions for the feature:list.
I am not getting something useful in my mind. (Maybe tomorrow. :D)

Still no idea of a useful sentence or paragraph.
I think this should be in one of the first sentences and differentiate between features as an output of all available bindings/bundles and bundles as installed/active features.

But i am really not sure if I got the core here.
Some help and Ideas from your side would be lovely. :)

ThomDietrich:
Hey @Confectrician and @bgilmer77! I'm honestly happy about your recent engagement in the documentation endeavor 🤗 keep up the spirit!

Regarding language indication: First thing we need to remember is, that we want a language for every code block. Our page uses the Kramdown Markdown engine with Rouge as the syntax highlighting component: http://rouge.jneen.net
Now ideally we would use `java` for java code, `python` for python code, and so on...
Sadly many code blocks in documentation are not of one predefined language. openHAB rules for example are constructed based on a custom Xtext DSL. Other snippets like logging lines are simply no code. Parts of code can also be misinterpreted using the wrong language. In these cases we need to look for the best match and that's no science to be honest. In most cases I 'd say "go with the flow". Check the existing configuration articles for examples (Items: java/javascript, Sitemaps: perl, Rules: java or php). As for non-code snippets: Sometimes a programming language might fit but most of the time the non-highlighting option of `text` is the best we can do.

ThomDietrich:
@Confectrician let's merge this now and add further improvements later. If you can summarize the missing aspects, please create an issue! Thanks and Thanks!

Issue #524, opened 2017-10-11
persistence.md should be rewritten


bgilmer77:
persistence.md is incomplete and references the old wiki article.  The page should be updated, and information from the wiki article should be incorporated into the new persistence.md where appropriate.

PR #525, opened 2017-10-11
Update persistence.md

bgilmer77:
-  Updates persistence.md by adding relevant information from the old wiki article.
-  Removes references to the old wiki article
-  Addresses #524

Signed-off-by: Brad Gilmer <brad@gilmer.tv> (github:bgilmer77)

### DIFF
--- a/configuration/persistence.md
+++ b/configuration/persistence.md
@@ -7,12 +7,238 @@ title: Persistence
 
 # Persistence
 
-Persistence support stores item states over time (a time series).
-openHAB is not restricted to a single data store.
-Different stores can co-exist and be configured independently.
-For further details, please check:
+## Introduction
 
-- The list of supported [Persistence Services]({{base}}/addons/persistence.html)
-- [The openHAB 1.x wiki persistence article](https://github.com/openhab/openhab/wiki/Persistence)
+openHAB has the ability to save item states according to rules you have configured.
+This is called persistence.
+Persistence can have several important uses.
+For example, persistence allows an openHAB system to restore itself to the state it was in prior to a restart (if you have configured it to do so).
+Persistence may also be used to store values that are subsequently displayed in a graph.
+(For an excellent graphing tutorial see [InfluxDB+Grafana persistence and graphing](https://community.openhab.org/t/influxdb-grafana-persistence-and-graphing/13761)
+
+openHAB persists item states in a database, and most popular databases are supported.
+You may have more that one persistence add-on loaded, and each of these may be configured independently.
+Examples of available persistence add-ons include relational databases, NoSQL databases, round-robin databases, Internet-of-Things (IoT) cloud services, simple log files, etc.
+All of these options are available in openHAB, and they are all configured in the same way.
+Note that the type of database(s) you use for persistence may be influenced by your requirements for exporting data (e.g. IoT services or log files), or your requirement to make queries against the database to retrieve historical data (e.g. MySQL and MongoDB).
+A complete list of supported persistence add-ons may be found in the [persistence]({{base}}/addons/persistence.html) section of the on-line openHAB documentation.
+
+## Persistence Add-on Configuration
+
+Each persistence add-on you install will need to be configured.
+Please refer to the specific [on-line documentation]({{base}}/addons/persistence.html) for your selected persistence add-on for configuration instructions.
+
+## Default Persistence Service
+
+You may install more than one persistence add-on.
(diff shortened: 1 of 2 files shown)


configuration/persistence.md
```diff
@@ -7,12 +7,238 @@ title: Persistence
...
+You may install more than one persistence add-on.
+Therefore, it is important to select a default persistence service.
+You should do this even if you have only one persistence add-on installed.
+
+To select a default persistence service, in paperUI, select Configuration and then System from the side menu.
+Scroll down to "Persistence", and select your Default Service from the drop-down list.
+Note that you must first install a persistence add-on before you make this selection.
+Be sure to save your choice once you have selected your default service.
+
+## configuration
```
Confectrician: Uppercase -> Configuration

configuration/persistence.md
```diff
@@ -7,12 +7,238 @@ title: Persistence
...
+
+You can choose how your persistence service operates, which item states it persists, and under what conditions it persists them.
+These things and more are configured in a file named `<persistenceservice>.persist`, where "persistenceservice" is replaced by the name of your add-on (e.g. `rrd4j.persist`).
+This folder is located in `$OPENHAB_CONF/persistence`.
+The configuration defines persistence "strategies" which are very similar to [triggers]({{base}}/configuration/rules-dls.html#rule-triggers) in openHAB rules.
```
Confectrician: The link throws an 404 not found error.

configuration/persistence.md
```diff
@@ -7,12 +7,238 @@ title: Persistence
...
+<item>.deltaSince(AbstractInstant) - Gets the difference value of the state of a given item since a certain point in time.
+<item>.previousState() - Retrieves the previous item (returns HistoricItem).
+<item>.previousState(true) - Retrieves the previous item, skips items with equal state values and searches the first item with state not equal the current state (returns HistoricItem).
+<item>.sumSince(AbstractInstant) - Retrieves the sum of the previous states since a certain point in time. (OpenHab 1.8)
+```
+
+These extensions use the default persistence service that is configured as the default persistence service.  (Refer to Default Persistence Service above to configure this.)
+
+Note that you can specify that a different persistence service be used with a particular extension.
+Do this by appending a String as an optional additional parameter at the end of the extension (e.g. "rrd4j" or "sense")
```
Confectrician: Maybe an example would be good here.

configuration/persistence.md
```diff
@@ -7,12 +7,238 @@ title: Persistence
...
+  touch $f
+done
+```
+
+The script waits for five seconds and then touches all `*.rules` files (except `refresh.rules`).
+This causes openHAB to reload all rules (openHAB automatically reloads rules when their creation date/time changes).
+Other rules files may be added on new lines.
+Note - you will have to experiment to find the appropriate sleep value for your specific system.
 
 {% include contribution-wanted.html %}
```
Confectrician: Idk if we need this explicit after this huge improvement.

configuration/persistence.md
Confectrician: ah i found the issue:
"dls" must be "dsl" 

-> [triggers]({{base}}/configuration/rules-dsl.html#rule-triggers)

bgilmer77: Agree - thank you and well spotted.  I will change this.

configuration/persistence.md
bgilmer77: I agree, and I was not aware that one could do this.  I have a development system in my office, and if I get time, I can check it out by installing another persistence add-on and then checking that it works as expected.  From what was in the wiki, I believe the syntax would be to add something like ".rrd4j" at the end of the function, but this should be verified before we publish an example.  For now I will add the following:

configuration/persistence.md
bgilmer77: Agree - I noticed that it would have been good to delete this once I had posted the PR.  I will delete this call for contributions.

bgilmer77:
Okay - sorry to be making such a mess of things.  Still learning my way around GitHub.  When I made the last commit, " Sitemaps: Improves readability", I was on the incorrect branch locally.  This commit needs to be deleted.  Help please.

bgilmer77:
Okay - what I have attempted to do is to revert the sitemaps.md file back to its original state. I am going to stop at this point.  If I can do something to fix this in a better way, please let me know.

bgilmer77:
I am deleting the branch on my repo and closing this PR.  I will create a new PR, starting over with my new additions to the article.

PR #532, opened 2017-10-20
Sitemap: Added MapView + Corrected Mismatching Screenshots

### LINKED ISSUE #492: Sitemaps: Add MapView Sitemap element

Confectrician:
Closes #492 

### Maybe something to discuss about

The table of contents (generated on the headline order) and the element type overview table have non matching orders.
Maybe we should order them in a consistent way directly.

So there are 2 options:

1. Change the element type table according to the real content
2. Change the content according to the (already alphabetic) table

I would prefer 2.
But maybe there is some logic behind the current situation, which i don't understand?!

Signed-off-by:  Jerome Luckenbach <github@luckenba.ch>


Confectrician:
Hm there seems to be something wrong with my commit history.
Maybe i didn't checkout this branch from gh-pages.

I will close this one and then start a corrected PR based on the current gh-pages commit.

ThomDietrich:
Okay.

Regarding the question: At first I built the table in the same order as elements occur in the article (which is based on popularity). In the last revision of the article I then reordered the table to be in alphabetic order. I do not see an issue here. However if you guys feel this is irritating I would actually prefer your option 1. I think popularity of elements is more pressing then an alphabetic sorting. @bgilmer77 your opinion?

PR #528, opened 2017-10-13
Sitemaps: improve comprehension

bgilmer77:
- made changes to improve comprehension in introduction
- will make additional commits as I work through the entire article

Signed-off-by: Brad Gilmer <brad@gilmer.tv> (github: bgilmer77)

PR #542, opened 2017-10-24
Additional improvements to sitemaps.md

bgilmer77:
-  Addresses most of @ThomDietrich comments in #528
-  See #528 for open items and questions

Signed-off-by: Brad Gilmer <brad@gilmer.tv> (github: bgilmer77)

### DIFF
--- a/configuration/sitemaps.md
+++ b/configuration/sitemaps.md
@@ -567,16 +567,6 @@ Note that expressions are evaluated from left to right; the first matching expre
 If the order of the expressions was reversed, the color assignment would not work properly.
 Note also, the effect of omitting `Temperature` and the comparison operator in the expression `0="white"` (as compared to `==0="white"`).
 
-The two lines below show that `item_name` may be omitted, in which case, the name from the previous line will be used:
-
-```perl
-Text item=Weather valuecolor=[Temperature<=4="blue"]
-                  labelcolor=[Temperature<=4="blue"]
-```
-The `Weather` Item will be blue (label and value) if `Temperature` is less than or equal to 4°C.
-
-![Presentation of the color parameters in BasicUI](images/sitemap_demo_colors.png)
-
 Below is a list of standard colors and their respective RGB color codes.
 
 | Color Name  | Preview and RGB Color Code              |
--- a/configuration/sitemaps.md
+++ b/configuration/sitemaps.md
@@ -112,7 +112,7 @@ The following element types may be used in a Sitemap definition file.
 |-------------------------------------------|-----------------------------------------------------------|
 | [Chart](#element-type-chart)              | Adds a time-series chart object for [persisted](persistence.html) data. |
 | [Colorpicker](#element-type-colorpicker)  | Allows the user to choose a color from a color wheel. |
-| [Default](#element-type-default)          | Renders an Item in the default UI representation specified by the type of the given item. |
+| [Default](#element-type-default)          | Renders an Item in the default UI representation specified by the type of the given Item. |
 | [Frame](#element-type-frame)              | Establishes an area containing various other Sitemap elements. |
 | [Group](#element-type-group)              | Concentrates all elements of a given group in a nested block. |
 | [Image](#element-type-image)              | Renders an image given by an URL. |
@@ -146,7 +146,7 @@ This provides the flexibility to present Items in the way desired in your home a
     - `label` sets the textual description displayed next to the preprocessed Item data (e.g. "`Now [%s °C]`"), [more details](items.html#item-label).
     - `icon` chooses the name of the icon file to show next to the element, [more details](items.html#icons).
 
--   When an Item is defined, you have the opportunity to assign a label and/or an icon at that point.
+-   When an [Item]({{base}}/configuration/items.html) is defined, you have the opportunity to assign a label and/or an icon at that point.
 If no label or icon are specified in the Sitemap, then the label and/or icon you assigned to the Item will be displayed.
 Setting a value for `label` or `icon` of a Sitemap element will override the values defined for the linked Item.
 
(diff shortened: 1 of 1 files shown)


bgilmer77:
@ThomDietrich please see #528.  Do you want me to re-enter your unresolved comments on this PR, or is it okay if we continue to work from #528?

ThomDietrich:
Hey! Thanks for your eager continuation on the issue ;) I've answered hopefully all your questions. I believe it is safe if you now just do how you see fit and then we can finishup the sitemap article and look at the next task :)

bgilmer77:
@ThomDietrich there are only a small number of open items from #528, so I am copying them here to make it easier to see what still needs to be resolved.  Please refer to #528 for the complete thread:

**Open item 1**
I have a development system here, so no problem to do a test. But what do you suggest? Is your point to test something like a Switch Item to see if the Setpoint can be used to control it?

**Open item 2**
IMPORTANT - is this sentence, which is currently submitted as a change, correct?

"It is important to note that visibility may only be determined based on one condition at a time."

FWIW, if it is correct, I would reword this to, "It is important to note that visibility is determined based upon only one condition at a time."

**Open item 3**
Actually, I quite like the idea of leaving "bread crumbs" for people who may be trying to do something more than just the basics. Many times I find that all I need is a small hint about how to attack a more complex issue and I can take it from there.

How about we change the "To achieve more complex..." sentence to:

If you are trying to do something more complicated than these simple examples, then you could define an additional Item which is set by a Rule.
Rules have a rich set of features that accommodate more complex scenarios.

ThomDietrich:
Heyho, good idea.

**1**

Yep sounds about right. See what happens if you link a slider to a switch. 

In the end it might be good to replace "The element is often used to gradually change a number item." with something more specific or remove it all together. A user should have a natural feeling which item can be used with which sitemap element in a meaningful may.

**2**

The example and the description was correct. In my [comment](https://github.com/openhab/openhab-docs/pull/528#discussion_r145164174) I wanted to suggest something that is maybe easiest explained by a resulting text:

> ...
> It is important to note that visibility is determined based upon only one condition at a time. If *any* one of the comparisons is evaluated as `true`, the Item will be visible, otherwise it will be hidden.
>
> To give an example: The third line above might control the visibility of a lawn sprinkler control. The control will be visible if it is Morning *OR* if it is Afternoon *OR* if the temperature is above 19°C. A combination of multiple conditions, like Morning *AND* above 19°C, is not supported in the visibility definition.
>
> Of course more complex visibility conditions are possible by, e.g., defining an additional Item which is set by a Rule. Rules have a rich set of features that accommodate more complex scenarios. 

**3**

I've tried to add a sentence above...

bgilmer77:
Breaking apart my responses so we can track progress separately...
**1**
Okay - I will give this a try when I get a chance.

bgilmer77:
**2**
Okay - I will put the original example back and then look at your text and go from there.

bgilmer77:
**3**
Once I put the example back, I will look at your sentence above and incorporate it.

bgilmer77:
**1**
I have tested this now - linking a switch to a Slider or Setpoint does not work.  I will remove the sentence.  The use case for a setpoint is clear without it.

bgilmer77:
**2**
I will put your wording referencing the third example back.

bgilmer77:
**3**
I will incorporate your rewording as in **2** above.

bgilmer77:
@ThomDietrich I believe this is now ready to merge.  Please double-check.

bgilmer77:
Ach!  Keep forgetting to sign off on commits.  Will work to improve this.

Confectrician:
@bgilmer77 You can sign off commits very easily with the command line.

`git commit -s -m "YOUR COMMIT MESSAGE GOES HERE"`

The sign off part wil be added automatically for you.
You just have to config the correct email and name once in your git configuration.

bgilmer77:
Thanks - problem is I am using GitDesktop.  It may do automatic signoffs, but I have not found that capability yet.

configuration/sitemaps.md
```diff
@@ -555,31 +546,34 @@ If an operator is not specified, the operator will default to `==`.
...
+The two lines below show that `item_name` may be omitted, in which case, the name from the previous line will be used:
 
 ```perl
-Text item=Temperature labelcolor=[>0="blue"] valuecolor=[22="green"]
-Text item=Temperature labelcolor=[>0="blue"] valuecolor=[==22="green"]
-Text item=Temperature labelcolor=[>0="blue"] valuecolor=[Temperature==22="green"]
+Text item=Weather valuecolor=[Temperature<=4="blue"]
+                  labelcolor=[Temperature<=4="blue"]
 ```
+The `Weather` Item will be blue (label and value) if `Temperature` is less than or equal to 4°C.
```
ThomDietrich: Extra empty line missing

configuration/sitemaps.md
```diff
@@ -555,31 +546,34 @@ If an operator is not specified, the operator will default to `==`.
...
 
 ![Presentation of the color parameters in BasicUI](images/sitemap_demo_colors.png)
```
ThomDietrich: Does the example still make sense? Seems like it doesn't fit.

configuration/sitemaps.md
```diff
@@ -555,31 +546,34 @@ If an operator is not specified, the operator will default to `==`.
...
+                                  >=25="orange", >=15="green", 0="white", <15="blue"]
+```
 
-In the second example, the effect of omitting the `item_name` and `operator` parameters is shown by omitting `Temperature` and by omitting the comparison operator in the expression `0="white"` (as compared to `==0="white"`).
+Note that expressions are evaluated from left to right; the first matching expression determines the color.
+If the order of the expressions was reversed, the color assignment would not work properly.
+Note also, the effect of omitting `Temperature` and the comparison operator in the expression `0="white"` (as compared to `==0="white"`).
 
-The following three lines are equivalent, and all of them are valid:
+The two lines below show that `item_name` may be omitted, in which case, the name from the previous line will be used:
```
ThomDietrich: I don't get it. Isn't the item name given (Temperature)?

ThomDietrich:
Hey @bgilmer77, I like the latest changes! Of course I once again found something to complain about 😄

configuration/sitemaps.md
bgilmer77: Agreed

configuration/sitemaps.md
bgilmer77: Agree.  See below.

ThomDietrich:
@bgilmer77 I can recommend SmartGit as a powerful free git GUI client

I'm okay with your proposed solution.

configuration/sitemaps.md
bgilmer77: Text, example and graphic have been removed.  This comment is addressed.

bgilmer77:
Okay - I think all comments have been addressed at this point?

Thanks for the pointer to SmartGit.

bgilmer77:
One benefit from running jekyll locally - found an error the linter did not catch.  Omitted double curly brackets on a link.  Fixed with this latest commit.

bgilmer77:
Okay - definitely fixed this time...

Issue #520, opened 2017-10-05
Implement versioned docs


BClark09:
Hi all, I found no issue discussion on how the docs would be separated into versions, but it comes up in discussion from time. I've done some playing around based on [this website](https://justwriteclick.com/2017/07/30/investigating-jekyll-for-versioned-content/):

![image](https://user-images.githubusercontent.com/8983584/31221175-b7b4229c-a9ba-11e7-921e-09dfe1549d8e.png)

The above site is built with no modifications, but with the following script: (For an actual server, you'd only need to build, not perform the initial serve.)

```sh
#!/bin/sh
git fetch --tags                                                            ## Currently, releases are kept as tags. Not stored in branches.
git checkout gh-pages                                                       ## Checkout the latest changes first.
jekyll serve --incremental --detach                                         ## A build will be fine, for an actual server.
tagsList="$(git tag --list [0-9].[0-9].[0-9])"                              ## Iterate through versions only.
echo "$tagsList" | while read version                                       ## For each version tag...
do
  git checkout "tags/$version"                                              ## Set contents to version.
  echo "baseurl : /$version" > _config.$version.yml                         ## Programatically make a new config
  jekyll build --config _config.yml,_config.$version.yml -d _site/$version/ ## Store the contents in [ip]:[port]/[version]/
done
git checkout gh-pages                                                       ## Continue with master branch.
```

Based on the above, I have the following suggestions:

 - Make a branch for each version, instead of a Tag, so that changes can be backported, (e.g. adding a dropdown to each version).
 - `{% assign base = site.baseurl %}` needs to be used in `_includes/base.html`.

Would this be the best way of doing it? Did anyone else have suggestions on how to implement it?

ThomDietrich:
Hey @BClark09,
you should have asked before going through all that trouble. I've got something pretty similar in a test-setup of mine :) I wanted to finish #516 before continuing. Nonetheless, this looks pretty good!! 👏 

You didn't find any discussion on the issue as this was so far only discussed in a non-public thread (which you are part of I believe). In this thread we were discussing different ideas with @kaikreuzer and @martinvw. The originally proposed solution was less elegant and I've since realized that it is not applicable.

I'm currently in the office and can't compare with my solution but the key ideas seem to be the same.
[...]

PR #552, opened 2017-11-05
Add precompiled v2.1.0 stable docs version

ThomDietrich:
Following the principle: Better than nothing.
This is a first approach in solving #520 here and now.
The PR includes a precompiled version of tag 2.1.0 as the new subfolder "2.1.0".
The PR also includes buttons to switch the version shown.

Modifications applied to 2.1.0 (will need to be addressed in the form of version-branch commits later):
- _config.yml base change
- _includes/base.html base change
- usermenu change for generated addons urls

Known limitation: Generation of the button link is derived from the current URL. If the location of an older article was different or the article didn't exist yet, a 404 error will be shown (e.g. https://docs.openhab.org/2.1.0/addons/bindings/some-brand-new-binding/readme.html).

I'm not striving for perfection with the PR. We should still discuss better options for the long run. See #520 

grafik

grafik


kaikreuzer:
Good start!
A few (early, but late at night ;-) ) comments:
- We should probably have some direct links to the different available versions on the main entry page of the docs website (over time, there will be more than latest release + snapshot).
- The "version button" imho takes too much precious space. I would prefer a place where it is not on the content pane, but rather in the header or the side menu (I assume it is more difficult to integrate it there, though). If we have links on the main page, I am actually not sure whether we need to support switching between versions on the sub-pages directly. Maybe we could add a toast 
that only briefly mentions the version to the user and then silently disappears again?
- I assume we only address the user guide, but not the tutorials or the developer guide (although in theory, they might be subject to versioning as well). Is this our common understanding?

ThomDietrich:
Hey Kai,
as I said, this is a first PR to get a 2.1.0 stable version of the documentation into place. Would you be okay to merge now and add improvements later?

Regarding your comments: 

* Yes, next PR
* Moving it somewhere else would actually be quite easy. However as soon as I had placed it into the top menu, we would once again have issues with the mobile view of the page. I can certainly look into a better fit but (1) that can be part of a later PR and (2) I'd rather see effort spent on this in the new webdesign instead of wasted time now...
* These are details we would still need to discuss. The current implementation includes the whole 2.1.0 content but only shows the switching button for articles of the User Manual.

kaikreuzer:
> Would you be okay to merge now and add improvements later?

Yes, but my main concern is that the button is simply huge (also compared to the menus and other parts) and thus adds a lot of white space that is wasted on small screens.
If you could reduce it to 50% of its size, I could live with it until a website relaunch - so this is the only improvement I'd ask you for before merging, ok?

BClark09:
Apologies for my delay in reply. It's been a busy period, but can see you've been busier 😃 !

I think this is as best as we can do whilst hosting on GitHub, I haven't seen any other way of doing this so the changes look good to me. A couple of non-blocking points:

- Is there a script that you used to include the new directory? Can it be included in the repo as a site-ignored file?
- I agree that the versioning button is rather on the big side but can be dealt with easily, perhaps above the left hand column this way, it should look fine on mobile (and still be at the top).
- Definitely for later, but I think we could make do with removing the versioned "Developers" and (maybe less so) the "Tutorials" sections?

ThomDietrich:
Hey guys,
I've invested a few more hours to fine-tune everything.

* I created a branch called "v2.1.0-patch". This one follows the idea we @BClark09 already discussed. There were more inconsistencies than I thought in the v2.1.0 AND current snapshot revisions I needed to sort out. The new branch consists of all these corrections (see [latest v2.1.0-patch commits](https://github.com/openhab/openhab-docs/commits/v2.1.0-patch)). The same set of corrections is also included in this PR against `gh-pages`. This WAS a very cumbersome task but at least now the expected effort for v2.2.0 is small.
* I've now excluded Tutorials and Developers. Links to these areas are always version-less. One "secret feature" I believe is good to have: If you specifically browse http://docs.openhab.org/v2.1.0/tutorials/beginner/index.html you can still see the older article revisions and the version-specific menu.
* I've replaced the Switching button with a Dropdown menu, preparing for v2.2.0
* I did NOT move or resize the button. It's a button following the Materialize Style Guide. The position right next to the article (the part of the page that provides versioning) makes sense. The space used is acceptable, also on mobile. Sounds (and looks) good to me. I did replace the blue color by grey though. So far this is my opinion, please remember that we have the chance to add a better position and style with the new page design. I am not going to invest more time on this here and now in this PR.
* As requested by @BClark09 I've added a [chapter to the README](https://github.com/openhab/openhab-docs/commit/ca55026576ad819dd057fe48dccb05ed0025e011) regarding the versioning design.

The whole PR is tested and ready to merge.

BClark09:
Nice! I think this is good now. I can't press the approve button just yet because my browser crashes when Im on the review page. 😐

Will try to use my main PC tommorow evening in which I can also test what it looks like serving to a couple of mobile devices.

kaikreuzer:
Let's merge then - many thanks @ThomDietrich!

PR #553, opened 2017-11-06
fix links to Homekit

B4dT0bi:
I've found some links getting 404 so I've fixed it.

### DIFF
--- a/_addons_ios/alexa-skill/readme.md
+++ b/_addons_ios/alexa-skill/readme.md
@@ -28,14 +28,14 @@ Currently supported are the English (U.S.), English (U.K.), and German language.
 
 ### Requirements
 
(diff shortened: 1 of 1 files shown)


ThomDietrich:
Hey @B4dT0bi,
Thanks! The change from "io" to "ios" was recent and seems like we missed that.

Many parts of the documentation are imported from other openHAB repositories, e.g., the Binding- or add-on specific repository. This way we ensure that the articles are always in sync with the actual source code.
Long story short, you'll need to propose your change in the correct repository: 
https://github.com/openhab/openhab-alexa/blob/master/USAGE.md

B4dT0bi:
created the pr in the right repo now, thanks for clarification
Link : [new PR](https://github.com/openhab/openhab-alexa/pull/50)

Issue #479, opened 2017-08-19
Add notification actions to documentation


metbril:
The documentation is currently missing proper documentation for sending notifications from rules through the official openHAB mobile apps. At least, I have not been able to find it.

The wiki used to have it listed here: https://github.com/openhab/openhab1-addons/wiki/Actions but that page https://my.openhab.org/docs/notifications is now 404.

- `sendNotification(<user>, <message>)`
- `sendBroadcastNotification(<message>)`
- `sendLogNotification(<message>)`

I am not sure where it should be, since it is only valid when using the openHAB Cloud instance from the Foundation.

Possible pages to add to:

http://docs.openhab.org/addons/io/openhabcloud/readme.html
http://docs.openhab.org/addons/actions.html

ThomDietrich:
@maziman wdyt?

Confectrician:
> since it is only valid when using the openHAB Cloud instance from the Foundation

In this case i would prefer:
http://docs.openhab.org/addons/io/openhabcloud/readme.html

metbril:
_Sorry for not responding sooner, but my github notifications are not working, I am not getting any email from them._

Would you accept a PR for that page? I am happy to contribute.

marziman:
I am not completely sure if OH cloud is the right place. Just because OH cloud is in the chain, it doesnt need to be the "right place".
Can you explain shortly would you have in mind?

Thanks alot
BR Mehmet

marziman:
But: Of course your effort is important and we will find a home for it!
Thanks for your support..

metbril:
My intention is to add sufficient documentation for using above **actions** in rules e.g.

```
rule "Demo"
when
    Item Demo_Item received command
then
    sendBroadcastNotification('Demo item received command ' + Demo_Item.state.toString)
end
```

This documentation used to be available through the old wiki and the my.openhab.org site (which is offline now).

My first guess would be to add it to the actions page (like the mail action) and if needed refer and link to the openHAB Cloud.

ThomDietrich:
That indeed sounds like a good plan and would fit nicely in this list http://docs.openhab.org/addons/actions.html

Only problem we need to solve: The list in the second half of the page is automatically generated from all `README.md`'s found here: https://github.com/openhab/openhab1-addons/tree/master/bundles/action

Confectrician:
> My first guess would be to add it to the actions page (like the mail action) and if needed refer and link to the openHAB Cloud.

[...]

PR #587, opened 2017-12-12
[actions] Add notification actions

bgilmer77:
- added three notification actions
- addresses #479 
Signed-off-by: Brad Gilmer <brad@gilmer.tv>

### DIFF
--- a/addons/actions.md
+++ b/addons/actions.md
@@ -64,29 +64,14 @@ To get a list of available voices use `say -v ?` and to get a list of devices us
 
 One can configure whether specific log entries are logged out and where they get logged to (e.g. to a separate file) by [editing the logger configuration]({{base}}/administration/logging.html).
 
-### Notification actions
+### Cloud Notification Actions
 
-Notification actions may be placed in `rules` to send alerts to mobile devices registered with <https://myopenhab.org>.  Three different actions are available:
+Notification actions may be placed in Rules to send alerts to mobile devices registered with an [openHAB Cloud instance](https://github.com/openhab/openhab-cloud) such as [myopenHAB.org](https://myopenhab.org).
+Three different actions are available:
 
-- `sendNotification("your myopenHAB user email address here","message")`: Sends a notification containing 'message' to device(s) registered with your myopenHAB email address at <https://myopenhab.org>
-- `sendBroadcastNotification("message")`: Sends a notification containing "message" to *all* devices of *all* users of your myopenHAB.org account
-- `sendLogNotification("message")`: Sends a log notification containing "message" to the `notifications` list at myopenHAB.org.  Notifications are NOT sent to any registered devices
-
-**Example**
-
-The following example sends a notification once per minute to the myopenHAB app running on your mobile device.
-This rule can be useful when testing notification actions.
-
-```java
-rule "pingMe"
-
-when
-  // Every 1 minute...
-  Time cron "0 0/1 * * * ?"
-then
-  sendNotification("user@email.com","pingMe test notification")
-end
-```
+- `sendNotification("your myopenHAB user email address here","message")`: Sends a notification containing 'message'
+- `sendBroadcastNotification("message")`: Sends a notification containing "message" to *all* devices of *all* users
+- `sendLogNotification("message")`: Sends a log notification containing "message" to the `notifications` list at your openHAB Cloud instance.  Notifications are NOT sent to any registered devices
 
 For information on making use of the [openHAB Cloud service](https://github.com/openhab/openhab-cloud/blob/master/README.md) hosted by the [openHAB Foundation e.V.](https://www.openhabfoundation.org/), visit the [myopehnab.org website](http://www.myopenhab.org).
 
--- a/addons/actions.md
+++ b/addons/actions.md
@@ -64,17 +64,6 @@ To get a list of available voices use `say -v ?` and to get a list of devices us
 
 One can configure whether specific log entries are logged out and where they get logged to (e.g. to a separate file) by [editing the logger configuration]({{base}}/administration/logging.html).
 
-### Cloud Notification Actions
-
-Notification actions may be placed in Rules to send alerts to mobile devices registered with an [openHAB Cloud instance](https://github.com/openhab/openhab-cloud) such as [myopenHAB.org](https://myopenhab.org).
-Three different actions are available:
-
-- `sendNotification("your myopenHAB user email address here","message")`: Sends a notification containing 'message'
-- `sendBroadcastNotification("message")`: Sends a notification containing "message" to *all* devices of *all* users
-- `sendLogNotification("message")`: Sends a log notification containing "message" to the `notifications` list at your openHAB Cloud instance.  Notifications are NOT sent to any registered devices
-
-For information on making use of the [openHAB Cloud service](https://github.com/openhab/openhab-cloud/blob/master/README.md) hosted by the [openHAB Foundation e.V.](https://www.openhabfoundation.org/), visit the [myopehnab.org website](http://www.myopenhab.org).
-
 ### HTTP Actions
 
 - `sendHttpGetRequest(String url)`: Sends an GET-HTTP request and returns the result as a String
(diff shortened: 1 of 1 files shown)


bgilmer77:
On my system, the example code block is not rendering properly.  Instead of seeing the example between codefences, I see the actual text with the three backslanting apostrophy marks and the word "java".  But looking at the file, I cannot see anything wrong.  Perhaps it is an error in the jekyll rendering?

addons/actions.md
```diff
@@ -23,62 +23,88 @@ The following Actions are a core part of openHAB and do not need to be separatel
...
-* **Complete String Example:** `logInfo("Garage", "This is a complete String")`
-* **Concatination Example:** `logDebug("Lighting", "This is a string concatination:" + Light.name)`
-* **Arguments Example:** `logWarn("Alarm", "The {} window has been open for {} hours!", Window.name, timeOpen)`
+- **Complete String Example:** `logInfo("Garage", "This is a complete String")`
+- **Concatination Example:** `logDebug("Lighting", "This is a string concatination:" + Light.name)`
+- **Arguments Example:** `logWarn("Alarm", "The {} window has been open for {} hours!", Window.name, timeOpen)`
 
 One can configure whether specific log entries are logged out and where they get logged to (e.g. to a separate file) by [editing the logger configuration]({{base}}/administration/logging.html).
 
+### Notification actions
```
ThomDietrich: Please capitalize

addons/actions.md
```diff
@@ -23,62 +23,88 @@ The following Actions are a core part of openHAB and do not need to be separatel
...
+
+Notification actions may be placed in `rules` to send alerts to mobile devices registered with <https://myopenhab.org>.  Three different actions are available:
```
ThomDietrich: Consider Rules instead of `rules`
Replace `<https://myopenhab.org>` with `an [openHAB Cloud instance](https://github.com/openhab/openhab-cloud) like [myopenHAB.org](https://myopenhab.org)`

The exact wording is up to you ;)

ThomDietrich: To be more in line with the other sections maybe move this below the list?

addons/actions.md
```diff
@@ -23,62 +23,88 @@ The following Actions are a core part of openHAB and do not need to be separatel
...
+
+- `sendNotification("your myopenHAB user email address here","message")`: Sends a notification containing 'message' to device(s) registered with your myopenHAB email address at <https://myopenhab.org>
```
ThomDietrich: redact myopenhab here?

ThomDietrich: and from the two others

bgilmer77: @ThomDietrich sorry - not sure what you mean by redact.  Would you like me to substitute an `[openHAB Cloud instance](https://github.com/openhab/openhab-cloud) like [myopenHAB.org](https://myopenhab.org)` for myopenHAB?

addons/actions.md
```diff
@@ -23,62 +23,88 @@ The following Actions are a core part of openHAB and do not need to be separatel
...
+- `sendBroadcastNotification("message")`: Sends a notification containing "message" to *all* devices of *all* users of your myopenHAB.org account
+- `sendLogNotification("message")`: Sends a log notification containing "message" to the `notifications` list at myopenHAB.org.  Notifications are NOT sent to any registered devices
+
+**Example**
+
+The following example sends a notification once per minute to the myopenHAB app running on your mobile device.
+This rule can be useful when testing notification actions.
```
ThomDietrich: Not sure if an example is needed, all other sections don't have one. If you don't want to remove it consider moving it to the top or add a section, example and link to this page in the rules-dsl article.

ThomDietrich: Also I don't like the example as it is not really meaningful. Check other rules-dsl examples to bring it in line.

addons/actions.md
```diff
@@ -23,62 +23,88 @@ The following Actions are a core part of openHAB and do not need to be separatel
...
+
+when
+  // Every 1 minute...
+  Time cron "0 0/1 * * * ?"
+then
+  sendNotification("user@email.com","pingMe test notification")
+end
+```
+
+For information on making use of the [openHAB Cloud service](https://github.com/openhab/openhab-cloud/blob/master/README.md) hosted by the [openHAB Foundation e.V.](https://www.openhabfoundation.org/), visit the [myopehnab.org website](http://www.myopenhab.org).
```
ThomDietrich: Good.

addons/actions.md
ThomDietrich: Also Notification seems to big. How about "openHAB Cloud Actions" or "Cloud Notification Actions"

bgilmer77: +1 Cloud Notification Actions

addons/actions.md
bgilmer77: @ThomDietrich Hum - not exactly sure what you would prefer here.  Speaking personally, having recently worked on notifications using OH Cloud, Prowl and others, this rule was EXTREMELY useful to me for troubleshooting (such a simple thing, but it really helped).  I am perfectly happy to remove it entirely, or to put it in rules-dsl, or to reformat it somehow, if you are saying it is not formatted correctly.  But for a new user, or even a user who has been using OH for a while but is just getting into notifications, this is a "breadcrumb" - a hint at a super-simple but useful troubleshooting tool.  Please let me know what you would like.

addons/actions.md
ThomDietrich: If you mention openHAB Cloud respectively myopenHAB above, I'd suggest to not mention it for each action specifically.

bgilmer77: Okay - working on something else right now, so I do not exactly understand your note, but I will look at this later.  Now I think I can figure out your point and correct the article appropriately.

addons/actions.md
bgilmer77: Will change to Cloud Notification Actions in next commit

addons/actions.md
bgilmer77: Agree.  Will change in next commit.

addons/actions.md
bgilmer77: Look at what is in my next commit.  I think I have addressed your issue.

addons/actions.md
bgilmer77: bump...

bgilmer77: I agree - why provide an example here when the other sections do not have one.  I will delete it.

addons/actions.md
ThomDietrich: @bgilmer77 do you think "alert" is the same as notification? Do you think users could interpret alert as something scary?

addons/actions.md
ThomDietrich: Looks good!

addons/actions.md
ThomDietrich: I've looked at the Actions article again and think we should reconsider. How about you move this section out of "Core Actions"? Other than those, this section is only relevant for Cloud users and it is a notifications action which is quite different what core actions do.
I'd suggest to make it a top-level headline between "Core Actions" and "Installable Actions".

About the example: With the above I now also agree an example is a good idea. Sorry for the back and forth. What I didn't like about yours is, that it highlights how to troubleshoot openHAB. Ideally that is not needed. Hence my suggestion would be a production-like example. How about:


```javascript
rule "Front Door Notification"
when
    Item Apartment_FrontDoor changed to OPEN 
then
    send Notification("me@email.com", "Front door was opened!")
end
```

ThomDietrich:
@bgilmer77 I've added comments above. One extra thought: Would you like to add an introductional sentence to the Installable Actions section? Something simple like "The following actions are available as installable add-ons."

bgilmer77:
@ThomDietrich added short intro per your request. Ready to merge?

ThomDietrich:
@bgilmer77 did you overlook https://github.com/openhab/openhab-docs/pull/587#discussion_r157731654
I think both parts could be valuable... what do you think?

ThomDietrich:
@bgilmer77  friendly ping

bgilmer77:
friendlyPing_ACK/
Unfortunately, this week I am having to make money (working in my day job), so I won't be able to get back on this.  But I have made a note to dig into it next week.

ThomDietrich:
No hurry. I just wanted to remind you in case you forgot... ;)

bgilmer77:
Okay @ThomDietrich, I found some time to look into your comments.  How does this look?

bgilmer77:
@ThomDietrich Thanks for the review!

Issue #189, opened 2017-01-13
Establish fixed term placeholders for the openHAB folders


ThomDietrich:
Quoting https://community.openhab.org/t/help-im-ready-to-give-up/20006/39

Having one OS specific doc page that defines variables that are then used in the doc is another way:
For example:
debian/apt: `$CONFIG = /etc/openhab2/`
RH: `$CONFIG = /etc/somewhere/openhab2/` (example only)
non-pkg install: `$CONFIG = /opt/openhab2/conf/` (replace /opt/openhab2 with you install folder)
windows `$CONFIG = c:\programfiles\openhab2\conf` (example only)

and the in the doc it could say "edit `$CONFIG/sitemaps/_default.sitemap`"
instead of saying "edit your default sitemap".

----

This could nicely be integrated into http://docs.openhab.org/installation/linux.html#file-locations and should then be used in the community forum.

kaikreuzer:
Definitely makes sense to come up with such placeholders.
We actually even have them already technically in place, see [here](https://github.com/openhab/openhab-distro/blob/master/distributions/openhab/src/main/resources/bin/oh2_dir_layout#L6-L10).

Having the `OPENHAB_` prefix might be a bit annoying to always type in the docs & discussions, though - but it would definitely be the correct way of referencing, wdyt?

ThomDietrich:
I would second starting off with these. We just need to agree on one style.

An examples with all kinds of variations:

* `$OPENHAB_CONF/items/test.items`
* `$CONF/items/test.items`
* `$conf/items/test.items`
* `<conf>/items/test.items`
* `<openhab_conf>/items/test.items`
* `<CONF>/items/test.items`
* `<OPENHAB_CONF>/items/test.items`

Attention: angle brackets are interpreted as html tags outside `code fences`.
The dollar sign notation would actually be usable if the underlying system (e.g. openHABian) would be configured accordingly. That's a big plus.

kaikreuzer:
So you vote for `$OPENHAB_CONF/items/test.items`?
Would be fine for me.

ThomDietrich:
Yep :)
What do the others think?

ThomDietrich:
One addition, we should also define a default placeholder for the openHAB hosts IP address.

rtvb:
For the technically challenged users using forward slashes in the docs , but then replacing them with backslashes for windows needs to be addressed somewhere.

ThomDietrich:
[...]

PR #588, opened 2017-12-13
Update Linux and Console instructions for 2.2.0

### LINKED ISSUE #579: Console: Rephrase how to connect to the console
https://community.openhab.org/t/interactive-mode-how-where/36642

http://docs.openhab.org/administration/console.html#accessing-the-console

The beginning of the section should be clearer. We should clearly differentiate between the two options (by setup path), making the ssh-based one the default/preferred one.

BClark09:
- Removed manual backup instructions as these were out of date and do not work across versions.

Closes #579 
Closes openhab/openhab-distro#554

### DIFF
--- a/administration/console.md
+++ b/administration/console.md
@@ -20,7 +20,11 @@ The method to access the console depends on how openHAB was started.
 * When started in interactive mode using the provided command line scripts, openHAB naturally transitions directly to the console prompt.
 * When started as a service (e.g. when installed from our package repository), openHAB is running as a background process.
 
-In both cases, the console can be reached via secure shell connection ([SSH](https://en.wikipedia.org/wiki/Secure_Shell)).
+The simplest way of accessing the console is by running the `client` (or `client.bat` for windows) script located in `$OPENHAB_RUNTIME/bin/`. Linux package based installations can also use the command `openhab-cli console`.
+
+### Connecting via SSH
+
+The console can also be reached via secure shell connection ([SSH](https://en.wikipedia.org/wiki/Secure_Shell)).
 
 To reach the console using SSH, use the following command to connect to the localhost interface on TCP port 8101:
 
@@ -55,6 +59,22 @@ The first successful connection triggers generation of the Karaf remote console 
 *Note:* On slower systems, such as Raspberry Pi or Pine64, this first SSH connection may even time out.
 If this happens, simply try connecting again until successful.
 
+If openHAB is removed and reinstalled, then further attempts by SSH may bring the following error:
+
+```text
+@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@
+@ WARNING: REMOTE HOST IDENTIFICATION HAS CHANGED! @
+@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@
+```
+
+This is normal, since a new instance of openHAB will generate a new identification. The warning should also come with instruction on how to resume access to the console through SSH. e.g. by using the command:
+
+```shell
+ssh-keygen -f "/home/<username>/.ssh/known_hosts" -R [localhost]:8101
+```
+
+## Using the Console
+
 After successful connection and authentication, the console will appear:
 
 ```text
@@ -63,7 +83,7 @@ After successful connection and authentication, the console will appear:
  / __ \/ __ \/ _ \/ __ \/ /_/ / /| | / __  |
 / /_/ / /_/ /  __/ / / / __  / ___ |/ /_/ /
 \____/ .___/\___/_/ /_/_/ /_/_/  |_/_____/
-    /_/                        2.0.0
+    /_/                        2.2.0
                                Release Build
 
 Hit '<tab>' for a list of available commands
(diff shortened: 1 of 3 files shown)


ThomDietrich:
@BClark09 I think I was the one recommending SSH as the default option to access the console. It is platform independent and can (with the right setting) also be accessed from remote.
Anyhow, I'm not against your change. Just wanted to make the point.

ThomDietrich:
I'm btw not convinced we need the last addition. The error clearly states what is the issue and how the user can solve the problem.

I believe the user reporting the issue was looking for an openHAB based solution and I think it would be easy and correct to remove the key during openHAB removal. Imho we should consider this before adding redundant information in the documentation.

BClark09:
Thanks @ThomDietrich! When you say 'platform independent' you mean for Linux/Mac right? For now, Windows users will have to install a 3rd party program (probablly PuTTY) to connect that doesn't seem right. The `client` command is an SSH client that we (through Karaf) are providing. For reference, [Karaf also mentions the use of `client` first](https://svn.apache.org/repos/asf/karaf/site/production/manual/latest-3.0.x/start-stop.html) but I agree that specifically using `ssh` needs to be documented as a main point.

I didn't get any feedback from anyone about what to to about the known_hosts, but would very much appreciate your comments (and help) in that issue too. I'll remove it from this documentation.

administration/console.md
```diff
@@ -20,7 +20,13 @@ The method to access the console depends on how openHAB was started.
 * When started in interactive mode using the provided command line scripts, openHAB naturally transitions directly to the console prompt.
 * When started as a service (e.g. when installed from our package repository), openHAB is running as a background process.
 
-In both cases, the console can be reached via secure shell connection ([SSH](https://en.wikipedia.org/wiki/Secure_Shell)).
+The simplest way of accessing the console is by running the `client` (or `client.bat` for windows) script located in `$OPENHAB_RUNTIME/bin/`. Linux package based installations can also use the command `openhab-cli console`.
```
bgilmer77: Consider, "In this case, the simplest way of accessing...", and if there is a blank line between the sentence, "When started as a service..." and the line we are talking about, consider deleting it so this becomes one paragraph.

installation/linux.md
```diff
@@ -278,6 +281,25 @@ Systems based on **systemd** (e.g. Debian 8, Ubuntu 15.x, Raspbian Jessie and ne
 {% include collapsible/item-end.html %}
 {% include collapsible/end.html %}
 
+#### Command Line Interface (CLI)
+
+After installing openHAB, a shortcut named `openhab-cli` provides access to the openHAB specific commands (such as [backup, restore](#backup-and-restore), and [console]({{base}}/administration/console.html)).
```
bgilmer77: Consider, "provides access to openHAB-specific commands..."

installation/linux.md
```diff
@@ -278,6 +281,25 @@ Systems based on **systemd** (e.g. Debian 8, Ubuntu 15.x, Raspbian Jessie and ne
...
+To use the shortcuts in a terminal, simply type `openhab-cli` followed by the command. For example:
```
bgilmer77: Consider moving "For Example:" to its own separate line to avoid linter errors.

installation/linux.md
```diff
@@ -646,6 +578,31 @@ sudo rm /lib/systemd/system/openhab2.service
...
+
+If you're unsure how to use the above files, just use `--help` or `-h`:
+
+```shell
+$OPENHAB_RUNTIME/bin/backup --help
+```
+
 ## Viewing Log Messages
 
 In order to get more insight on what your openHAB system is doing and to see occurring error messages, it is recommended to always have a look on the openHAB log files.
```
bgilmer77: Consider changing to, "You can learn more about openHAB and how it works by looking at your log files."

bgilmer77:
@BClark09 +1 regarding SSH.  Isn't SSH already on the openHAB machine?  Isn't the user logging in to localhost?  (Yes, somehow you first have to connect to the openHAB machine...)

installation/linux.md
BClark09: 👍 Might as well when I'm touching the file!

BClark09:
Hi all, thanks for your comments! I hope that my recent changes address the problems?

@bgilmer77, not necessarily. Windows for example does not come with an openSSH client (yet).

bgilmer77:
@BClark09 I think we all are not on the same page.  Let's see if I can explain what I was getting at.
1.  If you want to connect to an OH machine, you need SSH on your local machine to do it.
2. If you are in *NIX land, then you are good - SSH most likely is already there.  If you are in MAC OSx land, SSH is already installed.  If you run Windows of course, you are correct - you need to install an SSH client.
3.  Once you have SSHed into OH from your local machine, THEN if you want to get at the console, you use SSH running on your OH machine to SSH to localhost.  This was my point.  Anyway, I think I am distracting from your work on this PR.  Thank you for the contribution.

ThomDietrich:
Everyone, great teamwork!
@BClark09 these are excellent changes, really like how both articles, especially the linux article follow a clear path now.

Two ideas for future PRs I want to note down:

1. We need to adapt the Linux, Mac and Windows article to the placeholders idea described in #189 - it's about time!
2. In openHABian it would be a good idea to offer an option to perform cron-based weekly backups, ideally to a mounted device/share.

PR #592, opened 2017-12-15
removed designer from the editor list and moved network preparations …

kaikreuzer:
…to the top as this is relevant for all editors

### DIFF
--- a/configuration/editors.md
+++ b/configuration/editors.md
@@ -17,10 +17,10 @@ This documentation page can give you some guidance in choosing the right one for
 
 ### Network Preparations
 
-The editors need to be able to access the configuration files on the remote openHAB host.
+Any editors used to configure openHAB need to be able to access the configuration files on the remote openHAB host.
 
-You have to have a [network share](https://en.wikipedia.org/wiki/Shared_resource) set up on the remote host and mounted on your local computer.
-The required steps to do so are specific to the host operation system.
+This can be achieved by using a [network share](https://en.wikipedia.org/wiki/Shared_resource) set up on the remote host and mounted on your local computer.
+The steps required to set up a [network share](https://en.wikipedia.org/wiki/Shared_resource) on your local host computer are specific to the host operation system.
 How to setup and use Samba on a Linux system is described in the [Linux article]({{base}}/installation/linux.html#network-sharing).
 If you are using [openHABian]({{base}}/installation/openhabian.html), the network shares are readily configured for you, you only need to mount them locally.
 
--- a/configuration/editors.md
+++ b/configuration/editors.md
@@ -15,6 +15,17 @@ This documentation page can give you some guidance in choosing the right one for
 - TOC
 {:toc}
 
+### Network Preparations
+
+The editors need to be able to access the configuration files on the remote openHAB host.
+
+You have to have a [network share](https://en.wikipedia.org/wiki/Shared_resource) set up on the remote host and mounted on your local computer.
+The required steps to do so are specific to the host operation system.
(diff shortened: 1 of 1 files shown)


configuration/editors.md
```diff
@@ -15,6 +15,17 @@ This documentation page can give you some guidance in choosing the right one for
 - TOC
 {:toc}
 
+### Network Preparations
+
+The editors need to be able to access the configuration files on the remote openHAB host.
```
bgilmer77: Consider, "Any editors used to configure openHAB need to be able to..."

configuration/editors.md
```diff
@@ -15,6 +15,17 @@ This documentation page can give you some guidance in choosing the right one for
...
+
+You have to have a [network share](https://en.wikipedia.org/wiki/Shared_resource) set up on the remote host and mounted on your local computer.
```
bgilmer77: Consider, "This can be achieved by using a `network share`...

configuration/editors.md
```diff
@@ -15,6 +15,17 @@ This documentation page can give you some guidance in choosing the right one for
...
+The required steps to do so are specific to the host operation system.
```
bgilmer77: Consider, "The steps required to set up a `network share` on your local host computer are specific to..."

kaikreuzer:
Thanks, @bgilmer77! I have addressed your comments.

ThomDietrich:
@kaikreuzer a bold but imho good step forward. Thanks!

configuration/editors.md
```diff
@@ -34,60 +45,10 @@ You can find it in the [Microsoft Visual Studio Marketplace](https://marketplace
...
-### Usage
-
-You can now start to create and modify your configuration files.
-Changes are automatically loaded by the openHAB runtime.
-More details about the syntax of the different configuration files can be found in the [Configuration]({{base}}/configuration/index.html) chapter.
-
-Hint: Press `<Ctrl + N>` to create a new file.
-
 {: #others}
 ## Other Editor Integrations
```
Nadahar: `like the two tools above` below doesn't make sense after Designer was removed.

Nadahar:
Is there any information somewhere about why Designer was removed? I just started with z-wave and opted to check out openHAB first since I'm very familiar with Java and Eclipse, which theoretically should make it relatively easy for me to contribute if I find something missing or wrong.

I was thus very disappointed when I discovered that VS Code was the only "supported IDE". I just can't stand VS in general and VS Code in particular, with it's "metro" UI and general lack of functionality. I took it for granted that I could use Eclipse for a project that is build on Java and is so tightly related to the Eclipse organization.

I assume that there's a reason for removing it from the documentation other than simply "taste". Has the designer been deprecated, abandoned or does it lack something fundamental? I can see that it hasn't been updated since December 2017, so I assume there's something here.

I've installed them both, and from the little I see at the level I am now the Designer is miles better to work with, both because it has a usable UI and because it has sensible "completion"/"intellisense". VS Code doesn't show me what valid options I have in a particular position, instead it seems to list all kind of stuff that is more or less relevant. Designer on the other hand does what I'd expect.

Can someone shed some light on this or link to where this information is available? Shouldn't the documentation also at least mention this, or is Designer "so useless" that it should not even be mentioned?

Confectrician:
Designer has been deprecated long ago and isn't continued anymore.
For detailed reasons i would ask you to check the openHAB community forum.
I think there should be some threads about this.

That's also the reason for removing designer from the documentation.
The code extension is the currently recommended editor solution.
(Which is clearly some steps away from being perfect, to be honest.)

PR #611, opened 2017-12-31
Fixing a bug in the Apache Karaf change password command

dougmcdorman:
This command will remove the comma which makes the new password unusable so it needs put a "," in the replacement string as well.
For comparison look at the password change function which has the "," in both the search and replace.
https://github.com/openhab/openhabian/blob/master/functions/passwords.sh
sed -i "s/openhab = .*,/openhab = $passwordChange,/g" /var/lib/openhab2/etc/users.properties

### DIFF
--- a/administration/console.md
+++ b/administration/console.md
@@ -150,7 +150,7 @@ Alternately, run the following Linux shell command, which will perform the repla
 Substitute `securePassword` with your desired password.
 
 ```shell
-sudo sed -i -e "s/openhab = .*,/openhab = securePassword/g" /var/lib/openhab2/etc/users.properties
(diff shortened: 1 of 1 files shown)


Confectrician:
Hi @dougmcdorman,

Thanks for helping us improving the openHAB documentation.
I will merge your imrpovement later.

May I point you to our [Contribution Guidelines](https://github.com/openhab/openhab-docs/blob/gh-pages/CONTRIBUTING.md), since you are a first-time-contributor here?
This doesn't mean that there's something with your current PR.
It isn't affected because it is a one line change.
Maybe you want to help us out in the future too (which i would appreciate much of course).
So it could be helpful to have at least read something about signing off the contributed work.

BR
Jerome

PR #612, opened 2017-12-31
How to install the openHAB Cloud add-on

dougmcdorman:
It wasn't obvious where to find this in the pile of add-ons so adding a one liner to tell people (beginners) where to find it.

### DIFF
--- a/_addons_ios/openhabcloud/readme.md
+++ b/_addons_ios/openhabcloud/readme.md
@@ -17,6 +17,8 @@ install: auto
 The openHAB Cloud Connector allows connecting the local openHAB runtime to a remote [openHAB Cloud](https://github.com/openhab/openhab-cloud/blob/master/README.md) instance, such as [myopenHAB.org](http://www.myopenHAB.org), which is an instance of the
 openHAB Cloud service hosted by the [openHAB Foundation](http://www.openhabfoundation.org/).
 
+To install this add-on, open Paper UI and navigate to _Add-Ons > Misc > openHAB Cloud Connector > Install_.
+
 ## Features
 
 The openHAB Cloud service (and thus the connector to it) is useful for different use cases:


Confectrician:
Hi @dougmcdorman and thanks for your contribution! 👋 

Many parts of the documentation are [imported from other openHAB repositories](https://github.com/openhab/openhab-docs#automatically-generated-parts), e.g., the Binding- or add-on specific repository. This way we ensure that the articles are always in sync with the actual source code. This method has proven to be quite helpful to keep documentation up-to-date.

You could have seen that from line 11 of the edited file:
> Attention authors: Do not edit directly. Please add your changes to the appropriate source repository

The correct location to edit would be the openhab2-addons repo.
https://github.com/openhab/openhab2-addons/blob/master/addons/io/org.openhab.io.openhabcloud/README.md

The file will then be grabbed by the next update run and docs will be updated with your change.

Thanks again for your willingness to contribute! 🙇
Please close this PR, after you have proposed the change in the correct repo.

dougmcdorman:
That is rather confusing, it would take me a long time to determine what the appropriate source repository would be.
I would suggest including something the the contributors guidance which I did read but did not see anything on the topic. I don't know how to describe it so I won't be attempting to make an update. The contributors guidance is also why I have only made a couple of one line changes.

Confectrician:
Hey @dougmcdorman,

Sad to hear that.
If you think, we could improve the contributors docs, feel free to file an issue with your doubts and maybe ideas.
It's already a help if you can tell us all at which point you maybe were confused.

Unfortanetely this won't be a one click contribution thing in future, even if we try to improve it continously.
I am helping out here for quite a while now and i sometimes have to take a look and search for some things again.
And i think we won't be able to avoid that completely.
This is the price we have to pay due to this big structure which is splitted into many repositories.

> it would take me a long time to determine what the appropriate source repository would be.

To make it as easy as possible, we maintainers look up the right place directly in case of such an pull request.
That's why i directly linked you the correct readme file above.

> The correct location to edit would be the openhab2-addons repo.
https://github.com/openhab/openhab2-addons/blob/master/addons/io/org.openhab.io.openhabcloud/README.md

Most of the grabbed parts are described here (as already linked above):
https://github.com/openhab/openhab-docs#automatically-generated-parts

We should add a complete list and maybe move this section a bit above for the future.
That's one (small) improvement, i would have in mind now on the first run.

Would be nice to have some input/issues from you.
If we can get a better impression of what is confusing, we are able to solve it maybe.

> The contributors guidance is also why I have only made a couple of one line changes.

Can you tell my what exactly is the limitating factor for you with the guidelines?
"Just one online changes" sounds like you simply don't want to add a sign-off statement.
Or is it something different?


BR
Jerome

PR #609, opened 2017-12-30
Updated gitignore for usage with vscode

Confectrician:
I am used to open the docs repo in vscode for small changes and for tests, 
since i am using it frequently for openHAB stuff anyways.
Maybe i am not the only one here, who is working this way, since vscode gets more and more popular in the world of openHAB.

So here's a small gitignore update before we have the vscode working folder here by mistake.

Signed-off-by: Jerome Luckenbach <github@luckenba.ch>

PR #614, opened 2018-01-02
Use -e USER_ID instead of --user for systemd, too

martinrunge:
Docker uses --user flag which causes container to not be able to create groups. This has already been fixed in the docs for the docker run command. 
This change just adds the same fix toe the systemd service file.

Signed-off-by: Martin Runge <martin.runge@web.de>

### DIFF
--- a/installation/docker.md
+++ b/installation/docker.md
@@ -146,7 +146,8 @@ ExecStart=/usr/bin/docker run --name=%n --net=host --tty \
   -v /opt/openhab/addons:/openhab/addons \
   -v /opt/openhab/.java:/openhab/.java \
   --device=/dev/ttyUSB0 \
-  --user=<uid> \
+  -e USER_ID=<uid_of_openhab> \
+  -e GROUP_ID=<gid_of_openhab> \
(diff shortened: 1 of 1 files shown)


Confectrician:
Hi and thanks for the quick reaction and PR. 😃 

One small thing is missing.
Altough this is a small change you need to sign off your work, because of our contribution guidelines.
Take a look at [this page](https://github.com/openhab/openhab-docs/blob/gh-pages/CONTRIBUTING.md) for further information. Especially [this part](https://github.com/openhab/openhab-docs/blob/gh-pages/CONTRIBUTING.md#sign-your-work) for the signing off stuff.

Usually a sign-of-by: is done in a commit, but for the docs repo and a first time contribution it is ok for us to do that in your pull request message.
So for this time it would be enough to add the following line with a edit in your first post.

`Signed-off-by: YOUR NAME <your@for-github-used.email>`

[Here](https://github.com/openhab/openhab-docs/pull/609) is a PR from me for comparison if the sign off line.

BR
Jerome

martinrunge:
done

PR #616, opened 2018-01-04
corrected install folder

6r15u:
according to the commands in the install script, the folders must only be "openHAB" instead of "openHAB2"


ThomDietrich:
Hello and thanks!

@thopiekar could you please verify?

thopiekar:
Yes, this should be correct!

thopiekar:
Hmm, approved the changes twice, but GitHub still complains about a missing review.

ThomDietrich:
That is a good question, I think you can't :) But that shouldn't be of your concern. Approval and the final merge are things done by the maintainers. One of those is me and I'll approve your addition now.
Btw. you can become part of the team if you want to ;)

Issue #623, opened 2018-01-10
Update Docker Upgrade Docs now that the container performs the upgrade steps for you


rkoshak:
The Docker scripts have been updated to automatically perform the upgrade steps when the container version is different from the userdata version.

PR #624, opened 2018-01-10
Updated the Docker Upgrade Instructions

rkoshak:
Addresses #623 

The Docker Image is now capable of performing the same upgrade steps as a standard apt-get/yum/upgrade script automatically when a container detects it is a different version from that in the mapped userdata folder.

I also added some warnings to look at the instructions on Docker Hub for the most up to date and a the complete set of arguments and supported environment variables.

Singed off by Rich Koshak: rlkoshak@gmail.com

### DIFF
--- a/installation/docker.md
+++ b/installation/docker.md
@@ -80,6 +80,7 @@ chown -R openhab:openhab /opt/openhab
 
 ### Running the Container as a Service Managed by Docker
 
+Note, always review the README on [Docker Hub](https://hub.docker.com/r/openhab/openhab/) for the most up to date set of recommended arguments and environment variables.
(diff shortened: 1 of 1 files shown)


bgilmer77:
Not using Docker, so can't review this, but @rkoshak, nice work!  Anything we can do to make openHAB behave "as expected" such as apt, is really fantastic.

Confectrician:
Hey Rich,

Thanks for your contribution.
Since no one of us maintainers uses docker as far as i know, i have requested a review from @cniweb.
I think in this case the approval is just a formal thing, but i don't want to break the workflow.

BR
Jerome

PR #633, opened 2018-01-22
Fix sentence fragement

chip-rosenthal:
There is a sentence fragment in the "Event Bus Actions" section which makes the text non-understandable.

I added what I believe is the missing text to complete the sentence. Can a subject matter expert review?


chip-rosenthal:
Thanks @Confectrician, I'll review that doc.

PR #667, opened 2018-04-18
config: items: Correcting label in table of types

thopiekar:
I think that way it is more clear that we are talking about the name of the type.
My first thought on this was, that this column is about the name of the item, but that one is free to choose of course.

Signed-off-by: Thomas Karl Pietrowski <thopiekar@gmail.com> (github: thopiekar)


thopiekar:
My pleasure! :wink:

Issue #665, opened 2018-04-10
[Discussion] Docs Repo restructuring


Confectrician:
Hi all,

One thing that i could observe repeatedly since i am helping out here in the docs
is the confusion about external content.
I have added the `external source` label some time ago for this and i am using it frequently.

# The Problem

People come here to improve the docs and me or others have to tell them that they are in the wrong place.
This is one thing if its *only* a new issue.
Its a bigger problem if someone already invested time, prepared an improvement and opened a PR.

This is especially hard for new contributors who have to get around

* with the docs itself,
* with jekyll 
(which has a really flat learning curve and big entry hurdle, to be honest)
* with our big/grown repo structure
* with a difficult system of generated files from `external sources`

The result is that some of them give up fast or even don't start thinking about contributing here.
And i can understand that sometimes to be honest.

First of all you will have to get jekyll running on your machine.
I first tried the vagrant machine, which had problem with refreshing after a file save.
(I think there was an issue about it.)
After being annoyed of that after some time i switched to the ruby variant.
It works but you have to do many steps until you can do what you originally wanted.
Work on those documentation.

Some time later, to me happened the same that happened to many others.

- I had made an improvement (i think it was about some transformation)
- I overread the "Please edit the original file"- Comment
- I had to move it to eclipse smarthome later

Fun Fact:
We discussed this change with @ThomDietrich before and even he overread that sentence in this case.

For me it was no big deal, because i already provided some commits to the contribution and was already a bit familiar with it.
But i can imagine that this can be frustrating for someone with no experience here at all.

# What is my Resumee now

We are making contributing for the docs way to hard especially for unexpierienced People.
I could imagine that docs would be a perfect place for people who are not familiar with programming,
but with a will to contribute to the Project in any way.

# How can we Improve this?

This is the Topic i want to discuss here.
But maybe i will go a step back first and would ask you for your opinion:

- Do you even agree with my observations?
- Do you also think the docs are confusing with all its external content and structure?
- Do you maybe see some additional problems?

## Maybe some ideas for starting with improvements

Now i have complained for many lines.
Time to give something back in form of a possible improvement

### A first suggestion

My suggestion would be to change the Repo structure a little bit.

1. All `final` contents of the docs page in a new repo which is filled only through generated contents
from the different sources. (Bindings, Docs-Articles, ESH, ...)
Nothing should be changed here by Pull Request.
2. A `pure-docs` repo where only Documentation Articles can be found.

[...]

PR #670, opened 2018-04-24
Fix wording of UDP binding

jpmens:
There was a three-year old note in the documentation which nobody had clarified. I am not a developer, but just stumbled over the documentation simply not matching what 2.3 does so my wording fixes at least that. 

However: a developer  urgently needs to address whether all of this is still true ...

### DIFF
--- a/_addons_bindings/tcp1/readme.md
+++ b/_addons_bindings/tcp1/readme.md
@@ -25,7 +25,7 @@ The TCP & UDP Bindings act as a network client or as a network server.
 
 The TCP and UDP bindings can be configured in the files `services/tcp.cfg` and `services/udp.cfg`, respectively.
 
-> Note: This is optional for the configuration and not necessary for receiving data. Item-defintions are enough for receiving data. (Developer confirm? 20150128). There's a bug in the binding that requires at least one udp configuration to be defined or the binding will not send UDP messages.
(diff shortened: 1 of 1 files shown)


Confectrician:
Hello and thanks for your contribution @jpmens! 👋 

Many parts of the documentation are [imported from other openHAB repositories](https://github.com/openhab/openhab-docs#automatically-generated-parts), e.g., the Binding- or add-on specific repository. This way we ensure that the articles are always in sync with the actual source code. This method has proven to be quite helpful to keep documentation up-to-date.
Long story short, you'll need to propose your change in the correct repository:
https://github.com/openhab/openhab1-addons

File: 
https://github.com/openhab/openhab1-addons/blob/master/bundles/binding/org.openhab.binding.tcp/README.md

Thanks again for your willingness to contribute! 🙇

The updated content will be in our documentation with the next external parts generation.

BR
Jerome

jpmens:
[Done](https://github.com/openhab/openhab1-addons/pull/5562)

Maybe remove the possibility to create PR/issues here then? Seems counter-intuitive to me. If you cannot, then perhaps add a PR/issue template which warns people when they do try.

Confectrician:
Thanks for your feedback.
Unfortunately that is not that easy.
There are several contents that have to be improved directly here.
I was/am not happy with this too and we are discussing a restructuring in #665 already.

So i am currently living with the situation too 
and we maybe are able to make some progress with that in the next weeks.

Thanks for moving the PR to the right repository.

BR
Jerome

PR #681, opened 2018-05-21
Prevent non-binding folders from being added to the list

### LINKED ISSUE #661: Bindings Page broken.
Somehow the build does somethign weird with the bindings page.
There are dozens of empty table rows and tbh i don't know from what they are generated.

The table rwos should rather be generated based on the addons_bindings folder,
but i can't see something wrong there.

cc: @kaikreuzer, @ThomDietrich

BClark09:
Fixes #661 

I'd admit I'm unfamiliar with Jekyll, but from the looks of things, the [binding list is generated](https://github.com/openhab/openhab-docs/blob/gh-pages/_config.yml#L18) using a Jekyll collection, which covers the whole directory tree, not just the single level of folders we want to use.

I couldn't find a way of limiting the depth created by the for loop, but we can make a simple check inside it to see if the binding in the collection has a relevant item.

Before:
image

After:
image

### DIFF
--- a/addons/bindings.md
+++ b/addons/bindings.md
@@ -84,6 +84,7 @@ Bindings connect your smart home's devices and technologies to openHAB.
   </thead>
   <tbody>
     {% for binding in site.addons_bindings %}
+        {% if binding.install %}
         <tr class="install-{{binding.install}} since-{{binding.since}}">
           <td>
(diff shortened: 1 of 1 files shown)


kaikreuzer:
@BClark09 Please do not spend any efforts on this! We will launch a fully new website on Friday and @ghys is not using Jekyll for that anymore.

BClark09:
Oops, my bad. I was not aware of the upcoming change.

PR #683, opened 2018-05-22
Remove space in group aggregation function example

ghys:


ghys:
The new website will have syntax highlighting for DSLs but it's not perfect. This is a workaround for one of its shortcomings in a prominent place :)

image

image

PR #688, opened 2018-05-26
added recommendations which tools to use for configuration

mstormi:
Signed-off-by: Markus Storm <markus.storm@gmx.net> (github: mstormi)

### DIFF
--- a/configuration/index.md
+++ b/configuration/index.md
@@ -59,13 +59,13 @@ The following table shows what an interface can be used for. For a recommendatio
     <tr>
       <td>Manage Things</td>
       <td>✔️</td>
-      <td>✔️(recommended)</td>
+      <td>✔️<br />(recommended)</td>
       <td>✔️</td>
       <td>✔️</td>
     </tr>
     <tr>
       <td>Manage Items</td>
-      <td>✔️(recommended)</td>
+      <td>✔️<br />(recommended)</td>
       <td>✔️</td>
       <td>✔️</td>
       <td>✔️</td>
--- a/configuration/index.md
+++ b/configuration/index.md
@@ -29,48 +29,13 @@ In order to represent all of these, openHAB defines the following few base compo
 The individual articles hold all details needed to understand the concepts behind these building bricks for your Smart Home.
 For more details on the base concept behind openHAB, please visit the [Concepts Overview page]({{base}}/concepts/index.html).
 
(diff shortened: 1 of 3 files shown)


ThomDietrich:
Was this change discussed somewhere? The index article is supposed to be a quick intro into configuration. Your addition definitely goes beyond that and is imho not intro material...

mstormi:
No I didn't feel that to be appropriate to discuss on the community so I thought to discuss it here.
It wasn't obvious that this article was supposed to be an index page. There's not many links to get deeper into specific topics while at the same time it feels somewhat incomprehensive.
The thing with all this information in there is that to have all those possibilities is pretty confusing for most users and newbies in particular. Well, reason is OH is that flexible but that's what many criticize.
So I felt it's a good idea to give some guidance. We can discuss the contents, I tried not to dismiss anybody's work but still outline the route to take.
Agreed that's not intro level any more, but I believe a user should encounter these hints _before_ digging into details of one tool or the other, so it should at least be prominently visible early in the docs browsing hierarchy.
Let me know where's a better place for this info. Maybe another page and a link to it on this index page ?
Switched parts for sake of readability. This wasn't quite the best-written article we have in OH docs.

Confectrician:
I feel a bit bad, because this pr is laying around her that long without any ongoing discussion.
Unfortunately i have not the time to give an appropiate answer, but just some short feedback:

I like the intention of improving this article, because i think @mstormi is right.
We have better written articles.

But i can understand the point of going beyond the scope too.
This should be an easy entry and when i look at the changed version i feel a bit slained from it.

Maybe we can find some solution between those opinions.
I will make some thoughts about a good place for a more detailed explanation,
which we could link to.

ThomDietrich:
I'm also not in the right moment to spend much time here currently. Sorry bout that.

As a starting point I'd suggest to re-swap the two sections. I believe that the table has to stay in the middle, i.e after the itemization and before the deeper discussion into "graphical vs text".

Next @mstormi I'd suggest to condense your additions so that they fit into the introductional character of the article.

After those two steps I'd be happy to go into a more detailed review.

@mstormi please do not see my criticism as discouragement... 😃

mstormi:
well it was a little discouraging.
But ok, how about this version ?

mstormi:
too old to merge easily, I opened PR#757 instead

PR #693, opened 2018-06-02
Update rules-dsl.md

raaahbin:
The use of a deprecated example in the documentation causes unnecessary confusion. This edit should at least provide users with a functional alternative.

Signed-off-by: Robin Darroch <git@robin.darroch.id.au> (github: raaahbin)

### DIFF
--- a/configuration/rules-dsl.md
+++ b/configuration/rules-dsl.md
@@ -430,6 +430,13 @@ calendar.timeInMillis = now.millis
 val dtt = new DateTimeType(calendar)
 ```
 
+Note: as DateTypeType(calendar) above is now deprecated, the simplest non-deprecated way to update a DateTimeType item with a Joda DateTime value in a rule is by converting it to a string.
+
+```java
+// update myDateTime item with current time
+myDateTime.postUpdate(now.toString)
+```
+
 In certain cases it is needed to convert an epoch timestamp to a human readable and/or store it in a DateTimeType and a DateTime Item. 
 Here an option to do so utilizing SimpleDateFormat:
 
--- a/configuration/rules-dsl.md
+++ b/configuration/rules-dsl.md
@@ -430,7 +430,7 @@ calendar.timeInMillis = now.millis
 val dtt = new DateTimeType(calendar)
 ```
 
-Note: as DateTypeType(calendar) above is now deprecated, the simplest non-deprecated way to update a DateTimeType item with a Joda DateTime value in a rule is by converting it to a string.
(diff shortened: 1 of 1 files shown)


configuration/rules-dsl.md
```diff
@@ -430,6 +430,13 @@ calendar.timeInMillis = now.millis
 val dtt = new DateTimeType(calendar)
 ```
 
+Note: as DateTypeType(calendar) above is now deprecated, the simplest non-deprecated way to update a DateTimeType item with a Joda DateTime value in a rule is by converting it to a string.
```
Confectrician: DateTimeType?

kaikreuzer:
> This seems to be the simplest way of dealing with the deprecated stuff, but is it also the "prefered one", wdyt?

I saw this part of the rule docs today for the first time and it really worries me - instead of elaborating more and more in the docs on how to circumvent the Java DateTime to Joda mismatch, we should rather fix that in ESH asap - I actually thought that Jodatime had been removed there already, but the `now` keyword seems to prove me wrong. I also just saw https://github.com/eclipse/smarthome/issues/5562, which is a similar issue. It seems with changing the DateTimeType recently, we (accidentially) broke the API, so we might want to fix it completely right away. (cc @sjka)

configuration/rules-dsl.md
raaahbin: Oh, true - it should just read "... a DateTime item ...". Now this n00b has to figure out how to edit the proposed edit.

raaahbin: And also the DateTypeType. Clearly I was not paying attention.

Confectrician:
Hey @raaahbin,

Thanks for the quick improvements.
Theoretically we would be ready to merge at this state.

But as you may hve read above, this is not simply a missing part of the docs, but more an issue with the rules engine itself.

So i will keep this PR open for now and watch the corresponding issues.
Maybe we can solve the problem on code base, which would be my preferred version too.

Thanks for clarification @kaikreuzer.

raaahbin:
I agree it would be much better to fix this at the code base level - my proposed edit is just to get the docs to match the functionality of the current version of the system (so that others are spared the frustration of copying code directly from the docs only to have it fail, leading to searches in the forums, to find that it's a known problem and to look there for the workaround). I'm certainly happy to defer to your experience though - if keeping this PR open is the best way to get the preferred resolution, please feel free to leave it open.

Confectrician:
It seems that eclipse/smarthome#5562 will lead to a code based solution.

I think we should close this PR then.
No need to write down a aworkaruond which then maybe gets forgotten and isn't removed.
Not the best situation for the confused users, but i think its the best compromise.

PR #695, opened 2018-06-03
Update ZWave documentation

cdjackson:
Updated to fix incorrect information from the database documentation generator, and add a number of new things.
Signed-off-by: Chris Jackson <chris@cd-jackson.com>


Confectrician:
Isn't zwave grabbed by the update script @kaikreuzer?
Or did we change somethign due to the last problems?

cdjackson:
Kai said that a week or so back that he had to copy the files manually, so to try and save him some time, I provided this directly. If there is now a script, then that’s great.

https://github.com/openhab/org.openhab.binding.zwave/pull/915#issuecomment-392346196 <https://github.com/openhab/org.openhab.binding.zwave/pull/915#issuecomment-392346196>

Confectrician:
There is an update [bash script](https://github.com/openhab/openhab-docs/blob/gh-pages/update-external-resources.sh) existing for a long time, but it has to be executed manually (i think that is what Kai meant).
Kai provided a last update in https://github.com/openhab/openhab-docs/commit/0949e9f5d62a5140ff0c960b6326951b88730742 7 days ago at the time of your conversation.

Is that the current state of docs or have you updated something in the last 7 days in the zwave repo?

cdjackson:
Ok, sorry, I'll close this if there's a script to do it, but please run the script so we can have the updated docs.

Thanks.

kaikreuzer:
> but please run the script

I am currently hesitant to run it, because the current state of the documentation is the 2.3 release - and afaik, we do not yet have specific docs deployed anywhere. @ghys Do you already have a clue how we should proceed with that? Can you freeze the current docs and make them available as "2.3" in vuepress? I really would want to avoid that we have to put them on the old docs website instead...

cdjackson:
Understood - do you mind instead if we merge this PR to get the ZWave docs updated? This fixes some significant issues with the docs so it would be good to have it merged (ie the docs currently in 2.3 have some big errors due to the doc generator and that is fixed here).

kaikreuzer:
Ok with that - it just seems that the branch has already been deleted, so I cannot re-open and merge it.

cdjackson:
Thanks - I've repushed it...

ghys:
> @ghys Do you already have a clue how we should proceed with that? Can you freeze the current docs and make them available as "2.3" in vuepress?

That could be done - if there's a tag or branch for the 2.3 version, I can tell my script to clone that tag or branch, effectively "freezing" the docs. But it leads to another issue then: the latest 2.4 snapshot's docs won't be available anywhere...
It's not an easy problem as you know and heavily linked to #665.

What I'd like to suggest and eventually achieve if possible is to have in the frontmatter of Markdown articles (be it things or new bindings) an indication of minimum application version and ideally build number or date.
Something like:
```
---
since_version: 2.4
since_date: 2018-06-03T12:00:00Z (1st option)
since_build: 1294 (2nd option)
---
```
_(since_date is a good start, since_build is probably a little harder)_

With that info I could easily put a warning banner on the 2.4 pages (something like _"this is primilarily documentation, you need a 2.4.0-SNAPSHOT distribution build dated 3 June 2018/#1294 or later"_) which would automatically disappear when 2.4 is out.
Does that make sense? @cdjackson is this something you could make happen for the ZWave things pages?

cdjackson:
I don’t currently have a “first release” time in the ZWave database. I could (maybe!) create one by somehow processing the git repo to find the date the xml file was first added - I’m not sure it’s 100% reliable (due to for example renaming which might happen occasionally), but it would probably work for 99%…  Once I had such an entry, it would be easy to provide this.

I will add the first release date to the database though - then at least it’s there for new devices.

cdjackson:
Just to follow up on this, I’ve managed to write a short script to extract the date each XML file was added to git, so this information will be added to the database. This will allow me to support the “since_date” option if you want this added at some stage.

kaikreuzer:
> Does that make sense?

Actually, not too much I am afraid. We have many updates of existing pages during the course of a release - new channels being added to a binding, changes in the configuration options, etc.
We therefore definitely need a frozen version of every page - this could be taken from the release tags of the openhab-docs repo.
@ghys Those pages do not need to be included in the search or anywhere in the navigation - they should just be available using the same page style as the snapshot version.

Figuring out that a page has been changed or newly added since the last release could actually be done automatically by some script as well - I don't think we really require manually maintained frontmatter, do we?

PR #698, opened 2018-06-05
Clarify Contact Items

DomQuixote:
Clarification is needed that a contact item will not accept any commands (for example, MQTT binding with COMMAND does not update item and gives a 'command is NULL' error, that is in fact also confusing). 

The Contact item 'CommandType' should be renamed OpenClosed with a D (and not OpenClose)  as the status is in fact CLOSED and not CLOSE. 

Furthermore a switch is not only used for lights, in fact, lights are nowadays more often controlled by Color items so it's not the best example, best to not be too specific and just mention ON and OFF.

Signed-off-by: Bas Stronks bas.stronks@gmaill.com (github: DomQuixote)

### DIFF
--- a/concepts/items.md
+++ b/concepts/items.md
@@ -16,7 +16,7 @@ The following item types are currently available (alphabetical order):
 | Item Name          | Description | Command Types |
 |--------------------|-------------|---------------|
 | Color              | Color information (RGB) | OnOff, IncreaseDecrease, Percent, HSB |
-| Contact            | Item storing status of e.g. door/window contacts. Does not accept commands. | OpenClosed |
+| Contact            | Item storing status of e.g. door/window contacts | OpenClosed |
 | DateTime           | Stores date and time | - |
 | Dimmer             | Item carrying a percentage value for dimmers | OnOff, IncreaseDecrease, Percent |
 | Group              | Item to nest other items / collect them in groups | - |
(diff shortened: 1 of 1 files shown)


Confectrician:
Hi @DomQuixote,

Thanks for you will to improve the docs.
The concepts articles are maintained in the corresponding repo beneath the code.
We grab it from time to time.

For concepts/items.md the file would be:
https://github.com/eclipse/smarthome/blob/master/docs/documentation/concepts/items.md

## What does that mean now?

Unfortunately we can only handle the [configuration/items.md] improvement here.
The [concept/items.md] improvement has to be made in the eclipse/smarthome repo.

Also you would have to `sign-off` your change according to our [contribution guidelines](https://github.com/openhab/openhab-docs/blob/gh-pages/CONTRIBUTING.md#sign-your-work).

For first time contributions in the docs its ok to edit the Pull Request description and add a line like this one:

> Signed-off-by: Joe Smith <joe.smith@email.com> (github: github_handle)


My suggestion would be, to discuss the generic improvement here and just copy over the concepts files to a new pr in smarthom repo after finishing the discussion.

DomQuixote:
Like this?

Confectrician:
>Like this?
Correct. 👍 

Another hint that may help in the future:
The recommended way for sign off, is to do this for every commit.
If you use git frequently and it is set up correct on your environment:
There's even a command parameter that does the work for you.
_git commit **-s** -m "Commitmessage"_ will append the sign off statement after your commitmessage.
(We should mention that in our contribution stuff also at some time.)

## About the PR

I would suggest you to remove the **concepts/items.md** edits.
They would be overwritten on the next run of our external content script.
Afterwards we could merge these changes and propose a PR to the eclipse/smarthome repo.
(I think there is an eclipse account needed for that, so it would be fine for me to open the PR myself if you don't want to create another account. I already gave one.)

I would also suggest to propose onle the changes you made to the **configuration/items.md** file
to keep those tables equivalent.


If you have any questions, just drop them here.

Confectrician:
Hey @DomQuixote a friendly ping.

I would like to merge the configuration part, since its an improvement of course. 🙂 

As stated above concepts, has to be changed in the smarthome repo.
We can do that afterwards and it will be in one of the next generated website builds.

BR
Jerome

PR #705, opened 2018-06-07
Fix broken link to Beginner tutorial

eikowagenknecht:
Previously led to 404 error.

### DIFF
--- a/installation/index.md
+++ b/installation/index.md
@@ -93,7 +93,7 @@ After you got openHAB 2 set up and running, there are a few additional setup ste
 ## Getting started
 
 With the openHAB 2 distribution up and running, you should now continue with
-the [Beginner Tutorial]({{base}}/tutorials/beginner)
+the [Beginner Tutorial]({{base}}/tutorial)
 or by working on your own configuration.
 
 ## Help


Confectrician:
Hey @ghys & @kaikreuzer,

This fixes the link in the "new Website" but will break docs.openhab.org for this link.
Maybe time for a general decision.

I would focus on the new website and accept some broken links since docs.openhab.org could be called deprecated.
What do you both think how to handle it?

kaikreuzer:
Yes, I agree with that.
Anyhow, @ghys suggested to sooner or later do an automatic redirect to the new docs, so broken links won't be noticed by anyone anymore then.

Confectrician:
>Yes, I agree with that.

Then LGTM

>Anyhow, @ghys suggested to sooner or later do an automatic redirect to the new docs, so broken links won't be noticed by anyone anymore then.

Even better, so we won't have to deal that long time with this.

ghys:
Note that was the target of one of many "rules" in the website's prepare-docs.rb supposed to temporarily fix broken links:

https://github.com/openhab/website/blob/a2b9ac046174a03c8e3105c776980e14f5c74c99/prepare-docs.rb#L240

but it missed this one because of the trailing slash. This was at the time where we didn't know what would happen of the docs on docs.openhab.org. Fixing them at the source is surely better (as I said in my blog post those on-the-fly replacements could be removed from the script once they're fixed everywhere they appear in the source docs, but I don't see that happening any time soon...)

PR #714, opened 2018-06-11
[administration/jsondb.md]Changed images from imageshack to local saved ones.

Confectrician:
Signed-off-by: Jerome Luckenbach <github@luckenba.ch>

PR #715, opened 2018-06-11
[configuration/addons.md] Initial commit (Moved from deprecated addons page)

### LINKED ISSUE #702: Move addons installation instruction to configuration section
Needed due to the website refactoring.
Reference: #701 

Idea:
Move to new `configuration/addons.md` file

Confectrician:
Moved addons installation article to a new page and edited headline hierarchy.

Closes #702


configuration/addons.md
```diff
@@ -0,0 +1,98 @@
+---
+layout: documentation
+---
+
+{% include base.html %}
+
+# Installation of Add-ons
+
+Depending on the [package]({{base}}/configuration/packages.html) you have choosen during your first time setup, there are already some pre-installed add-ons.
```
ghys: `{{base}}` doesn't exist in VuePress (the site base is in `$site.base` but it's unnecessary to add it, it's handled by the framework), so this would normally fail, however to maintain compatibility with Jenkins it's transformed into `/docs/` by the prepare-docs.rb build script.

So we could leave this and let the prepare-docs script replace it, but I believe we decided to drop that Jekyll compatibility altogether, especially since it's a new page?
In that case it should be replaced everywhere it appears:
`{{base}}/configuration/packages.html` -> `/docs/configuration/packages.html`

configuration/addons.md
```diff
@@ -0,0 +1,98 @@
...
+Additional add-ons can be installed in the different ways, described below.
+
+## Through Paper UI
+
+Navigate to the add-ons section.
+Search for the desired add-on in the categories and press install.
+
+![installing Add-ons through Paper UI](images/addons_index_installation_paperui.gif)
```
ghys: Can you change this to `./images/addons_index_installation_paperui.gif`? (see my remark in #714)

configuration/addons.md
Confectrician: Thanks for clarifying this!
We should start a collection in team discussions, with things that change while we are moving to vuepress.

Confectrician:
All done.

ghys:
Looks good!

> So for now the new page would have to be added there, but as you can see for add-ons JSON files are imported and inserted into the navigation, I wanted to go that route for the whole /docs sidebar too - so the navigation would be maintained in this repo and imported during the website's build, I think it makes better sense.

Linking to #717

PR #723, opened 2018-06-17
Added 'everyMinute' strategy to persistence examples

cweitkamp:
We talk about it [in the sitemap section](https://www.openhab.org/docs/configuration/sitemaps.html#element-type-chart) and [in the page for rrd4j](https://www.openhab.org/addons/persistence/rrd4j/#example) but never gave an explanation how it looks like. This will make it easier for new user.

### DIFF
--- a/configuration/persistence.md
+++ b/configuration/persistence.md
@@ -112,8 +112,9 @@ Below you will find a complete example persistence configuration file:
 ```java
 // persistence strategies have a name and definition and are referred to in the "Items" section
 Strategies {
-        everyHour : "0 0 * * * ?"
-        everyDay  : "0 0 0 * * ?"
+        everyMinute : "0 * * * * ?"
+        everyHour   : "0 0 * * * ?"
+        everyDay    : "0 0 0 * * ?"
 
         // if no strategy is specified for an Item entry below, the default list will be used
        default = everyChange
--- a/configuration/sitemaps.md
+++ b/configuration/sitemaps.md
@@ -454,7 +454,7 @@ See this [Tutorial](https://community.openhab.org/t/13761/1) for more details.
 
 **Technical constraints and details:**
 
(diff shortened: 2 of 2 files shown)


Confectrician:
Perfect thanks!

This will be one candidate for "fastet pr merge in OH so far" 😅

cweitkamp:
> This will be one candidate for "fastet pr merge in OH so far" 

:+1:

ThomDietrich:
I don't like the incentive this PR provides.
I'd argue that in 99% of all cases where an enduser uses `everyMinute` they had other intentions. Use of this strategy will needlessly bombard the persistence service with data. Instead of promoting the `everyMinute` strategy we should clarify the importance of `everyChange` and `everyUpdate`.

Confectrician:
> I just realized that for rrd4j one needs to use everyMinute. Is that really the case? One more reason to use a better service...

I didn't check, but trusted in this case that this is needed.
I agrre, but if it is that way, we should provide an example as made in this pr.

>Instead of promoting the everyMinute strategy we should clarify the importance of everyChange and everyUpdate.

That's a thing we could point out as important anyway , since this seeems to have more senseful usecases.

cweitkamp:
> Is that really the case?

Yes, it is.

An idea: We can move the explanation for the `everyMinute` strategy to the rrd4j documentation instead and remove it from the persistence and maybe the sitemap documentation? wdyt?

Confectrician:
Would be fine for me.
When its especially needed for rrd4j, it should definetely be told there.

And if one really need everyMinute somewhere else, he/she will be able to find the needed cron expression, since i would say this is a more advanced usecase. :)

ThomDietrich:
Sounds good to me!

cweitkamp:
Have a look at openhab/openhab1-addons#5602 and #724

Issue #711, opened 2018-06-09
[installation] Restructure the whole area and adapt it to the download page?!


Confectrician:
I have already seen some "my download was wrong, but i used the links from the docs" posts in the community.

We should from now on only point to our download page.
If something changes, we only have to fix it there and all doc pages will be fine without the need to touch them again.

Confectrician:
i was nearly already starting with fixing some urls when i struggled a bit.

Since we have got "short installation instructions" already on the download page
it maybe would make sence to rework the installation section in complete, to adapt it to the new website structure.

I would like to have some feedback first.

MyRaceData:
I think we have two issues here. 
1. dead or incorrect links to downloads being mentioned in the documentation. I agree all links for downloading openHAB in the documentation should point to the download page only so bad links don't ruin the experience for a new user attempting to install the program and easier maintaining of the documentation. If I find any such links, I will create PR to fix
2. What you refer to as "short installation instructions" which are included on the download page are for Linux installations and may confuse users of other systems. I feel this section should be relocated in the Linux installation section and leave the download page just for download links and download instructions

Confectrician:
>None the less, I'm still not convinced the download page is a good place for installation instuctions

Hey @ghys do you have an ide how we could deal with that?
@MyRaceData has a point here and we could simplify the website with providing only download relted information.

Maybe point to the corresponding docs section for each download type?

ghys:
Sorry for missing this. I believe a lot of people just click on the Download page and want to get started quickly without skimming through the docs (although they probably will have to, hence the warning). That's why the download page has _simplified_ instructions.

PR #725, opened 2018-06-17
Fix Windows download links and files to delete during manual upgrade

wborn:
The Windows instructions are incomplete regarding the files to delete during a manual upgrade (see [community topic](https://community.openhab.org/t/fail-in-update-to-2-3-0-windows/46552?u=wborn)). This results in an installation that won't start.

* Updated and sorted list of manually to be deleted files
* Updated download links and badges for openHAB stable and snapshot versions

### DIFF
--- a/installation/windows.md
+++ b/installation/windows.md
@@ -37,7 +37,7 @@ On this page, we will be using `C:\openHAB2` as an example.
 The openHAB runtime is distributed using a platform-independent zip file.
 To install it, follow these simple steps:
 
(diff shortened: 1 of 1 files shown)


installation/windows.md
```diff
@@ -37,7 +37,7 @@ On this page, we will be using `C:\openHAB2` as an example.
 The openHAB runtime is distributed using a platform-independent zip file.
 To install it, follow these simple steps:
 
-1.  Choose between the Stable Version [![Download](https://api.bintray.com/packages/openhab/mvn/openhab-distro/images/download.svg)](https://bintray.com/openhab/mvn/download_file?file_path=org%2Fopenhab%2Fdistro%2Fopenhab%2F2.2.0%2Fopenhab-2.2.0.zip) or the latest Snapshot Version [![Download](https://api.bintray.com/packages/openhab/mvn/openhab-distro/images/download.svg?version=2.3.0)](https://openhab.ci.cloudbees.com/job/openHAB-Distribution/lastSuccessfulBuild/artifact/distributions/openhab/target/openhab-2.3.0-SNAPSHOT.zip) of openHAB.
```
Confectrician: I would remove the cloudbees links completely and point to `https://www.openhab.org/download/` instead.
That is what i wanted to do anyway for all installation articles in the future.
See #711 for reference.

It simply doesn't make sense for me, to fix those links after every release on every article.

wborn: I was thinking about that solution as well. I'll remove them!

Confectrician: I have also directly opened openhab/website#78 too get this a bit smoother in the future, if possible.

wborn: Excellent! That would indeed make it a bit easier and less confusing. :smile:

PR #737, opened 2018-07-09
Update linux.md

a-tibi:
It mentioned: 
{Be sure to use the actual host name instead of openhab-device. When asked, authenticate with the username "openhab" and the chosen password.}

But for me it works for when I replace 'openhab-device.local' (including local) with 'openhabianpi',
and the username/password: "openhabian"

### DIFF
--- a/installation/linux.md
+++ b/installation/linux.md
@@ -775,8 +775,8 @@ You can also manually connect:
 - **On Mac OS X:** Open Finder → Go → Connect to Server: `smb://openhab@openhab-device.local`
 - **On Windows:** Open Windows Explorer → Address bar: `\\openhab-device.local` → Right click a share and assign a drive letter
 
-Be sure to use the actual host name instead of `openhab-device`.
-When asked, authenticate with the username "openhab" and the chosen password.
+Be sure to use the actual host name instead of `openhab-device.local`, the default host name is `openhabianpi`.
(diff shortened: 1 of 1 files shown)


installation/linux.md
```diff
@@ -775,8 +775,8 @@ You can also manually connect:
 - **On Mac OS X:** Open Finder → Go → Connect to Server: `smb://openhab@openhab-device.local`
 - **On Windows:** Open Windows Explorer → Address bar: `\\openhab-device.local` → Right click a share and assign a drive letter
 
-Be sure to use the actual host name instead of `openhab-device`.
-When asked, authenticate with the username "openhab" and the chosen password.
+Be sure to use the actual host name instead of `openhab-device.local`, the default host name is `openhabianpi`.
```
Confectrician: The `local` depends on your network infrastructure i think.
For me it would be `openhab-device.fritz.box`, so i dont think it is a good idea to add a specific domain ending over there.

Confectrician:
So only the `.local` would be correct and the username/password would have to be left as `openhab` correct?

ThomDietrich:
Well it's a gray area :) the hostname in the example is "openhab-device". The additional "local" is a local domain name.

https://en.wikipedia.org/wiki/.local

As for the username: I'd say we should stay with "openhab" but make clear that it is just an example user name and the actual name depends on your system and config

PR #738, opened 2018-07-09
items.md: override labels

Josar:
Added information about how to override label text and state representation on sitemaps.

https://github.com/eclipse/smarthome/issues/5867#issuecomment-403367270

### DIFF
--- a/configuration/items.md
+++ b/configuration/items.md
@@ -202,15 +202,32 @@ Two naming schemes are established in the community for Group names:
 {: #label}
 ### Label
 
-Label text is used to describe an Item in a human-readable way.
-Graphical UIs will display the label text when the Item is included, e.g. in [Basic UI]({{base}}/addons/uis/basic/readme.html) in a [Sitemap]({{base}}/configuration/sitemaps.html) definition.
+The label is used to describe an Item in a human-readable way.
+Graphical UIs will display the label when the Item is included, e.g. in [Basic UI]({{base}}/addons/uis/basic/readme.html) in a [Sitemap]({{base}}/configuration/sitemaps.html) definition.
 Some I/O services (e.g. the Amazon Alexa skill) also use the label to match an external voice command to an Item.
 
-In textual configurations the label, in quotation marks, appears next to the optional state presentation field in square brackets (see below).
-The label for the Item in the following example is "Temperature":
+In textual configurations the label in quotation marks, contains the label text and the optional state presentation field in square brackets (see below).
+They will be showed side by side, the text left aligned and the state representation right-aligned.
+The label text for the Item in the following example is "Temperature" and the optional state representation is set to be displayed e.g as "23.9 °C":
 
 ```java
 Number Livingroom_Temperature "Temperature [%.1f °C]"
+```
+
+The label defined for an item can be overwritten by the label definition on a sitemap with the label atribute.
+If the label definition on the item contained label text and state representations these have to be considered seperatly.
+
+```perl
+sitemap demo label="My home automation" {
+    Frame label="Date" {
+        // Overrides label text only, keeps state representation defined by Item
+        Text item=Livingroom_Temperature label="Livingroom"
+        // Overrides label text and sets state representation to be empty, not displayed
+        Text item=Livingroom_Temperature label="Livingroom []"
+        // Overrides label text and state representation
+        Text item=Livingroom_Temperature label="Livingroom [%.2f °F]"
+    }
+}
 ```
 
 {: #state}
--- a/configuration/items.md
+++ b/configuration/items.md
@@ -214,8 +214,8 @@ The label text for the Item in the following example is "Temperature" and the op
 Number Livingroom_Temperature "Temperature [%.1f °C]"
 ```
 
-The label defined for an item can be overwritten by the label definition on a sitemap with the label atribute.
-If the label definition on the item contained label text and state representations these have to be considered seperatly.
+The label defined for an Item can be overwritten by the label definition on a sitemap.
+If the label definition on the Item contained label text and the state representation these have to be considered separatly.
 
 ```perl
 sitemap demo label="My home automation" {
--- a/configuration/items.md
+++ b/configuration/items.md
@@ -202,33 +202,18 @@ Two naming schemes are established in the community for Group names:
 {: #label}
 ### Label
 
-The label is used to describe an Item in a human-readable way.
-Graphical UIs will display the label when the Item is included, e.g. in [Basic UI]({{base}}/addons/uis/basic/readme.html) in a [Sitemap]({{base}}/configuration/sitemaps.html) definition.
+Label text is used to describe an Item in a human-readable way.
+Graphical UIs will display the label text when the Item is included, e.g. in [Basic UI]({{base}}/addons/uis/basic/readme.html) in a [Sitemap]({{base}}/configuration/sitemaps.html) definition.
 Some I/O services (e.g. the Amazon Alexa skill) also use the label to match an external voice command to an Item.
 
-In textual configurations the label in quotation marks, contains the label text and the optional state presentation field in square brackets (see below).
-They will be showed side by side, the text left aligned and the state representation right-aligned.
-The label text for the Item in the following example is "Temperature" and the optional state representation is set to be displayed e.g as "23.9 °C":
+In textual configurations the label, in quotation marks, appears next to the optional state presentation field in square brackets (see below).
+The label for the Item in the following example is "Temperature and the optional state representation is set to be displayed, e.g. as "23.9 °C":
 
 ```java
 Number Livingroom_Temperature "Temperature [%.1f °C]"
 ```
 
-The label defined for an Item can be overwritten by the label definition on a sitemap.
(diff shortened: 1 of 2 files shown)


configuration/items.md
```diff
@@ -202,17 +202,34 @@ Two naming schemes are established in the community for Group names:
...
+In textual configurations the label in quotation marks, contains the label text and the optional state presentation field in square brackets (see below).
+They will be showed side by side, the text left aligned and the state representation right-aligned.
+The label text for the Item in the following example is "Temperature" and the optional state representation is set to be displayed e.g as "23.9 °C":
 
 ```java
 Number Livingroom_Temperature "Temperature [%.1f °C]"
 ```
 
+The label defined for an Item can be overwritten by the label definition on a sitemap.
+If the label definition on the Item contained label text and the state representation these have to be considered separatly.
```
Confectrician: This was really hard to read for me.
Did i get it correct?

1. The item definition is made with some text and a string formatter for the item state
2. We then have to take care of state and "text-label" differently in the sitemap

If yes i would suggest something like:

>If the label defined in an Item contains state and text, these representations have to be considered separately in the sitemap.

configuration/items.md
```diff
@@ -202,17 +202,34 @@ Two naming schemes are established in the community for Group names:
...
+
+```perl
+sitemap demo label="My home automation" {
+    Frame label="Date" {
+        // Overrides label text only, keeps state representation defined by Item
```
Confectrician: Suggestion:
Overrides only the text, but will keep the state format from the Item definition

configuration/items.md
```diff
@@ -202,17 +202,34 @@ Two naming schemes are established in the community for Group names:
...
+        Text item=Livingroom_Temperature label="Livingroom"
+        // Overrides label text and sets state representation to be empty, not displayed
```
Confectrician: Suggestion:
Overrides the text and hides any state representation.

configuration/items.md
```diff
@@ -202,17 +202,34 @@ Two naming schemes are established in the community for Group names:
...
-Label text is used to describe an Item in a human-readable way.
-Graphical UIs will display the label text when the Item is included, e.g. in [Basic UI]({{base}}/addons/uis/basic/readme.html) in a [Sitemap]({{base}}/configuration/sitemaps.html) definition.
+The label is used to describe an Item in a human-readable way.
+Graphical UIs will display the label when the Item is included, e.g. in [Basic UI]({{base}}/addons/uis/basic/readme.html) in a [Sitemap]({{base}}/configuration/sitemaps.html) definition.
 Some I/O services (e.g. the Amazon Alexa skill) also use the label to match an external voice command to an Item.
 
-In textual configurations the label, in quotation marks, appears next to the optional state presentation field in square brackets (see below).
-The label for the Item in the following example is "Temperature":
+In textual configurations the label in quotation marks, contains the label text and the optional state presentation field in square brackets (see below).
+They will be showed side by side, the text left aligned and the state representation right-aligned.
```
ThomDietrich: Depends on UI. I'd not claim this behavior here.

configuration/items.md
```diff
@@ -202,17 +202,34 @@ Two naming schemes are established in the community for Group names:
...
+The label text for the Item in the following example is "Temperature" and the optional state representation is set to be displayed e.g as "23.9 °C":
```
ThomDietrich: ", e.g.,"

configuration/items.md
ThomDietrich: Please capitalize "Sitemap"

ThomDietrich: This is an important piece of information. I wonder if we did not yet describe this somewhere.

I also feel the explanation is a bit complicated. You can define label text and the state representation in Items and in Sitemaps. Redefiniton in Sitemaps overwrites item definition.

I like to work with absolute names in Item labels and shorter specialized versions in sitemaps.

Confectrician: Changed my comment.
A whole pr with capitalization in smarthome open and still doing the same typos...

ThomDietrich: Just for info: https://www.openhab.org/docs/configuration/sitemaps.html#element-types

> When an Item is defined, you have the opportunity to assign a label and/or an icon at that point. If no label or icon are specified in the Sitemap, then the label and/or icon you assigned to the Item will be displayed. Setting a value for label or icon of a Sitemap element will override the values defined for the linked Item.

configuration/items.md
ThomDietrich: We have a problem here. Sitemaps are technically not Perl syntax but look best with it. However `//` does not introduce comments in Perl. That is why you need to write: `//# Comment`

ThomDietrich:
Hey guys, I like the addition! It is an important topic and your examples cover everything! Good work 👏 
However I'm uncertain if it should go into the Items article. Overwrite capability in Sitemaps clearly sounds like a Sitemaps article topic.

configuration/items.md
ThomDietrich: Sadly this is not the case. There are at least cases in which a state will be shown. You need to explicitly define `[]` if you want an empty representation.

This overwrite is good for just defining a custom label text in sitemap

Confectrician: Isn't this the commen for line 226 whre the `[]` are explicitly defined?
Or am getting your comment wrong now?!

Confectrician:
You are right @ThomDietrich,

To be honest, i have not read clear enough in the referenced smarthome discussion.
@htreu asked also for an improvement of the sitemap docs over there.

configuration/items.md
ThomDietrich:
```
String Whg_Light_Preset "Wohnung Beleuchtung [%s]"
Switch item=Whg_Light_Preset icon="light" label="Test"
```

grafik

configuration/items.md
```diff
@@ -202,17 +202,34 @@ Two naming schemes are established in the community for Group names:
...
+
+```perl
+sitemap demo label="My home automation" {
+    Frame label="Date" {
+        // Overrides label text only, keeps state representation defined by Item
+        Text item=Livingroom_Temperature label="Livingroom"
+        // Overrides label text and sets state representation to be empty, not displayed
+        Text item=Livingroom_Temperature label="Livingroom []"
+        // Overrides label text and state representation
+        Text item=Livingroom_Temperature label="Livingroom [%.2f °F]"
```
ThomDietrich: I think the example is confusing. A user might expect automatic conversion...

configuration/items.md
ThomDietrich: You are totally right!!
Sorry @Josar, please ignore my critique here. You did an excellent job finding all cases 😅

ThomDietrich:
@Confectrician I missed the ESH issue. Now I see where this is coming from. @Josar it's great that you followed though! I'd suggest to add a new subsection to the Sitemaps article on Item overwrite capabilities. Here in the Items article I'd add a link to that new section and everyone is happy :)

configuration/items.md
Josar: And i think he would get it if the temperature has an unit as defined for the UoM 

https://community.openhab.org/t/list-of-units-in-the-uom-unit-of-measurement/46842

Josar:
I am somehow allready confused be the spread information. 
On the sitemaps doc, there is a list which lists Common parameters and links to the section for labels at the items page. 
Then in the next line it addresses the overwriting of item parameter. 

This makes it for me as a reader or someone who is searching for information hard to find all this.

I think a better aproach would be to bundle information depending on e.g. labels in one place and list all depending things there. As the sitmaps doc was allready to confusing i decided that the label section at the items page would be the best place to put this information. I Don't think a new label section on the sitemaps page increases the accessability to this information it  would spread it further.

Confectrician:
I disagree slightly.
Overriding can only be done in sitemaps so this is the place to be.

Unrelated to that i would agree in the fact that the sitemaps article is really huge.
But for me it would be bad practice to throw the info to the second best location, because the first one has deficits.
That doesn't feel right for me.

So in conclusion the best way (agin -> from my point) would be to add the sitemaps section and think about how we can improve the sitemaps article in general.
As a first feeling i would say sitemaps has simply too much contents for "just one" page.
Also its a complex topic, so we could split those contents in useful parts in the future.

ThomDietrich:
Hey guys, the sitemaps article might be long but it is well structured and modular. Most of the time a user will just look for information on one aspect, e.g. the Switch item. They will find it under the according headline. Similarly you can find details on Visibility etc. Links between all section, where need, make sure that relations are clear. Splitting up the article would add imho add the confusion you want to avoid.

Regarding labels: Remember that Sitemaps are not the only way to access Items. Overwriting their Label is strongly linked to Sitemaps and should hence be part of the Sitemaps article. 

Once again: Strong linkage is key here. If you add a reference to the Items article saying "Beware: Item Labels can be overwritten in Sitemaps. See LINK" - this would bring the clarity you are wishing for.

I hope I was able to explain my opinion on the matter.

bgilmer77:
I agree with @ThomDietrich.  The sitemap article may be long, and in this age of TL;DR, some might argue that is a problem.  But I believe that it is well structured, and conveys the information users need in order to understand how sitemaps work.  Having new users troll through different locations for information on sitemaps is not a good approach.

PR #740, opened 2018-07-12
Update Null Annotations Conventions

wborn:
Updates the Null Annotations Conventions so they are in sync again with Eclipse SmartHome.

See also: 
https://www.eclipse.org/smarthome/documentation/development/conventions.html
https://github.com/eclipse/smarthome/blob/master/docs/documentation/development/conventions.md


wborn:
I think I need to update the base from gh-pages to master right? :-)

PR #742, opened 2018-07-13
Moved documentation for audio & voice actions to multimedia section

cweitkamp:
- Moved documentation for audio & voice actions to multimedia section
- Fixed actions with volume -> parameter is `PercentType`

I saw some confusion about audio & voice actions for DSL rules in the past days which leads me to do some changes in the docs. The [openHAB audio actions section](https://www.openhab.org/docs/configuration/actions.html#core-actions) lists supported functions only partially. The [openHAB multimedia section](https://www.openhab.org/docs/configuration/multimedia.html) contains information about audio and voice handling but not how to control the volume. To prevent double work and increase maintainability I moved everything into the openHAB multimedia section.

I noted the linked generated JavaDoc is not the latest version: https://www.eclipse.org/smarthome/documentation/javadoc/org/eclipse/smarthome/model/script/actions/Audio.html.


addons/actions.md
```diff
@@ -35,19 +35,10 @@ As a general rule, is better to call `MyItem.sendCommand(command)` and `MyItem.p
...
-- `increaseMasterVolume(float percent)`: Increases the volume by the given percent
-- `decreaseMasterVolume(float percent)`: Decreases the volume by the given percent
-- `float getMasterVolume()`: Returns the current volume as a float between 0 and 1
-- `playSound(String filename)`: Plays the given sound file. The file must be an mp3 or wav and located in `${openhab.home}/sounds`
-- `playStream(String url)`: Plays the audio stream at the given url
-- `say(String text)`: Says the given text through Text-to-Speech
-- `say(String text, String voice)`: Says the given text through the given voice (depends on the TTS engine and voices installed)
-- ` say(String text, String voice, String device)`: Says the given text through the given voice to the given output device (Mac OSX only).
-
-To get a list of available voices use `say -v ?` and to get a list of devices uses `say -a ?`.
```
cweitkamp: Does anybody know if this statement is true? Both expression did neither work for me in the DSL rules nor in the Karaf console.

Confectrician: I can confirm, that they don't work in the console.
Maybe we should replace them with the smarthome console commands.

Confectrician: There is already an example vor sinks, so adding the voices command would be enough.

Confectrician: Ah sorry, i missed that one...
there is always another section to expand in the files section. 😄

PR #762, opened 2018-08-22
replace .calendar with .zonedDateTime

Rosi2143:
getCalender() is deprecated getZonedDateTime() should be used

### DIFF
--- a/configuration/rules-dsl.md
+++ b/configuration/rules-dsl.md
@@ -436,13 +436,13 @@ With epoch, one can compare two dates together, convert a Joda DateTime to a Dat
 
 ```java
 // Get epoch from DateTimeType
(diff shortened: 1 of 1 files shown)


Confectrician:
Thanks for jumping in with this improvement.
Did you check other files too for instances of getcalendar()?

Rosi2143:
Hi,

no not yet. I will check when I'm back home.

Confectrician:
>no not yet. I will check when I'm back home.

It's not urgent, but would be a good chance to improve this global in the docs.
We can merge this PR righ away and check it in a second step, if you need more time. :)

Rosi2143:
I checked now and did not find any other place with getCalender -- yet.

PR #764, opened 2018-08-26
Update to Amazon Dash Button Binding Documentation

meingraham:
I have two suggested edits to the documentation.

The current write-up implies that the Thing declaration <thing_id> must be the Dash button's MAC address. In fact, the <thing_id> can be any valid string the user desires. Moreover, it is likely better to avoid using a version of the MAC address (i.e., the MAC address but with dashes instead of colons) to avoid having the MAC address hard coded in more than one place.

The second edit involves a statement regarding the ease of replacing a battery in a Dash button. In no case is "The built-in AAA battery can be easily replaced." In all cases, getting to the battery involves separating the two sections of the button's case. The pieces are glued together. Thus the seal must be "broken" and then later glued back. For the second generation button (model JK29LP) once the case is open, the AAA battery can indeed be easily replaced as it is in a typical battery clip. However, for first generation Dash buttons (model JK76PL), the battery terminals are soldered to the button's PCB and the battery itself is glued to the board. Replacing the battery in the older 1st generation buttons is far from easy... if even worth the effort. Probably the best approach is to stock up on a few spare buttons when they go on sale for $0.99. In fact, doing so now would get you the slightly easier to maintain 2nd gen buttons.

I am happy to make changes to the documentation... but I am far from being able to know how. Any guidance on the process of contributing documentation enhancements is appreciated.

Regards.

Mike


ghys:
Hi, I think you wanted to open an issue instead of a pull request.

This repository is for the "documentation" part i.e. https://www.openhab.org/docs/.
If you want to contribute to the reference docs for a specific add-on (or an user interface), the easiest is to click on the "Edit this page on GitHub" at the bottom of the page on openhab.org, in your case https://www.openhab.org/addons/bindings/amazondashbutton/ - it will take you to the adequate location where you can propose changes. Please review and try to follow the contributing rules, notably the sign-off requirements, before submitting the PR.

Thanks!

meingraham:
That's exactly what I needed. Thank you!

PR #770, opened 2018-09-03
Redirect automatically to pages on the new site

ghys:
I think the new site is well established now, so perhaps it's time to phase out the old docs.openhab.org?
I tried doing some search engine queries and while a lot of new URLs are now surfaced, there are still obsolete ones appearing as the content is still there.

This final commit to the gh-pages branch will replace all layouts by a redirect page (both meta link methods and with JavaScript - HTTP 301 redirects are not possible on GitHub Pages AFAIK); according to the info I've gathered this should transfer their PageRank on the new URLs, although it's not well documented:
- https://support.google.com/webmasters/answer/139066?hl=en
- https://www.seroundtable.com/google-meta-refresh-redirects-work-25335.html
- https://plus.google.com/+JohnMueller/posts/E4PqAhRJB2V

I believe the most important part here is having a `<link rel="canonical">` tag so that the search engines know the page has a new URL when indexing.

The series of "ifs" in header.html is the same as the one used for the link on the current warning banner, which should be valid for all pages.

What do you think?

### DIFF
--- a/_includes/header.html
+++ b/_includes/header.html
@@ -1,72 +1,63 @@
 <!DOCTYPE html>
+
+{% assign newsite = 'https://www.openhab.org' %}
+{% assign newurl = newsite | append: '/docs' | append: page.url %}
+{% if page.url == '/tutorials/' %}
+    {% assign newurl = newsite | append: '/docs/tutorial/' %}
+{% elsif page.url contains '/tutorials/beginner/' %}
+    {% assign newurl = newurl | replace: '/tutorials/beginner/', '/tutorial/' %}
+{% elsif page.url contains '/tutorials/migration.html' %}
+    {% assign newurl = newsite | append: '/docs/configuration/migration/' %}
+{% elsif page.url == '/introduction.html' %}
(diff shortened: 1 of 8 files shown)


kaikreuzer:
If I get it correctly, this will completely deactivate docs.openhab.org and redirect ALL urls, is that right?
What about the v2.1 version of the docs, will they still be accessible? Are there any other pages which do not exist on the new site and thus the redirect would result in an 404?

ghys:
> What about the v2.1 version of the docs, will they still be accessible?

Yes, they are static copies of the generated pages, so those will remain.

> Are there any other pages which do not exist on the new site and thus the redirect would result in an 404?

I think they're all covered, pages which don't exist anymore are redirected to the most relevant equivalent, did my best but can't guarantee it 100%.
This is deployed to https://ghys.github.io/openhab-docs/ so this can be tested out -
 for example https://ghys.github.io/openhab-docs/installation/openhabian.html, https://ghys.github.io/openhab-docs/addons/bindings.html

PR #778, opened 2018-09-16
Rename all variants to macOS

wborn:
The documentation contains a lot of (outdated) variants of the OS that runs on Apple computers named "macOS" nowadays.

### DIFF
--- a/installation/macos.md
+++ b/installation/macos.md
@@ -1,11 +1,11 @@
 ---
 layout: documentation
-title: Mac OS X
+title: macOS
 ---
 
 {% include base.html %}
 
-# Installation on Mac OS X
+# Installation on macOS
 
 This page is structured as follows:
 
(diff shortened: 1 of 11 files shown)


installation/macos.md
```diff
@@ -1,11 +1,11 @@
 ---
```
ghys: Renaming this file would require a change to:
https://github.com/openhab/website/blob/master/.vuepress/config.js#L209
as well.
This menu structure (at least for the docs & addons) should really be migrated to this repo at some point.

wborn: Thanks for the comment @ghys! I've now created https://github.com/openhab/website/pull/105 as well.
I was unsure if it would pick this up automatically but apparently it does not. :wink:

ghys:
LGTM, so I'm gonna go ahead and merge this, since it has to be synchronized with https://github.com/openhab/website/pull/105.
cc @Confectrician

PR #779, opened 2018-09-18
Fix configuration file path references

wborn:
The configuration files are stored in a directory named "conf" and not "config".


wborn:
Should we perhaps replace `conf` with some kind of placeholder because that directory could also be `/etc/openhab2`? See also [file locations](https://www.openhab.org/docs/installation/linux.html#installation).

BClark09:
I think this is also discussed in https://github.com/openhab/openhab-docs/issues/189

I think the best solution would be to use `$OPENHAB_CONF`, this is close to the folder structure used for manual installations, and if used literally for APT/YUM then this placeholder is parsed as `/etc/openhab2`.

Would it be possible on the site to auto create a tool tip for these placeholders (see https://github.com/openhab/openhab-docs/issues/189#issuecomment-360506100). i.e. if a webpage contains "$OPENHAB_CONF", then it is highlighted and hovering over it would display:

<table>
    <tr>
        <th>Placeholder</th>
        <th>Example on UNIX/MacOS</th>
        <th>Example on Windows</th>
        <th>Apt/Yum Packages</th>
    </tr>
    <tr>
        <td>$OPENHAB_CONF</td>
        <td>/opt/openhab/conf</td>
        <td>C:\\openhab\conf</td>
        <td>/etc/openhab2</td>
    </tr>
</table>

wborn:
> I think the best solution would be to use $OPENHAB_CONF

I see it's already being used like that at several locations in the documentation so I've now used that placeholder instead. We could also update the remaining `conf/` references in the documentation with `$OPENHAB_CONF`. WDYT?

BClark09:
Makes sense to me 🙂

PR #784, opened 2018-10-09
Updated Section DateTime Item

imtheuser:
The program code suggested in the documentation in section "// Convert DateTimeType to Joda DateTime" did not work for me.
@rkoshak proposed two options in this post:
https://community.openhab.org/t/error-timeinmillis-is-not-a-member-of-java-time-zoneddatetime/53079/3
Both worked for me.
I have therefore replaced the existing code with @rkoshak 's two options.

As this is my first contribution to the docs I hope I did everything right. Also the first time I use github.
I read the Contribution Guidelines and if I understand right, I have to add the following line to every git commit message. Not sure if I understand right because english is not my native langugage.

Signed-off-by: Rudolf Allgaier <meinpcpostfach@yahoo.de> (github: github_handle)

### DIFF
--- a/configuration/rules-dsl.md
+++ b/configuration/rules-dsl.md
@@ -442,8 +442,8 @@ val Number epoch = (MyDateTimeItem.state as DateTimeType).zonedDateTime.timeInMi
 val Number nowEpoch = now.millis
 
 // Convert DateTimeType to Joda DateTime
-val joda = new DateTime(MyDateTimeItem.state.toString)
-val joda = new DateTime((MyDateTimeItem.state as DateTimeType).zonedDateTime.toInstant.toEpochMilli)
+val jodaVariantOne = new DateTime(MyDateTimeItem.state.toString)
+val jodaVariantTwo = new DateTime((MyDateTimeItem.state as DateTimeType).zonedDateTime.toInstant.toEpochMilli)
 
 // Convert Joda DateTime to DateTimeType
 val calendar = java.util.Calendar::getInstance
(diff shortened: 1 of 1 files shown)


configuration/rules-dsl.md
```diff
@@ -442,7 +442,8 @@ val Number epoch = (MyDateTimeItem.state as DateTimeType).zonedDateTime.timeInMi
 val Number nowEpoch = now.millis
 
 // Convert DateTimeType to Joda DateTime
-val joda = new DateTime((MyDateTimeItem.state as DateTimeType).zonedDateTime.timeInMillis)
+val joda = new DateTime(MyDateTimeItem.state.toString)
+val joda = new DateTime((MyDateTimeItem.state as DateTimeType).zonedDateTime.toInstant.toEpochMilli)
```
Confectrician: Maybe we should call this value different.
Technically it would be an error to initialize a variable two times with the same name and i think openHAB will probably reject the rules file, if it is done like this.

I know this is only an example, but sometimes people paste docs code to try something out and we should take care of code that works right after pasting it into a file.

My suggestion would be:
```
val jodaVariantOne = new DateTime(MyDateTimeItem.state.toString)
val jodaVariantTwo = new DateTime((MyDateTimeItem.state as DateTimeType).zonedDateTime.toInstant.toEpochMilli)
```
What do you think?

imtheuser: `sometimes people paste docs code to try something out`
Yes, you're right. Me too.
Thank you!

Confectrician:
Ok, we are ready here.

Some additional info, if you would like to help us out here in the future too:
The sign-off part should usually be added to every commit you make.
I will add it now manually while merging your improvements.


And of course:
Thanks again for taking care of this improvement. :)

imtheuser:
Thank you for adding the sign-off part. 
Yet I'm not used to pull-requests and commits and the whole procedure. But hopfully I will be in the future.
Thanks again.

Confectrician:
You're welcome.
Feel free to contact me if there are any questions.

PR #721, opened 2018-06-14
Added google assistant to external ressources.

Confectrician:
Closes #690

Tested local and the maven build succeeded.
Assistant page was pulled and added.
Font matter is introduced in openhab/openhab-google-assistant#40

Signed-off-by: Jerome Luckenbach <github@luckenba.ch>

PR #788, opened 2018-10-14
process addons groovy update

Confectrician:
Added improvements based on discussion in #721


Confectrician:
@ghys can yu have a short look and merge this one and #721 ?
Would be nice if you do so, since it goes into the final branch.

All dependency PRs/Issues have been resolved meanwhile.

PR #799, opened 2018-10-25
Update items.md

Rossko57:
extra detail for numeric dynamic icons

### DIFF
--- a/configuration/items.md
+++ b/configuration/items.md
@@ -391,7 +391,7 @@ On the filesystem, the following icon files are provided by the user:
 Take note, that the Transformation used in the `Livingroom_Light_Connection` Item doesn't effect the needed state specific icons - the icon selection considers "myerror", not the contents of the `error.map` file.
 
 **Number State Matching Rule:**
-For Number Items the equal or next lowest state icon that can be found will be used.
+For Number or Dimmer Items the equal or next lowest state icon that can be found will be used. The default icon will be used for negative numers, or above 100 i.e. the available filename range is icon-0 to icon-99 only.
 For a dimmable light (0-100%), you might provide icons as in the example:
 
 | File name          | Description                                          |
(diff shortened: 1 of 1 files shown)


ThomDietrich:
Hey @Rossko57, not too sure about your addition. It is specific to Dimmers. Values below 0 or above 100 are undefined for a Dimmer item and hence are "Undefined" -> the default icon is used. This fact is given in the table but you could mention it specifically (but general) in the text.

Rossko57:
The rest of the paragraph refers to Numbers, which of course can have negative and >100 values.

Quite happy for any other suggested form of words, the intention is just to highlight that Number values -10 or +250.55 "won't work" with dynamic icons e.g. will default

configuration/items.md
```diff
@@ -391,7 +391,7 @@ On the filesystem, the following icon files are provided by the user:
 Take note, that the Transformation used in the `Livingroom_Light_Connection` Item doesn't effect the needed state specific icons - the icon selection considers "myerror", not the contents of the `error.map` file.
 
 **Number State Matching Rule:**
-For Number Items the equal or next lowest state icon that can be found will be used.
+For Number or Dimmer Items the equal or next lowest state icon that can be found will be used. The default icon will be used for negative numers, or above 100 i.e. the available filename range is icon-0 to icon-99 only.
```
Confectrician: I would remove the `or Dimmer` and extract that to a new sentence.
Also please break down to one sentence per line and make `The default..` appear on its own line.
(We do this to keep a nice track of changes per sentence in git.)

Confectrician: So my version would look like this:

> For Number Items...
The default...
Please notice that his also is valid for Dimmer Items since valus below 0 or above 100 are undefined for them.

Of course the sentences can be adapted.
This is just how i think it would look organized.

ThomDietrich:
```suggestion
For Number Items the equal or next lowest state icon that can be found will be used.
The default icon will be used as fallback for undefined states. 
Take note that a Dimmer Item is only defined for values between 0 to 100 and the default icon is shown outside that range.
```

ThomDietrich:
I need to raise a general question: Can a Dimmer Item actually take a value outside of the 0..100 range? Why? This feels wrong:

https://github.com/eclipse/smarthome/blob/master/bundles/core/org.eclipse.smarthome.core/src/main/java/org/eclipse/smarthome/core/library/items/DimmerItem.java#L80-L84

Confectrician:
>being limited to 0-100 anyway.

From my view this is already covered in the last sentence.

PR #802, opened 2018-11-04
Increase recommended JDK revision

pfink:
see openhab/openhab2-addons#4013


Confectrician:
Thanks. 👍

AngelosF:
(excuse the silly question...but...) how come this change doesn't appear (yet) on the live website?

Confectrician:
No silly question at all. :)

The website has a build mechanism for the docs that isn't executed on every change here.
It will run periodically and this commit will be included in the next build.

I think it's a bit different for the add-on repos, which will trigger at least a partial build after merges.

pfink:
@mstormi: Sorry for the late response! I were in vacations and then it got somehow lost... I didn't need to install JCE when using Java >= 161. I didn't tested it with the exact revision 161, but with the newest one at that time. The number is taken from the post linked at the start of this PR.

PR #832, opened 2018-12-24
Update ide.md

bdleedy:
Include possible need for Eclipse Projects - SmartHome during install.

### DIFF
--- a/developers/development/ide.md
+++ b/developers/development/ide.md
@@ -28,7 +28,7 @@ The Eclipse IDE is used for openHAB developments. The Eclipse Installer automati
 ![Step 0](images/ide0.png)
 3. Choose the "Eclipse IDE for Java Developers" and select "Next":
 ![Step 1](images/ide1.png)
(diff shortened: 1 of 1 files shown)


kaikreuzer:
No, this won't work - the openHAB IDE does not support the ESH repo in the workspace, so work on ESH should be done through the ESH IDE instead (but this will most likely change soon anyhow, so I'd suggest to not make changes to the docs right now.)

bdleedy:
Ok. It was the only way I could get it to work and found another post saying the same.


On Dec 24, 2018, at 4:13 AM, Kai Kreuzer <notifications@github.com> wrote:

No, this won't work - the openHAB IDE does not support the ESH repo in the workspace, so work on ESH should be done through the ESH IDE instead (but this will most likely change soon anyhow, so I'd suggest to not make changes to the docs right now.)

kaikreuzer:
No, we should rather find a way to make the IDE work with the latest IDE versions, see https://github.com/openhab/openhab-distro/issues/832.

PR #841, opened 2018-12-31
Replace old cloudbees.com snapshot link

MyRaceData:
Replace the old link for the snapshot at cloudbees.com with new link https://ci.openhab.org


kaikreuzer:
Thanks, but why did you remove the path to the resp. jobs that the links are supposed to point to?

MyRaceData:
Because I'm a moron? Sorry, I made a mistake. Hopefully I got it right this time and thank you for checking my poor work

PR #848, opened 2019-01-05
Changed Zulu table link

crxporter:
Link previously pointed to non java 8 download links. Changed to the same link as the above mentioned download link.


crxporter:
Sorry - didn’t realized I submitted twice. Thanks!

Confectrician:
Nope. Just another location that had a wrong link too. :)

PR #854, opened 2019-01-07
Changed smarthome:script (incorrect syntax)

vzorglub:
By the way, it doesn't work
**test.script**
```
logInfo("TEST", "SCRIPT")
```

`openhab> smarthome:script test.script`
```
 ___ test.script
   The method or field test is undefined; line 1, column 0, length 4
```

### DIFF
--- a/administration/runtime.md
+++ b/administration/runtime.md
@@ -107,7 +107,7 @@ Usage: smarthome:send <item> <command> - sends a command for an item
 
 | Command | Description |
 |---------|-------------|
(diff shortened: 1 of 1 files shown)


Confectrician:
>By the way, it doesn't work

I think thas has to be one of the jsr223 scripts like jython or javscript.
I don't think that there is something existing like `.script` files.

kaikreuzer:
> I don't think that there is something existing like .script files.

No? https://github.com/openhab/openhab-distro/blob/master/features/distro-resources/src/main/resources/scripts/demo.script

Confectrician:
🙈 

Goin to bed now...

PR #856, opened 2019-01-08
Update Docker installation documentation

wborn:
* Remove --user option which should not be used because it causes permission issues (https://github.com/openhab/openhab-docker/issues/221)
* Remove unnecessary --tty option when running as service, nowdays the container will be started in server mode when there is no tty (https://github.com/openhab/openhab-docker/pull/181)
* Update list of default environment variables
* Correct Ubuntu 16.1 to 16.04
* Nest 1.x binding uses .java whereas Nest 2.x binding does not

### DIFF
--- a/installation/docker.md
+++ b/installation/docker.md
@@ -88,7 +88,6 @@ If using docker to manage the service, run the following command:
 docker run \
         --name openhab \
         --net=host \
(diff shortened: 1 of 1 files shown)


cniweb:
LGTM

PR #858, opened 2019-01-09
Hotfix/workaround openhab distro issue 423

stefan-muc:
Two small changes to documentation, see https://github.com/openhab/openhab-distro/issues/423 for background

### DIFF
--- a/installation/security.md
+++ b/installation/security.md
@@ -140,6 +140,7 @@ server {
         proxy_set_header X-Real-IP            $remote_addr;
         proxy_set_header X-Forwarded-For      $proxy_add_x_forwarded_for;
         proxy_set_header X-Forwarded-Proto    $scheme;
+        proxy_set_header X-Forwarded-Scheme   $scheme; # workaround for https://github.com/openhab/openhab-distro/issues/423
     }
 }
 ```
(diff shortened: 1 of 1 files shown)


Confectrician:
So this one will be outdated when openhab/openhab-distro#840 got merged?
In this case we should not merge it. At least not into master.

stefan-muc:
@Confectrician 
0e37515 is outdated as soon as https://github.com/openhab/openhab-distro/pull/840 got merged and a new release including this fix will be released. I proposed it so that users have a workaround in between.

5189202 is an improvement that I'd like to see merged / cherry-picked. It took me some time (and and search) to figure that out, so it might save other users some time.

Confectrician:
Hey @stefan-muc, 

theoretically i could cherry pick 5189202 this, but unfortunately the commit was not signed-off
You would have to `sign-off` your change according to our [contribution guidelines](https://github.com/openhab/openhab-docs/blob/gh-pages/CONTRIBUTING.md#sign-your-work).

The easiest solution for both of us would be to open a new pull request with only the chagnes from 5189202 and add a sign off statement like this

> Signed-off-by: Joe Smith <joe.smith@email.com> (github: github_handle)

to your commit message.
I think this is less work than trying to remove one commit here and change the second one.

Would that be ok for you?

stefan-muc:
It's okay for me, see new pull request above - I just hope that I won't get spam to this address, as this is now public (GitHub normally hides mail addresses for commits for non-logged in users).

It also would be good to have the requirement of the sign-off text as template while creating a pull request (as other projects also have their templates), so I wouldn't have created that pull request without sign-off in first place.

Confectrician:
>It also would be good to have the requirement of the sign-off text as template while creating a pull request (as other projects also have their templates), so I wouldn't have created that pull request without sign-off in first place.

Valid suggestion! I will take care of that.

PR #863, opened 2019-01-12
Fix app links in addons/uis

LightIsLife:
Addressing comments from code review.


LightIsLife:
Ghee, so sorry for the chaos I am making here.....github beginners confusion with repos and branches...

PR #888, opened 2019-01-31
Update Things.md in order to show exampels for file created things an…

JueBag:
…d bridges

As the new mqtt binding has a much larger audience then other OH2-bindings which already used bridges the missing examples of bridge definitions in the openHAB core documentation became more then obvious.
I used the text parts from the eclipse smarthome documentation and added an example of a mqtt bridge with generic things.

Signed-off-by: Juergen Baginski opus42@gmx.de (github: JueBag)

### DIFF
--- a/configuration/things.md
+++ b/configuration/things.md
@@ -128,21 +128,21 @@ For the contained notation of things the UID will be inherited and the bridge ID
 
 **Example of a MQTT Bridge with Generic MQTT Things :**
 ```xtend
-Bridge mqtt:broker:MyMQTTBroker [ host="192.168.178.50", secure=false, username="MyUserName", password="MyPassword"]
-{
-Thing topic sonoff_Dual_Thing "Light_Dual" @ "Sonoff" {  
+Bridge mqtt:broker:MyMQTTBroker [ host="192.168.178.50", secure=false, username="MyUserName", password="MyPassword"] {
+  Thing topic sonoff_Dual_Thing "Light_Dual" @ "Sonoff" {  
     Channels:
-        Type switch : PowerSwitch1  [ stateTopic="stat/sonoff_dual/POWER1" , commandTopic="cmnd/sonoff_dual/POWER1", on="ON", off="OFF"]
-        Type switch : PowerSwitch2  [ stateTopic="stat/sonoff_dual/POWER2" , commandTopic="cmnd/sonoff_dual/POWER2", on="ON", off="OFF"]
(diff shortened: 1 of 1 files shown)


configuration/things.md
```diff
@@ -64,15 +64,19 @@ See the [configuration tutorial]({{base}}/tutorials/beginner/configuration.html)
...
-Benefits of defining Things, Items and other aspects of openHAB in configuration text files are, that they are statically defined, unambiguous, flexible and easy to backup and restore.
-The main downsides of configuration files are the effort needed to compose them and the probability for typing errors.
-
 The syntax for `.things` files is defined as follows (parts in `<..>` are required):
 
 ```xtend
 Thing <binding_id>:<type_id>:<thing_id> "Label" @ "Location" [ <parameters> ]
 ```
 
+The first keyword defines whether the entry is a bridge or a thing. The next statement defines the UID of the thing which contains of the following three segments: binding id, thing type id, thing id. So the first two segments must match to thing type supported by a binding (e.g. `network:device` or `astro:moon`), whereas the thing id can be freely defined. Optionally, you may provide a label in order to recognize it easily, otherwise the default label from the thing type will be displayed.
```
hakan42: Could please put each sentence on a line by itself? This makes reviewing further changes much easier, and the website builder takes care of rendering proper paragraphs.

JueBag: Done.

configuration/things.md
```diff
@@ -64,15 +64,23 @@ See the [configuration tutorial]({{base}}/tutorials/beginner/configuration.html)
...
 ```xtend
 Thing <binding_id>:<type_id>:<thing_id> "Label" @ "Location" [ <parameters> ]
 ```
 
+The first keyword defines whether the entry is a bridge or a thing. 
+The next statement defines the UID of the thing which contains of the following three segments: binding id, thing type id, thing id. 
+So the first two segments must match to thing type supported by a binding (e.g. `network:device` or `astro:moon`), whereas the thing id can be freely defined. 
+Optionally, you may provide a label in order to recognize it easily, otherwise the default label from the thing type will be displayed.
+
+To help organizing your things, you also may define a location (here: `Location`).
```
Confectrician:
```suggestion
To help organizing your things, you also may define a location (`Location` in the example above).
```

JueBag: Agreed on all three changes!
I'll change the spaces in the example after this one.

Signed-off-by: Juergen Baginski opus42@gmx.de (github: JueBag)

PR #889, opened 2019-01-31
Remove current esh concept docs and also remove the import from pom.xml

Confectrician:
This PR removes the whole concepts area and also the import from ESH which was defined in pom.xml
It is a preparation for reintegrating the conepts area in the master branch of the openhab-docs repo which will follow in a second comment right away.


davidgraeff:
Why is this still not merged?

PR #892, opened 2019-02-02
Element Type 'Slider' does not yet support minValue, maxValue and step

igitti:
minValue, maxValue and step for slider is not implemented yet
https://github.com/eclipse/smarthome/pull/6777

### DIFF
--- a/configuration/sitemaps.md
+++ b/configuration/sitemaps.md
@@ -309,7 +309,7 @@ Setpoint item=KI_Temperature label="Kitchen [%.1f °C]" minValue=4.5 maxValue=30
 ### Element Type 'Slider'
 
 ```perl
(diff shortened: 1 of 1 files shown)


Confectrician:
Hey @igitti  and thanks for your interest in contributing.

From what i have read so far, the feature has been merged and will be (or is already) included in one of our milestone or snapshot builds.
Since our documentation master branch is ths source of our "latest" documentation there should be no need in removing this.

When you use the release version of our docs the min and max values aren't included.

See here:
https://www.openhab.org/v2.4/docs/configuration/sitemaps.html#element-type-slider

You can set the docs version via dropdown:
image

igitti:
Hi @Confectrician,

really? I'm sorry for the wrong contribution. Shame on me.

Confectrician:
No need to apologize.
The situation here is not the best in general.

We have talked more than once about showing the latest *release version* of the docs as default and not the *experimental/snapchot/whatever* version, exactly because of situations like this.
I will try to reanimate the discussion about this again and hopefully get a better long term solution finally.

Issue #895, opened 2019-02-12
Add clarification to the AC Governance process


rkoshak:
See https://community.openhab.org/t/provide-policy-and-procedure-for-public-comment/67445 for details. We need to document the process for engagement between the AC and the community.

kaikreuzer:
Closing due to https://community.openhab.org/t/provide-policy-and-procedure-for-public-comment/67445/6.

PR #896, opened 2019-02-12
Added clarification to the AC governance

rkoshak:
I added clarification that the clock doesn't start to tick on a vote until there is actually a call for a vote. 

Added a section to describe how the AC can generate an idea and elicit feedback and discussion from the community at large. This section also discusses how a member of the community can bring the attention of the AC to a topic that may need AC input.

Solves #895

Signed off by Richard Koshak <rlkoshak@gmail.com>

### DIFF
--- a/developers/contributing/governance.md
+++ b/developers/contributing/governance.md
@@ -83,7 +83,7 @@ The Architecture Council is responsible for all decisions that have a significan
 
 Discussions and decisions are done in a [dedicated category](https://community.openhab.org/c/development/architecture) in the forum.
 Decisions are taken unanimously.
-This is the case if no member of the AC vetos against a proposed resolution within a week.
+This is the case if no member of the AC vetos against a proposed resolution within a week of the call for votes on the topic.
 
 The AC is the entity entitled to do changes to the governance rules.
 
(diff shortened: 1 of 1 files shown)


developers/contributing/governance.md
```diff
@@ -105,3 +105,16 @@ AC members can be excluded from the AC through an unanimous vote by the other AC
 When excluding a member, a reason has to be provided, such as inactivity, mis-behaviour, etc. and the excluded member must be given the chance to publicly respond to this.
 
 The openHAB Foundation's General Assembly can request that the AC is dissolved and has to be newly appointed.
+
+### Architecture Council Engagement
+
+Any member of the AC can generate a proposal for a decision to be made. 
+They do so by opening a topic in the [dedicated Architecutre category](https://community.openhab.org/c/development/architecture) on the forum.
```
openhab-5iver: Typo... Architecture

Confectrician:
```suggestion
They do so by opening a topic in the [Architecture category](https://community.openhab.org/c/development/architecture) on the forum.
```

developers/contributing/governance.md
```diff
@@ -105,3 +105,16 @@ AC members can be excluded from the AC through an unanimous vote by the other AC
 When excluding a member, a reason has to be provided, such as inactivity, mis-behaviour, etc. and the excluded member must be given the chance to publicly respond to this.
 
 The openHAB Foundation's General Assembly can request that the AC is dissolved and has to be newly appointed.
+
+### Architecture Council Engagement
```
kaikreuzer: Sleeping about this and seeing the actual effect of such a process, I think this goes against my original reason for establishing the AC, which is to remove bottlenecks and speed up decision making. A "government" does not need to do a people vote on every decision. This should at max be a case by case decision, if there's any doubt. So I'd suggest to leave not add this section as it effectively is not "a more detailed description", but rather a change in the process.

kaikreuzer:
I've included the first change in https://github.com/openhab/openhab-docs/pull/898, so this PR can be closed.

PR #897, opened 2019-02-14
Updated docs wrt IDE setup (both add-on and core development IDEs)

kaikreuzer:

### DIFF
--- a/developers/development/ide.md
+++ b/developers/development/ide.md
@@ -19,7 +19,7 @@ Please ensure that you have the following prerequisites installed on your machin
 1. [Maven 3.x](https://maven.apache.org/download.cgi) (optional, Eclipse m2e can also be used)
 1. [Oracle JDK 8](http://www.oracle.com/technetwork/java/javase/downloads/jdk8-downloads-2133151.html)
 
-## Installation
+## Installing the Eclipse IDE
 
 The Eclipse IDE is used for openHAB developments. The Eclipse Installer automatically prepares the IDE so that it comes with all required plug-ins, the correct workspace encoding settings, pre-configured code formatters and more. Simply follow these steps:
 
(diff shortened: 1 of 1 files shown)


developers/development/ide.md
```diff
@@ -28,41 +28,50 @@ The Eclipse IDE is used for openHAB developments. The Eclipse Installer automati
...
+### Option 2: Installation of IDE for Core Development
 
-Running:
+1. Select (double-click) the "openHAB Core Development" item and click "Next".
+2. Now provide an installation folder (don't use spaces in the path on Windows!) and your Github id (used to push your changesets to) and select "Next".
+3. The installation begins when you press "Finish".
+4. Trust any certificate you are asked for.
+5. Once it is done, you will see the Eclipse Welcome Screen, which you can close by clicking "Workbench" on the top right. The initial tasks are executed. Wait until the bottom right corner of the IDE does not show "Build" or "Refreshing workspace" or any other progress messages.
+6. Select "Help", "Perform Setup Tasks..." and press "Finish".
+7. The dialog disappears but could be opened again using the icon in the (bottom) status bar (near to the left side). You should open it and check it, because the chance if high that a restart is requested. The process (perform setup tasks) continues after the restart has been done. You can bring the dialog to the front after the restart using the same icon. Wait until all tasks have finished (wait for an empty buttom right status field).
```
openhab-5iver: Typo... if -> is

openhab-5iver:
I have my IDE working again, with Paper UI, but I selected both addons and core. The autobuild appears to get stuck in a loop, and I've had to do manual builds. I'll reinstall, but maybe we want to provide an explanation why both shouldn't be used in the same setup?

kaikreuzer:
@openhab-5iver The reason is mentioned here: https://github.com/openhab/openhab-docs/pull/897/files#diff-9cf29213cd7db30ca870a6b8602189bbR33

developers/development/ide.md
kaikreuzer: fixed.

openhab-5iver:
That is the reason, but I was suggesting to include some detial on what will happen if you do. As it reads, it is not clear if it is OK to select both. Providing some details on what the side effects will be may help clarify this, and to help troubleshoot the resulting issues. Just my opinion!

kaikreuzer:
It clearly states that you have to different paths (which means the one excludes the other). Not sure why we should document what happens if you do not follow the documentation...

Technically, it actually IS possible to install both, but I didn't want to mention it because such a setup will rather confuse people a lot (and I would not want to support them in that).

Issue #902, opened 2019-02-19
[Eclipse IDE setup] Update ide picture


Jochen1980:
Users are thinking that they should pick a very old version of Eclipse IDE because the picture shows so. Therefore we should change and update it.

davidgraeff:
Close?

Jochen1980:
Actually I was not able to verify it, but we can close it and I will verify it later.

PR #903, opened 2019-02-19
[Eclipse IDE setup] Update IDE picture for installation

Jochen1980:
see #902


davidgraeff:
The changes list is empty?

Jochen1980:
Dear David, 

I just changed the picture - the file name is the same, so I guess it looks like 'no changes'.

Best regards,
Jochen

davidgraeff:
Sweet. It would be even more helpful if you focus the list view onto the extended Github projects->openHAB selection, don't you agree?

Jochen1980:
Actually I do not think so in that picture. The extended Github-projects view will show up after hitting "Next". In my opinion it would be more appropriate to add another picture if you want to focus on that matter.

Just my two cents ...

PR #921, opened 2019-03-16
Update restore state on restart documentation

jswim788:
See discussion with rossko57 here: https://community.openhab.org/t/nightmare-stop-oh2-turning-on-my-relays/70016/15
Update "Restoring Item States on Restart" to give an example of two Items instead of all Items. Generally restoring all states has led to issues, so it is better to guide users to restore specific Items.

Signed-off-by: John Schmitz jswim400im@gmail.com (github: jswim788)

### DIFF
--- a/configuration/persistence.md
+++ b/configuration/persistence.md
@@ -135,11 +135,11 @@ Items {
 
 ## Restoring Item States on Restart
 
-When restarting your openHAB installation you may find there are times when your logs indicate some Items have the state, `UNDEF`.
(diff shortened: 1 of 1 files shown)


jswim788:
No idea why the DCO fails - I think I have the write format for it?

Confectrician:
Hi,

I think the dco fails because the signoff was not done in the commit message.

But that's nothing to worry about here.
I will add it to the merge commit when merging this one.

Will have a look tonight when I am not on mobile phone.

PR #922, opened 2019-03-21
Update configuration.md

swissbyte:
Clarified the meaning of things in terms of being "ONLINE"
This was confusing to me, since im a beginner and it was written wrong in the original tutorial

### DIFF
--- a/tutorials/beginner/configuration.md
+++ b/tutorials/beginner/configuration.md
@@ -78,7 +78,7 @@ There you will find your network device thing, including the name "John's Mobile
 
 ![](images/picture_12.jpg)
 
-You can already see the status, "ONLINE", of the thing, so it's currently reachable by openHAB.
+You can already see the status, "ONLINE", of the thing, this is confusing for new beginners since "ONLINE" here means not that the Device is reachable through the network. Instead it means that the thing is "present" or "usable" for openhab. The online state of the network device is only accessible through the channel "online".
 If you click on the name, you'll get to the next page showing the available channels of the thing.
 Channels represent the different functions the items provide.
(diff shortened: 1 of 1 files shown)


Confectrician:
Hey @swissbyte and thanks for your contribution.

I am a bit unsure if this is the correct place and way of improving the situation.
Not every thing has an "online" channel, so we would add a wrong information here.

Maybe we should break it down to an explanation that "online" just means reachable from the openHAB installation.

tutorials/beginner/configuration.md
```diff
@@ -78,7 +78,7 @@ There you will find your network device thing, including the name "John's Mobile
 
 ![](images/picture_12.jpg)
 
-You can already see the status, "ONLINE", of the thing, so it's currently reachable by openHAB.
+You can already see the status, "ONLINE", of the thing, this is confusing for new beginners since "ONLINE" here means not that the Device is reachable through the network. Instead it means that the thing is "present" or "usable" for openhab. The online state of the network device is only accessible through the channel "online".
```
Confectrician: I would suggest something like this:

```suggestion
You can already see the status, "ONLINE", of the thing, which only means that the thing is **present** or **usable** for openHAB.
It doesn't give any indication of the real reachability of this device/api through the network,
```

What do you think?

davidgraeff:
I will make a breaking change with the network binding for OH 2.5.
The Thing status will represent the reachability from there on and the "online" channel will be removed.

It is not necessary to add that misleading information to the documentation. It is a fault of the network binding.

Confectrician:
>It is not necessary to add that misleading information to the documentation. It is a fault of the network binding.

Thanks for this information.
I will close this PR then and we will solve the problem on code base. 🙂

PR #924, opened 2019-03-23
Update ide.md

davidgraeff:
Add deprecated status to the IDE setup for binding development.

### DIFF
--- a/developers/development/ide.md
+++ b/developers/development/ide.md
@@ -35,6 +35,13 @@ Both are currently based on different build systems and thus require separate ID
 
 ### Option 1: Installation of IDE for Add-on Development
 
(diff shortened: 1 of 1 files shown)


davidgraeff:
The page was not updated? I thought netlify is immediately pushing the page?

Confectrician:
Nope, the automatic triggering is done when something happens in the addon repos for those external ressources.

We didn't agree on a solution yet for this repo.
Maybe a schedule would be better here.
Currently i am triggering direct changes manually in jenkins.

davidgraeff:
You can setup a travis cron job for once a week.
It will have a specific environment variable set to identify it.

PR #927, opened 2019-04-02
Added warning and small updates

openhab-5iver:
I did some tidying up, corrected some of the classes that changed due to the ESH reintegration, and added a warning about using anything other than the official APIs (requested here... https://github.com/openhab/openhab-core/pull/663#issuecomment-478233196).

Signed-off-by: Scott Rushworth openhab@5iver.com

### DIFF
--- a/configuration/jsr223.md
+++ b/configuration/jsr223.md
@@ -13,103 +13,63 @@ title: JSR223 Scripting
 
 ## Overview
 
-[JSR223](https://docs.oracle.com/javase/6/docs/technotes/guides/scripting/) ([spec](https://jcp.org/aboutJava/communityprocess/pr/jsr223/index.html)) 
-is a standard scripting API for Java Virtual Machine (JVM) 
-[languages](https://en.wikipedia.org/wiki/List_of_JVM_languages). 
+[JSR223](https://docs.oracle.com/javase/6/docs/technotes/guides/scripting/) ([spec](https://jcp.org/aboutJava/communityprocess/pr/jsr223/index.html)) is a standard scripting API for Java Virtual Machine (JVM) [languages](https://en.wikipedia.org/wiki/List_of_JVM_languages). 
 The JVM languages provide varying levels of support for the JSR223 API and interoperability with the Java runtime. 
 Currently the following languages are known to work well for openHAB scripting: 
-[**Oracle Nashorn**](http://www.oracle.com/technetwork/articles/java/jf14-nashorn-2126515.html) (Javascript bundled with Java),
-[**Jython**](http://www.jython.org/) (Python on the JVM) and
+[**Jython**](https://jython.github.io/) (Python on the JVM), 
+[**Nashorn Javascript**](http://www.oracle.com/technetwork/articles/java/jf14-nashorn-2126515.html) (ECMAScript implementation included in JDK8 through 10, [deprecated in 11](https://openjdk.java.net/jeps/335)), and 
 [**Apache Groovy**](http://www.groovy-lang.org/) (JVM scripting language).
 
-Although JSR223 scripts can be used as a general-purpose extension language for openHAB, 
-the most common usage will be for defining openHAB rules.
+Although JSR223 scripts can be used as a general-purpose extension language for openHAB, it is most commonly used for the creation of rules, and within scripted Actions or Conditions. 
+Currently, openHAB allows JSR223 scripting to access all packages, which may not be included in the official APIs. 
+This provides great flexibility for the users of JSR223, but is also *use at your own risk*, since changes outside of the offical APIs occur frequently, and are not considered to be *breaking changes*.
+New APIs are planned to be implemented in the future, which will provide standardized interfaces for interacting with openHAB through scripted automation.
 
 ### Language-Specific Documentation
 
-- [Javascript](jsr223-js.html)
 - [Jython](jsr223-jython.html)
+- [Javascript](jsr223-js.html)
 - Groovy (TODO)
 
 ### Script Locations
 
-Scripts should be placed in the 'USERDATA/automation/jsr223' directory. For example, 
-in Linux installations created with a package installer, 
-the directory would be `/etc/openhab2/automation/jsr223`.
+Scripts should be placed in the `${OPENHAB_CONF}/automation/jsr223/` directory. 
+This directory will vary, [based on the type of openHAB installation used](https://www.openhab.org/docs/installation/linux.html#installation).
+For example, Linux installations created with a package installer will use `/etc/openhab2/automation/jsr223/`, and manual installations will use `/opt/openhab2/conf/automation/jsr223/`.
 
-At system start time the scripts will be loaded in an order based on their name 
-and then top-down through the directory hierarchy. 
-Note that while scripts may be placed in subdirectories of `jsr223`, 
-they should not have the same names as scripts in other directories (this may be changed in the future). 
-For example, with the following directory structure...
+When openHAB starts, scripts will be loaded in an order based on their file name and path. 
+For example, with the following scripts and directory structure...
 
 ```text
-automation/jsr223
-  01/
-    script1.py
-    scriptx.py
-  00script.py
-  script2.py
+├── automation/jsr223
+│   ├── 00
+│   │   ├── 00_00.py
+│   │   └── 01_00.py
+│   ├── 00.py
+│   ├── 01
+│   │   ├── 00_01.py
+│   │   └── 01_01.py
+│   └── 01.py
 ```
 
-the load order will be: `00script.py`, `01/script1.py`, `script2.py`, `01/scriptx.py`.
-
-<a name="trigger_types"></a>
-
-### Trigger Types (all JSR223 languages)
-
-The following trigger types are defined by openHAB (custom triggers can also be defined)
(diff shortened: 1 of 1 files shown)


configuration/jsr223.md
```diff
@@ -13,46 +13,44 @@ title: JSR223 Scripting
...
-[**Jython**](http://www.jython.org/) (Python on the JVM) and
+[**Jython**](https://jython.github.io/) (Python on the JVM), 
+[**Nashorn Javascript**](http://www.oracle.com/technetwork/articles/java/jf14-nashorn-2126515.html) (ECMAScript implementation included in JDK8 through 10, [deprecated in 11](https://openjdk.java.net/jeps/335)), and 
 [**Apache Groovy**](http://www.groovy-lang.org/) (JVM scripting language).
 
-Although JSR223 scripts can be used as a general-purpose extension language for openHAB, 
-the most common usage will be for defining openHAB rules.
+Although JSR223 scripts can be used as a general-purpose extension language for openHAB, it is most commonly used for the creation of rules, and within scripted Actions or Conditions. 
+Currently, openHAB allows JSR223 scripting to access all packages, which may not be included in the official APIs. 
+This provides great flexibility for the users of JSR223, but is also 'use at your own risk', since changes outside of the offical APIs occur frequently, and are not considered to be 'breaking changes'.
```
Confectrician:
```suggestion
This provides great flexibility for the users of JSR223, but is also *use at your own risk*, since changes outside of the offical APIs occur frequently, and are not considered to be *breaking changes*.
```

configuration/jsr223.md
```diff
@@ -13,46 +13,44 @@ title: JSR223 Scripting
...
 - [Jython](jsr223-jython.html)
+- [Javascript](jsr223-js.html)
 - Groovy (TODO)
 
 ### Script Locations
 
-Scripts should be placed in the 'USERDATA/automation/jsr223' directory. For example, 
-in Linux installations created with a package installer, 
-the directory would be `/etc/openhab2/automation/jsr223`.
+Scripts should be placed in the '${OPENHAB_CONF}/automation/jsr223/' directory.
```
Confectrician:
```suggestion
Scripts should be placed in the `${OPENHAB_CONF}/automation/jsr223/` directory. 
```

configuration/jsr223.md
```diff
@@ -61,48 +59,61 @@ the load order will be: `00script.py`, `01/script1.py`, `script2.py`, `01/script
...
+|--|--|--|
 | `eventTopic` | default: "smarthome/*" (for channel events: "smarthome/channels/*/triggered") | TEXT |
 | `eventSource` | item name, channel UID | TEXT |
 | `eventTypes` | ItemCommandEvent, ItemStateEvent | TEXT |
 
 ### Additional Information
 
 - [Scripted Rule Support](https://github.com/eclipse/smarthome/wiki/Scripted-Rule-Support) - 
-    Early documentation on JSR223 usage in openHAB 2. Includes IDE setup instructions.
+    Early documentation on JSR223 usage in openHAB 2, incliuding IDE setup instructions
```
Confectrician:
```suggestion
    Early documentation on JSR223 usage in openHAB 2, including IDE setup instructions
```

configuration/jsr223.md
```diff
@@ -61,48 +59,61 @@ the load order will be: `00script.py`, `01/script1.py`, `script2.py`, `01/script
...
 
 <a name="presets"></a>
 
 ### Predefined Script Variables (all JSR223 languages)
 
-To faciliate JSR223 scripting, several openHAB-related variables are automatically predefined. These presets include:
+To faciliate JSR223 scripting, several openHAB-related variables are automatically predefined, and can be loaded into the script context using `scriptExtension.importPreset(<name of preset>)`.
```
Confectrician:
```suggestion
To faciliate JSR223 scripting, several openHAB-related variables are automatically predefined and can be loaded into the script context using `scriptExtension.importPreset(<name of preset>)`.
```

openhab-5iver:
I had just noticed the missing sign-off... so easy to do when making changes directly through GH. I just setup a load repo to ammend the commit, and will add in your suggestions.

openhab-5iver:
I couldn't get the tables to look the way I wanted with markdown, so I converted it to HTML and put them all on one. Hopefully, that will be acceptable. IMO, it looks much cleaner.

Confectrician:
>Hopefully, that will be acceptable.

Unfortunately we are avoiding html tables where possible and those where nested as far as i remember.
What was the problem?

openhab-5iver:
> What was the problem?

There wasn't a problem... just a limitation of markdown tables. Using HTML, I eliminated using separate tables for each trigger and repeating the column headers by spanning rows and including the trigger names inside the table. TMK, you can't span rows using markdown.

How about using...
```
<details>
    <summary>core.ItemStateChangeTrigger</summary>

| Parameter | Description |
|-----------|-------------|
| `itemName` | The name of the `Item` |
| `previousState` | The previous `State` (optional) |
| `state` | The `State` (optional) |
</details>
```
It's still using some HTML, but the markdown tables can be used with it.

<details>
    <summary>core.ItemStateChangeTrigger</summary>

| Parameter | Description |
|-----------|-------------|
| `itemName` | The name of the `Item` |
| `previousState` | The previous `State` (optional) |
| `state` | The `State` (optional) |
</details>

Confectrician:
>How about using...

I think we could test that.

openhab-5iver:
@Confectrician, this is ready for another review. I used `<details>` and moved the TriggerTypes to the bottom. I also corrected the explanation of the script loading order, which was incorrect. I might have also changed some other minor things, but I don't recall, so this probably needs another full review.

openhab-5iver:
@Confectrician, I thought the documentation changes showed up immediately after being merged. I must not understand the process... what am I missing?

https://www.openhab.org/docs/configuration/jsr223.html

Confectrician:
## TL;DR:

Website build process is waiting for openHAB Distribution, which has a failing build at the moment.

## Explanation for interested

See https://ci.openhab.org/view/Documentation/

There are two things necessary for getting changes from here into the website:
1. They have to be prepared by the `merge docs` job
2. They have to be deployed by the `redeploy website` job

### So whats the status of this PR

1. Your changes should be prepared for the next website build.
The `merge docs` job runs twice a day currently and copys changes to the final branch.
2. There was no `redeploy website` job yet.
This is caused by the failing distribution build.
We have a mechanism that works like this:
-> `distribution build` successful 
-> start `gather external docs` job 
-> `gather external docs` job successful 
-> start `redeploy website` job

It seems that the distribution job is a bit unstable due to the bigger chagnes that are currently done.
I have startet the `gather external docs` job manually now.

davidgraeff:
The process / that post should be part of the main readme file of this repo. I also wondered what is going on.

Is there a reason that the distribution is involved?

Confectrician:
Yes there was a reason,
but i can't explain it word by word now and would have to lookup the corresponding issues here too.

What comes to my mind:
Distribution build is the point where newly changes from the addon repos are available in the unstable version. (Correct me if i am wrong)
That is the point where we wanted to grab all external changes and update the website.

Since changes in this repository are not that often, there was a decision to not trigger this build that often.
I have no added a time based trigger which runs twice a day.
Changes should then get included in the next website build after the distribution succeeds,
which usually is fast enough, when theres nothing failing with that build.


>The process / that post should be part of the main readme file of this repo.

Agreed.
I will add this information, alltough i wanted to rewrite that anyway at some point.
But since i didn't make it that far yet its fair to add this already.

davidgraeff:
> Distribution build is the point where newly changes from the addon repos are available in the unstable version. 

But to crawl all the readme files of the addons repo and other text files the distribution doesn't need to be build successfully. That's what I'm wondering.

Confectrician:
Yeah, but the code changes in the repos are getting "relevant" when they arrive at the distribution, am i right?

So we don't need to build each and every time something has been comitted to one of the external repos. With the distro build we can add them in bulk.

Maybe @ghys will jump in and tell us if there are other reasons.
He made the initial build job configuration.

davidgraeff:
I think it was really done to de-bounce the build, let's see what Yannick has to say to this matter

ghys:
Sorry missed that 😕 
The point was indeed to update the website only when there were "significant changes", because updating the website means invalidating the caches in users' browsers (and we were already walking on eggshells with Netlify about the amount of traffic, fortunately it has been resolved just a few days ago, now we have a plan with ample quota).

At first everything was manual, usually triggered by @Confectrician, then someone (not me) added gather-external-docs as downstream to the distro build in Maven.

I would have no objections of adding redeploy-website as downstream to merge-docs so that it's updated after a successful master>final merge, but only if there are changes (with the twice-a-day timer it wouldn't always be the case) to avoid invalidating users' caches for nothing, and preferably not too often - so that if PRs are merged in bulk there wouldn't be as many redeploys. That's why merge-docs was initially triggered manually until we came up with a more clever solution.

ghys:
What could maybe work, since merge-docs and gather-external-docs both end up committing on the _final_ branch here, is to make update-website trigger on a SCM change on that branch 🤔 (and remove the downstreams).

Confectrician:
Sounds good and is worth a test @ghys.
I will adapt the update trigger and remove it from the gather external docs job aside.

PR #939, opened 2019-04-16
guidelines:  exception stack traces

davidgraeff:
Kai phrased it very good on when to print exception stack traces.
See https://github.com/openhab/openhab-core/pull/715#issuecomment-482879882

### DIFF
--- a/developers/development/guidelines.md
+++ b/developers/development/guidelines.md
@@ -130,17 +130,21 @@ void myFun() {
 }
 ```
 
-* Where ever exceptions are caught and logged, the exception should be added as a last parameter to the logging. 
+* Decide on a case by case base, if you want exception stack traces in the log.
 
+If you add an exception as a last parameter to the logging, the stack trace will be printed.
+ 
 ```java
 void myFun() {
   try {
     doSomething();
   } catch (IOException e) {
-    logger.warn("Explain where we are in the code", e);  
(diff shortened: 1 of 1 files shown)


developers/development/guidelines.md
```diff
@@ -130,18 +130,22 @@ void myFun() {
 }
 ```
 
-* Where ever exceptions are caught and logged, the exception should be added as a last parameter to the logging. 
+* Decide on a case by case base, if you want exception stack traces in the log.
```
openhab-5iver: I think you can replace this with...

Exceptions with stacktraces in the log are considered to be bugs in your binding that should be reported and fixed. 
If you add an exception as a last parameter to the logging, the stack trace will be printed. 
Configuration errors by users should only print log messages about what's wrong. In that case you would use `e.getMessage()`.

developers/development/guidelines.md
```diff
@@ -130,18 +130,22 @@ void myFun() {
...
 
+If you add an exception as a last parameter to the logging, the stack trace will be printed.
+ 
 ```java
 void myFun() {
   try {
     doSomething();
   } catch (IOException e) {
-    logger.warn("Explain where we are in the code", e);  
+    logger.warn("Explain what went wrong and how to fix it. You can have arguments {}.", 12, e);
```
openhab-5iver: > 12

IMO, this would look better as a variable name like `someVariable`.

openhab-5iver:
Looks good!

PR #940, opened 2019-04-16
Delete conventions.md

davidgraeff:
This has been incorporated into guidelines with the last merge.


Confectrician:
@ysc can you give some information about the failing preview build?
I would guess that it is related to the fact that this pr contains only a deletion, but i am not sure.

ghys:
> @ysc can you give some information about the failing preview build?
>I would guess that it is related to the fact that this pr contains only a deletion, but i am not sure.

Yep, it's pretty unforgiving with non-existing pages on the sidebar.
What bugs me is that Netlify now "shuts down logging" with pending messages so we can't see the actual exception, it's unfortunate, I'll have to investigate.
Note that the same must also still be done in the website repo - done in https://github.com/openhab/website/pull/175.

PR #949, opened 2019-04-25
Sitemaps: Fix links

mueller-ma:
And change "openHAB Android App" to "openHAB App for Android"

> "Android", or anything confusingly similar to "Android", cannot be used in names of applications or accessory products, including phones, tablets, TVs, speakers, headphones, watches, and other devices. Instead, use "for Android".

from https://developer.android.com/distribute/marketing-tools/brand-guidelines?hl=en


Confectrician:
Hey @mueller-ma and thanks for the improvement.
Could you fix the dco check error?

PR #962, opened 2019-05-04
Update subtitles in linux.md to make them more visible

johannesE:
That was kinda annoying for me.


Confectrician:
Thanks for the catch and the fix. Looks very well this way.

Issue #955, opened 2019-04-28
Developer section: New buildsystem IDE setup


davidgraeff:
We have no working IDE setup at the moment and therefore no complete guide in the developer section on how to setup a proper build environment.

As soon as the bnd migration is finished and the openHAB distro is fully bnd centric this section need to be (re)written.

I propose to guide in a pictureful way through setting up:

* Eclipse
* Intellij
* Visual Studio Code

mvalla:
working on a very first version of the Eclipse section with images

mvalla:
completed a first version of the guide for the Eclipse IDE see #968

tmrobert8:
FYI - first version on the VSCode instructions:

https://github.com/openhab/openhab-docs/pull/994

mvalla:
new version of the Eclipse IDE instructions using Eclipse Installer & openhab-distro 
(not depending on maggu2810/openhab-demo anymore).

#995

Jochen1980:
The Link zu OpenJDK installation brings you to Oracle and I guess the most preferred way is to install Zulu? 
see https://www.openhab.org/v2.4/docs/developer/development/ide.html

zephod77:
> completed a first version of the guide for the Eclipse IDE see #968

Can I add a request that there be an Eclipse IDE setup guide for when you are not installing Eclipse from scratch? In other words assume that a user already has a Java development environment set up and they just want to import the openHAB sources.

kaikreuzer:
@zephod77 I'd think this would easily get complicated. The whole idea of the Oomph setup is to simplify and homogenize the IDE setup and by this, simplify the support.

zephod77:
@kaikreuzer My issue is that I have several other projects besides openHAB and so I use a C++ development plugin, a Python development plugin and a modeling plugin to name a few as well as the Java development plugin. In fact, before this reorganization started, I was using the modeling tool to generate the Java code for a binding. To have to reload and reconfigure those each time is not a viable solution for me. 

It might well be very complicated but the requirements to a "working" system must be known and so it should be possible.

openhab-5iver:
> To have to reload and reconfigure those each time is not a viable solution for me.

Why couldn't you use a separate Eclipse installation for OH?

zephod77:
[...]

PR #968, opened 2019-05-05
first version for Eclipse IDE instructions

mvalla:
Signed-off-by: M Valla <mvbruincc@gmail.com>

See #955. I wrote a first version of the guide with Eclipse IDE. It works for me.
I also added a section to start writing a new binding using skeleton (which fails when adding the newly crated binding to features).
I did not find a better way to add the newly created binding to the demo.app other than recompiling openhab2-addons, probably there is a better way.


Confectrician:
Hey @ghy, i am still a netlify rookie.
Can you give me a hint why the preview is failing here?
I would guess that the uploaded images are a problem for netlify preview.
Is that possible?

Confectrician:
>Do not why it does not look as signed-off: I did all last commits with --signoff opt,

There's one commit missing, but this is no problem.
I can override the DCO Check, isnce there are signed off commits at all. 
image

About the build error:
I think this could be related due to the fact, that those images aren't existing in the repo yet.

Confectrician:
Looks indeed like its just the pictures missing. (Which is fully ok since they are getting uploaded after this PR for the first time.)

image

Confectrician:
@davidgraeff may i ask you to have a short look at the changed files?

I can find nothing that would prevent this pr from being merged.
I think the build issue will solve itself when the images are in the repo physically.

davidgraeff:
The preview must succeed. Has the `https://github.com/openhab/openhab-docs/blob/master/.vuepress/process_main_docs.rb` file being changed to accomodate for the new pictures? I really don't like the current buildsystem in here tbh. Adding files should be enough.

I have read the changes already some days ago but didn't comment yet. The guide is now more detailed indeed, but at the same time Kai is working on making the setup file a valid option again. That might be easier for users. And we also don't use Markus demo repository anymore since like 2 days or so the demo app has been added to openhab-distro. So all in all it's a moving target.

We can of course merge this one, as it is an improvement. But we'll need to adapt sooner or later again.

Confectrician:
>I really don't like the current buildsystem in here tbh. Adding files should be enough.

This is a problem for sure.
We can't expect new contributors to know this.
I haven't thought about this yet, until you pointed to it.

Confectrician:
Here we go,

Thanks for the prepare script hint @davidgraeff.
https://5cd3424c7ff5e5000876fa3e--openhab-docs-preview.netlify.com/docs/developer/

Looks good from my side.

>We can of course merge this one, as it is an improvement. But we'll need to adapt sooner or later again.

This is a problem that will escort us every time.
Hopefully other IDEs will supplement the eclipse instructions in the future too.

davidgraeff:
They will. I have worked on a Visual Studio Code workflow. That requires some ehm... changes that the OH2-addon guys need to be convinced of first.

Confectrician:
Thanks for the effort @mvalla.

As you may see we have to improve out preview a bit.
Sorry for the delay caused to this.

>I have worked on a Visual Studio Code workflow.

I am already waiting for it. 🙂

mvalla:
why the merged version is not yet online here: https://www.openhab.org/docs/developer/  ?
people out there are struggling to have an IDE working and this one guide at least seems to work

Confectrician:
Please follow our issues.
The website build is broken currently and we are working on a solution.

PR #972, opened 2019-05-08
Update guidelines.md

jossuar:
Corrected link to ESH.xml

Signed-off-by: Georgios Moutsos <georgios.moutsos@gmail.com>

### DIFF
--- a/developers/guidelines.md
+++ b/developers/guidelines.md
@@ -45,7 +45,7 @@ The following directory and file layout must be respected:
 1. Every Java file must have a license header. You can run ```mvn license:format``` on the root of the repo to automatically add missing headers.
 1. Code must be formatted using the "ESH" code formatter (in Eclipse).
    - This is set up automatically by the official [IDE setup](ide.html)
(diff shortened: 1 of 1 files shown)


jossuar:
I'm not sure if the changed link to download the ESH.xml file works correctly.

Confectrician:
It doesn't work.

@ghys 
It looks like vuepress can't or won't handle xml files or at least this special one.
I can't acces it from this preview or the live build, alltough it should be there since it is definetely in the repo.

Can you have a look?

Confectrician:
I have added the direct raw link to github for now until we have a long term solution or i have found the reason for the missing xml file processing.

When the preview works i will merge this one.

ghys:
> It doesn't work.
> 
> @ghys
> It looks like vuepress can't or won't handle xml files or at least this special one.
> I can't acces it from this preview or the live build, alltough it should be there since it is definetely in the repo.
> 
> Can you have a look?

Sorry @Confectrician - I was in holidays and missed that!

Confectrician:
No problem.
I thought it was important to have a working solution meanwhile.

We can revert that at an time, when the problem is addressed.

PR #978, opened 2019-05-14
Fix Quartz 404 cron links

freak3dot:
Cron link for quartz scheduler has been moved to a new url without a redirect. Suggested fix is to update OpenHAB link.


Confectrician:
Thanks for the fix. 👍

PR #979, opened 2019-05-16
Change description of System started description

rkoshak:
Based on the discussion at https://community.openhab.org/t/systemstart-rule-trigger-if-i-change-any-item/74542 and after performing a few tests of my own including creating a brand new .items file with an Item not used by ANY Rule I have verified that all System started Rules trigger for any change to any .items file. The "related to that rule file" part of the description is misleading and incorrect and should be removed.

Sign off by Richard Koshak: rlkoshak@gmail.com

### DIFF
--- a/configuration/rules-dsl.md
+++ b/configuration/rules-dsl.md
@@ -183,7 +183,7 @@ Two system-based triggers are provided as described in the table below:
 
 | Trigger           | Description                                                                                                                                                                                        |
 |-------------------|----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
-| System started    | System started is triggered upon openHAB startup, after the rule file containing the System started trigger is modified, or after item(s) related to that rule file are modified in a .items file. |
+| System started    | System started is triggered upon openHAB startup, after the rule file containing the System started trigger is modified, or after item(s) are modified in a .items file. |
 | System shuts down | Rules using the 'System shuts down' trigger execute when openHAB shuts down.                                                                                                                       |
 
 You may wish to use the 'System started' trigger to initialize values at startup if they are not already set.


rkoshak:
So is just adding "signed off by blah blah blah" no longer sufficient to sign off on a PR? I didn't see an option to signoff in the in browser editor when I submitted the PR. Does this mean quick and easy edits like this can no longer be done in the browser?

Please advise.

Thanks!

openhab-5iver:
Just add the signoff by into the commit message

rkoshak:
I thought I did.

I just changed it from "Signed off" to "Sign off". 

This is the first time I've seen that DCO error and the solution is to amend the PR and commit using the -signoff flag to git, but that necessitates working from a local git repo and I couldn't find a way to do that through the browser. 

I don't particularly want to clone the repo and go through all of that to make a six word change to the docs.

Can I safely ignore that DCO error since my commit message includes the sign off?

openhab-5iver:
Looks to me that this would fall under the small change exception, so I think you are good.

rkoshak:
But if I were to make a major change, I'd need to use --signoff in my git commit message to avoid that DCO error, correct? Just adding "Signed off by Richard Koshak: rlkoshak@gmail.com" appears to no longer be sufficient, at least as far as DCO is concerned.

openhab-5iver:
If you submit from the website, add the signoff to the commit message too. I don't see that on this commit.

rkoshak:
I know I wrote it but it clearly isn't there. I must have messed something up when I copied and pasted.

Confectrician:
You commit message seems very long and it was maybe cut.

The commit message isn't the right place for long explanations.
There is more than enough space for additional comments in the pr description.

DCO check:

I think the sign off has to be in the last line of a commit and it has to be in the correct form.

`Signed-off-by: Jerome Luckenbach <github@luckenba.ch>`

1. correct wording
2. no github handle or anything additional than name and email

Anyway i can override this check, if i think we have a small patch or the sign off is done in the description so i have the needed data which i can copy over in the merge commit.

Of course it would be nice to have a green checkmark directly.

Confectrician:
Thanks for the improvement anyway. 👍 

I have copied over your sign off statement in a correct form in the commit.
Have a look at it for reference.

BR
Jerome

PR #982, opened 2019-05-17
Corrected link to ESH.xml

openhab-5iver:
Signed-off-by: Scott Rushworth openhab@5iver.com


openhab-5iver:
I think the signoff needs a blank line above it, but I'm not going to take the time to amend the commit (sorry!).

Confectrician:
>I think the signoff needs a blank line above it

I think you have forgotten the `<>` brackets, which lets dco fail.
But no need to amend anything. I will override the check and correct this.

Confectrician:
Copy of #972

Confectrician:
And it doesn't work like #972 also doesn't work.

@ghys can you have a look?
Is it possible that `xml` files don't get processed?

I think i had already a look into the settings and faced this as reason.
But i am not sure if we should change the script or better put a full url link to the raw version on github.

Confectrician:
I have added a temporary solution through #972 and created #992 to get a stable one on long term.

PR #995, opened 2019-05-20
new instructions using Eclipse Installer and openhab-distro

mvalla:
Signed-off-by: M Valla <mvbruincc@gmail.com>


developers/ide/eclipse.md
```diff
@@ -14,38 +14,53 @@ For now follow these steps:
...
+1. Download the "Eclipse Installer": [can be downloaded from the Eclipse web site](https://wiki.eclipse.org/Eclipse_Installer)
+1. Launch the Installer and select `Advanced Mode...` on the menu (3 bars) on the right, then select `Eclipse IDE for Java Developers` then `Next>`
+1. Select `Github Projects > openHAB > openHAB Development`. 
+    
+    Do *NOT* select other projects for now.
 
-   Restart Eclipse to activate Bndtools.
-1. Using `git` from the command line, checkout the openHAB demo application: `git clone --depth=1 https://github.com/maggu2810/openhab-demo`
-1. Import the demo application from Eclipse: File -> Import -> Maven -> Existing Maven Projects -> select the root directory of the demo application project -> Finish.
+   ![select projects](images/ide_setup_eclipse_projects.png)
```
ghys: This image is missing :)
(hence the build failure!)
Please also use make relative URLs like this: `./images/ide_setup_eclipse_projects.png`

Gozilla01:
Hi @massi,
which version should you choose 03-2019, laster 03-2019 or laster 06-2019?
There is still a lot of confusion about the versions, I also know that the maintainers don't want to specify the version in the documentation. I think that in this case, given the difficulties that exist, is it not better to specify the version to use?

kaikreuzer:
> @kaikreuzer
Which version is correct to use?

We should probably go for the latest released version, which should be the default selection.

Gozilla01:
I ran tests with the last 06-2019, does not support bndtools.
This problem was also reported by @massi in the new documentation
https://www.openhab.org/docs/developer/ide/eclipse.html#eclipse-ide-setup

PR #1001, opened 2019-05-29
Minor edits to Thing.md.

ezeeman:
Changed some non-idiomatic phrases, corrected minor grammatical issues, and replaced usages of "state" with "status" to be consistent.

Signed-off-by: Eric Weitzman <eweitzman@acm.org>


ezeeman:
Will redo from local machine instead of using the github web interface to fix DCO validation failure.

PR #1014, opened 2019-07-02
Basic UI Visibility Live Update Works

rkoshak:
I just tested this and BasicUI visibility does work. The warning should be removed.

Signed off by: Richard Koshak rlkoshak@gmail.com


rkoshak:
Sorry, I forgot to add the signed off to the commit message.

PR #1022, opened 2019-08-03
Update migration.md

mstormi:
added hint to use Karaf console commands to simulate events in rules

### DIFF
--- a/tutorials/migration.md
+++ b/tutorials/migration.md
@@ -398,7 +398,12 @@ cp $OPENHAB1_CONF/configurations/sitemaps/* $OPENHAB_CONF/sitemaps/*
   Add 'legend=true' to restore the chart's legend. 
 
 ### Rules
-Test your rules using commands in Karaf console. Use smarthome:send to send a command, smarthome:update to update an Item's state, and smarthome:status to get an Item's state. Use smarthome --help for more details. These can be used to simulate events that can trigger a rule, even in an installation without "active" devices.
+
+Test your rules using commands in Karaf console.
+Use smarthome:send to send a command, smarthome:update to update an Item's state, and smarthome:status to get an Item's state.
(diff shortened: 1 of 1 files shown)


tutorials/migration.md
```diff
@@ -398,6 +398,7 @@ cp $OPENHAB1_CONF/configurations/sitemaps/* $OPENHAB_CONF/sitemaps/*
   Add 'legend=true' to restore the chart's legend. 
 
 ### Rules
+  * test your rules using commands in Karaf console (`smarthome:status`, `smarthome:send/smarthome:update` to simulate `sendCommand()/updates`). You can even do this on a OH2 install that has no "active" devices while you still run OH1.
```
openhab-5iver: Test your rules using commands in the Karaf console (`smarthome:status` to get an Item's state, `smarthome:send` to send a command, `smarthome:update` to update an Item's state). Use `smarthome --help` for more details.

Capital T. The commands in Karaf are not simulating anything... Karaf is an actual command line interface and the commands are real :wink:. The second sentence is confusing, so I'd leave it out.

mstormi: Well seems we should reword it but I don't want to leave that sentence out.
My main point is you can test your rules' OH2 compatibility without running them on real devices, actually you can do this while still running your 1.x install so the critical phase to find & fix issues when running 2.X with your HW can be shortened.

openhab-5iver: The second sentence and statement about simulation totally make sense after realizing which document this is! How about this...

Test your rules using commands in the Karaf console. Use  `smarthome:send` to send a command, `smarthome:update` to update an Item's state, and `smarthome:status` to get an Item's state. Use `smarthome --help` for more details. These can be used to simulate events that can trigger a rule, even in an installation without "active" devices.

PR #1032, opened 2019-08-17
Correct Apt-based stable repository url

TRSx80:
I am not sure if this applies also to the Yum/Dnf based links for stable or not, as I didn't investigate those.

Also, it's pretty late and I gave a quick look over the rest of the page, but nothing else jumped out. I figure an imperfect PR is better than none at all, even if I was almost too tired to do one.

This also closes https://github.com/openhab/openhab-linuxpkg/issues/150 but I didn't reference it, because it's actually in a different repository.

### DIFF
--- a/installation/linux.md
+++ b/installation/linux.md
@@ -82,7 +82,7 @@ Then, you can choose between, *Official (Stable)*, *Beta* or *Snapshot* builds:
     Add the **openHAB 2 Stable Repository** to your systems apt sources list:
 
     ```shell
-    echo 'deb https://dl.bintray.com/openhab/apt-repo2 stable main' | sudo tee /etc/apt/sources.list.d/openhab2.list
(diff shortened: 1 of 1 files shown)


TRSx80:
Also, [Contribution Guidelines](https://github.com/openhab/openhab-docs/blob/master/CONTRIBUTING.md) state that I don't need to sign off if it's just a one line change to the docs.

Confectrician:
Hi,

This seems to be a duplicate of #1031.

Confectrician:
>Also, Contribution Guidelines state that I don't need to sign off if it's just a one line change to the docs.

The DCO bot doesn't know about this exclusion and i don't know if it even would be able to check things like this.
Usually i am overruling the bot, when i see that a sign-off isn't needed. 🙂

TRSx80:
@Confectrician,

Aah yes, seems that you are correct. Sorry for the noise (closing)!

Confectrician:
No noise recognised from my side. 🙂 
Your're welcome.

PR #1033, opened 2019-08-17
Corrected bad name to use hyphen (which is not allowed)

WardMD90745:
Bad image filename was correct (with underscore) - corrected bad name to use hyphen (which is not allowed)

### DIFF
--- a/configuration/items.md
+++ b/configuration/items.md
@@ -322,7 +322,7 @@ The following guidelines apply to user-added icon files:
 - Hyphens (`-`) are reserved for [Dynamic Icons](#dynamic-icons) (see below)
 - Example filenames:
   - Good: `myswitch.svg`, `power_meter.png`, `error23.svg`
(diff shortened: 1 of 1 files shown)


ThomDietrich:
Hey guys. The original document was by me and this change doesn't seem meaningful. The example was meant to show that uppercase letters are prohibited. The change to hyphen doesn't make sense as the second example highlights this. My suggestion would be "MySwitch.svg" or similar. @WardMD90745 wdyt? It would be nice if you'd create another PR

Confectrician:
Maybe we should then think about a general change in structure here.
We should add a good/bad example for each explanation.

I think it is currently difficult to make a connection between the explanation and the examples.

ThomDietrich:
Not sure.
I for one seldom really "read" documentation (I know, weird coming from me) but rather orient myself on examples. THEN if something is not clear or becomes more important I would actually start reading the fine print. That's why I *do not dislike* the current structure.

PR #1036, opened 2019-08-21
Update Unit of Measurements with new Units from openhab-core

Cossey:
@cweitkamp requested that I update the documentation to reflect the new Unit of Measurements currently available in openHAB core.

_Someone may want to check the stylizing of some of the unit names._

**Note: I have a Pull Request open for openhab/openhab-core#990 which corrects two unit symbols I found when documenting the new Units.**

Signed-off-by: Stewart Cossey <stewart.cossey@gmail.com>

### DIFF
--- a/concepts/units-of-measurement.md
+++ b/concepts/units-of-measurement.md
@@ -97,63 +97,93 @@ SI:
 
 | Type                   | Unit                    | Symbol |
 |------------------------|-------------------------|--------|
-| Acceleration           | Metre per square Second | m/s²   |
-| AmountOfSubstance      | Mole                    | mol    |
-| Angle                  | Radian                  | rad    |
-| Angle                  | Degree                  | °      |
-| Angle                  | Minute Angle            | '      |
-| Angle                  | Second Angle            | ''     |
-| Area                   | Square Metre            | m²     |
-| ArealDensity           | Dobson Unit             | DU     |
-| CatalyticActivity      | Katal                   | kat    |
-| Dimensionless          | Percent                 | %      |
-| Dimensionless          | Parts per Million       | ppm    |
-| Dimensionless          | Decibel                 | dB     |
-| ElectricPotential      | Volt                    | V      |
-| ElectricCapacitance    | Farad                   | F      |
-| ElectricCharge         | Coulomb                 | C      |
-| ElectricConductance    | Siemens                 | S      |
-| ElectricCurrent        | Ampere                  | A      |
-| ElectricInductance     | Henry                   | H      |
-| ElectricResistance     | Ohm                     | Ω      |
-| Energy                 | Joule                   | J      |
-| Energy                 | Watt Second             | Ws     |
-| Energy                 | Watt Hour               | Wh     |
-| Energy                 | KiloWatt Hour           | kWh    |
-| Force                  | Newton                  | N      |
-| Frequency              | Hertz                   | Hz     |
-| Illuminance            | Lux                     | lx     |
-| Intensity              | Irradiance              | W/m²   |
-| Length                 | Metre                   | m      |
-| Length                 | Kilometre               | km     |
-| LuminousFlux           | Lumen                   | lm     |
-| LuminousIntensity      | Candela                 | cd     |
-| MagneticFlux           | Weber                   | Wb     |
-| MagneticFluxDensity    | Tesla                   | T      |
-| Mass                   | Kilogram                | kg     |
-| Mass                   | Gram                    | g      |
-| Power                  | Watt                    | W      |
-| Pressure               | Pascal                  | Pa     |
-| Pressure               | hectoPascal             | hPa    |
-| Pressure               | Millimetre of Mercury   | mmHg   |
-| Pressure               | Bar                     | bar    |
-| Pressure               | milliBar                | mbar   |
-| Radioactivity          | Becquerel               | Bq     |
-| RadiationDoseAbsorbed  | Gray                    | Gy     |
-| RadiationDoseEffective | Sievert                 | Sv     |
-| SolidAngle             | Steradian               | sr     |
-| Speed                  | Metre per Second        | m/s    |
-| Speed                  | Kilometre per Hour      | km/h   |
-| Speed                  | Knot                    | kn     |
-| Temperature            | Kelvin                  | K      |
-| Temperature            | Celsius                 | °C     |
-| Time                   | Second                  | s      |
-| Time                   | Minute                  | min    |
-| Time                   | Hour                    | h      |
-| Time                   | Day                     | d      |
-| Time                   | Week                    | week   |
-| Time                   | Year                    | y      |
-| Volume                 | Cubic Metre             | m³     |
+| Acceleration           | Metre per square Second          | m/s²   |
+| Acceleration           | Standard Gravity                 | ɡₙ     |
+| AmountOfSubstance      | Mole                             | mol    |
+| AmountOfSubstance      | Deutscher Härtegrad              | °dH    |
+| Angle                  | Radian                           | rad    |
+| Angle                  | Degree                           | °      |
+| Angle                  | Minute Angle                     | '      |
+| Angle                  | Second Angle                     | ''     |
+| Area                   | Square Metre                     | m²     |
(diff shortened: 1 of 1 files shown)


concepts/units-of-measurement.md
```diff
@@ -97,63 +97,93 @@ SI:
...
+| ElectricCapacitance    | Farad                            | F      |
+| ElectricCharge         | Coulomb                          | C      |
+| ElectricConductance    | Siemens                          | S      |
+| ElectricCurrent        | Ampere                           | A      |
+| ElectricInductance     | Henry                            | H      |
+| ElectricResistance     | Ohm                              | Ω      |
+| Energy                 | Joule                            | J      |
+| Energy                 | Watt Second                      | Ws     |
+| Energy                 | Watt Hour                        | Wh     |
+| Energy                 | KiloWatt Hour                    | kWh    |
```
cweitkamp:
```suggestion
| Energy                 | Kilowatt Hour                    | kWh    |
```

concepts/units-of-measurement.md
```diff
@@ -97,63 +97,93 @@ SI:
...
+| Energy                 | Megawatt Hour                    | MWh    |
+| Energy                 | Kilovar Hour                     | kvarh  |
+| Force                  | Newton                           | N      |
+| Frequency              | Hertz                            | Hz     |
+| Illuminance            | Lux                              | lx     |
+| Intensity              | Irradiance                       | W/m²   |
+| Intensity              | microWatt per square centimeter  | µW/cm² |
```
cweitkamp:
```suggestion
| Intensity              | Microwatt per square Centimeter  | µW/cm² |
```

concepts/units-of-measurement.md
```diff
@@ -97,63 +97,93 @@ SI:
...
+| Length                 | Kilometre                        | km     |
+| LuminousFlux           | Lumen                            | lm     |
+| LuminousIntensity      | Candela                          | cd     |
+| MagneticFlux           | Weber                            | Wb     |
+| MagneticFluxDensity    | Tesla                            | T      |
+| Mass                   | Kilogram                         | kg     |
+| Mass                   | Gram                             | g      |
+| Power                  | Watt                             | W      |
+| Power                  | Kilovar                          | kvar   |
+| Power                  | Decibel-milliwatts               | dBm    |
```
cweitkamp:
```suggestion
| Power                  | Decibel-Milliwatts               | dBm    |
```

concepts/units-of-measurement.md
```diff
@@ -97,63 +97,93 @@ SI:
...
+| Pressure               | Pascal                           | Pa     |
+| Pressure               | hectoPascal                      | hPa    |
```
cweitkamp:
```suggestion
| Pressure               | Hectopascal                      | hPa    |
```

concepts/units-of-measurement.md
```diff
@@ -97,63 +97,93 @@ SI:
...
+| Pressure               | Millimetre of Mercury            | mmHg   |
+| Pressure               | Bar                              | bar    |
+| Pressure               | milliBar                         | mbar   |
```
cweitkamp:
```suggestion
| Pressure               | Millibar                         | mbar   |
```

Cossey:
Will correct the styling of the items you’ve pointed out. Regarding Binary prefixes, I’ll add this to the prefixes table. 

_The standard gravity unit symbol (the g with the subscript style “n”) doesn’t appear to render correctly on Safari iOS when looking at the preview URL on my phone. This was copied straight from the source code. I wonder if there might be issues with this particular character on other platforms._

Cossey:
Unit styling all fixed up now. I ended up creating another table specifically for the Binary Prefixes since they reside in a different class. Also changed the Column Value to Factor for the Prefix tables and also mentioned the BinaryPrefixes classes. Also some slight formatting changes and another example showing a Binary Prefix.

*Used this to double-check the changes: https://physics.nist.gov/cuu/Units/binary.html*

ThomDietrich:
Can confirm the subscript n issue. It is neither rendered on GitHub nor in the netlify preview on macos and Firefox.

image

I'm a big fan of unicode symbols but using this one doesn't seem to be a good idea :-/ I propose to go with `g<sub>n</sub>` g<sub>n</sub>

Cossey:
Does anyone else have any other proposals regarding this character? Should this character also be changed in `openHAB-core` as one would assume rendering issues in PaperUI due to the particular Unicode character used?

Cossey:
@cweitkamp This seems stuck until I can get some feedback from others; may also need to open a PR for changing this UoM symbol in openhab-core depending on what we go with.

Happy to wait for further feedback from other relevant parties.

cweitkamp:
Maybe we can ask @Hilbrand for his opinion as he introduced standard gravity in https://github.com/openhab/openhab-core/pull/617. It looks like we are using a different (Unicode) character in OHC. I would suggest to add a unit test over there to verify if it is really working before we change it.

Hilbrand:
This is what the page shows when I open it in Firefox:

image

However. I noticed the `g` seems to be different from what is in openHAB core:

where  | unicode | visual
-- | -- | --
core | g\&#x2099;  | g&#x2099;
this pr |  \&#x0261;\&#x2099; | &#x0261;&#x2099;

Using `<sub>` assumes the context is always HTML. I don't know if that is true.

Cossey:
Hmm okay let me look into this in the next few days so we can get this sorted and issue closed.

Cossey:
> This is what the page shows when I open it in Firefox:
> 
> image
> 
> However. I noticed the `g` seems to be different from what is in openHAB core:
> 
> where	unicode	visual
> core	g&#x2099;	gₙ
> this pr	&#x0261;&#x2099;	ɡₙ
> Using `<sub>` assumes the context is always HTML. I don't know if that is true.

Looking at this Unicode 0261 is the lowercase g so the subscript `n` in question is still the same. Still unsure how to proceed since it appears that many fonts don't display this character.

Maybe we have to settle to using a standard `n` character?

cweitkamp:
Just a stupid guess: What about using an image?

Cossey:
@cweitkamp I think until more system fonts support this unicode character your suggestion is the best way forward and I will do this. 
However how would core handle this as this would be a problem in Paper UI and classic UI too? Should I open an issue there for discussion or just leave it until someone raises a problem with it? AFAIK no one uses this UoM yet.

cweitkamp:
@Cossey I will test it.

cweitkamp:
I am late ...

channel type:
```xml
       <channel-type id="gravity">
               <item-type>Number:Acceleration</item-type>
               <label>Gravity</label>
               <description>Gravity.</description>
               <state readOnly="true" pattern="%.1f %unit%" />
       </channel-type>
```

Java code:
```java
        updateState(channelUID, new QuantityType<>(1, SmartHomeUnits.STANDARD_GRAVITY));
```

Paper UI:

grafik

Basic UI:

grafik

concepts/units-of-measurement.md
```diff
@@ -95,95 +97,135 @@ Imperial:
...
+| Pressure               | Hectopascal                      | hPa    |
+| Pressure               | Millimetre of Mercury            | mmHg   |
+| Pressure               | Bar                              | bar    |
+| Pressure               | Millibar                         | mbar   |
+| Radioactivity          | Becquerel                        | Bq     |
+| RadiationDoseAbsorbed  | Gray                             | Gy     |
+| RadiationDoseEffective | Sievert                          | Sv     |
+| SolidAngle             | Steradian                        | sr     |
+| Speed                  | Metre per Second                 | m/s    |
+| Speed                  | Kilometre per Hour               | km/h   |
```
cweitkamp: Can we add `mm/h` (see https://github.com/openhab/openhab-core/issues/1297)?

Confectrician:
I am just waiting for a "go" or "pr is ready now" :)

Cossey:
Okay let me deal with #1297 and then it should be good to go.

concepts/units-of-measurement.md
Cossey: Done.

Cossey:
Sorry for the outstanding PR - if it looks good to yous then LGTM?

PR #1065, opened 2019-10-23
[WIP]Minor updates collection

Confectrician:
This is meant for doing several minor updates to the docs.

### DIFF
--- a/configuration/jsr223.md
+++ b/configuration/jsr223.md
@@ -31,6 +31,90 @@ New APIs are planned to be implemented in the future, which will provide standar
 - [Javascript](jsr223-js.html)
 - Groovy (TODO)
 
+#### Example rules for a first impression
+
+:::: tabs
+
+::: tab JavaScript
+```js
+'use strict';
+
+scriptExtension.importPreset("RuleSupport");
+scriptExtension.importPreset("RuleSimple");
+
+var sRule = new SimpleRule() {
+    execute: function( module, input) {
+        print("This is a 'hello world!' from a Javascript rule.");
+    }
+};
+
+sRule.setTriggers([
+    TriggerBuilder.create()
+        .withId("aTimerTrigger")
+        .withTypeUID("timer.GenericCronTrigger")
+        .withConfiguration(
+            new Configuration({
+                "cronExpression": "0 * * * * ?"
+            })).build()
+    ]);
+
+automationManager.addRule(sRule);
+```
+:::
+
+
+::: tab Jython
+```python
+scriptExtension.importPreset("RuleSupport")
+scriptExtension.importPreset("RuleSimple")
+
+class myRule(SimpleRule):
+    def execute(self, module, inputs):
+        print "This is a 'hello world!' from Jython rule."
+
+sRule = myRule()
+sRule.setTriggers([
+    TriggerBuilder.create()
+        .withId("aTimerTrigger")
+        .withTypeUID("timer.GenericCronTrigger")
+        .withConfiguration(
+            Configuration({
+                "cronExpression": "0 * * * * ?"
+            })).build()
+    ])
+
+automationManager.addRule(sRule)
+```
+:::
+
+::: tab Groovy
+```groovy
+import org.eclipse.smarthome.automation.*
+import org.eclipse.smarthome.automation.module.script.rulesupport.shared.simple.*
+import org.eclipse.smarthome.config.core.Configuration
+
+scriptExtension.importPreset("RuleSupport")
+scriptExtension.importPreset("RuleSimple")
+
+def sRule = new SimpleRule() {
+    Object execute(Action module, Map<String, ?> inputs) {
+        println("Hello World from Groovy")
+    }
+}
+sRule.setTriggers([
+    TriggerBuilder.create()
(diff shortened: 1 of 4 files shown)


Confectrician:
Feedback appreciated @mstormi 

image

I think we should use this banners more in such cases in the future.

Confectrician:
Hey @openhab-5iver,

I have added the native jsr223 exmaples to the jsr223 page to test the tabs, that are included meanwhile.
Would love to get some feedback, alltough we already knew that it should be an imporvement. :)

configuration/rules-ng.md
```diff
@@ -11,7 +11,9 @@ title: Next-Gen Rules
 
 Since openHAB 2.4 another Rule Engine has been added. It works fundamentally different than what you find with our current [Rules](https://www.openhab.org/docs/configuration/rules-dsl.html). It allows Rules to be edited in a graphical fashion and to interact with [JSR223 Scripts (Javascript, Jypthon, etc)](https://www.openhab.org/docs/configuration/jsr223.html).
 
-> Note: If you use it, this is at your own risk. Especially, you should not expect any documentation or community experience besides what you find on this page.
+::: tip Note
+If you use it, this is at your own risk. Especially, you should not expect any documentation or community experience besides what you find on this page.
```
openhab-5iver: IMO, this can be removed. There are plenty of topics in the forum and there is a lot of documentation in the helper library repo.

configuration/jsr223.md
```diff
@@ -31,6 +31,90 @@ New APIs are planned to be implemented in the future, which will provide standar
...
+                "cronExpression": "0 * * * * ?"
+            })).build()
+    ]);
+
+automationManager.addRule(sRule);
+```
+:::
+
+
+::: tab Jython
```
openhab-5iver: Jython should be first

openhab-5iver:
For some reason the netlify preview is not working... would love to see the tabs!

I think that a /configuration/rules.html pages should be put in to give a quick comparison of the rule engines. This could unbury rules-ng.html from under Interfaces and Ecosystems.

configuration/rules-ng.md
Confectrician: Sounds fair. Will do so later.

Confectrician:
Preview is working fine for me. 🤷‍♂ 

grafik

From
https://deploy-preview-1065--openhab-docs-preview.netlify.com/docs/configuration/jsr223.html

configuration/jsr223.md
Confectrician: The question is why? :D
From my point groovy should be first, if we go for an alphabetical order.
I think we need to define some blueprint for this in general for the future.

openhab-5iver:
Oh... OK. I didn't realize you needed to add the rest. The content of the scripts should be taken from the helper library examples... https://openhab-scripters.github.io/openhab-helper-libraries/Guides/Rules.html#raw-api, but expanded for multiple triggers like the Python example. This gets into the OH doc rewrite I'd done before the Helper Library doc rewrite. A lot of the HL docs can be migrated to the OH docs.

configuration/jsr223.md
openhab-5iver: Sorry, I thought this was common knowledge, mainly due to Kai's comment about making Jython a first class citizen ([first link](https://community.openhab.org/search?context=topic&context_id=83943&q=Jython%20first%20class%20citizen&skip_context=true)). It is also the most used and the helper libraries are the most advanced. I'm familiar with both JavaScript and Python, and in comparison, JavaScript is definitely the least friendly for writing automation. The HL docs would need a lot of updates if the order needs to change! The tab should be Python vs Jython, since that is the programming language used. It will also migrated easier when Graal comes in. This also gives a reverse alpha all the way down to DSL Rules. :smile: jRuby is almost ready to be added and I've had a look into Kotlin too, so we should have examples for more languages soon!

rkoshak: Is the plan for the Helper Libraries to become part of OH (minus the Community submissions perhaps) or are they to remain forever separate? I ask because the difference between these examples and especially the Python example using the decorators is pretty big. And the decorators make the Python look a lot more like Rules DSL in structure which will aide users who want to migrate.

openhab-5iver: The plan for the core helper libraries is obsolescence... except for any functionality that cannot fit into the Scripting API. These examples are the raw API, which IMO people should be educated on. Once the Scripting API is available, everything simplifies and pulls back into OH.

Confectrician:
> Oh... OK. I didn't realize you needed to add the rest.

The preview builds the complete documentation repository as website. :)

configuration/jsr223.md
Confectrician: Hm, this is going in the wrong direction.
My main intention here was to add a TAB example and i thought jsr would be a good starter for this.
But i don't want to discuss this in general now.

So i should maybe remove this tab example.
There is still much unclear in how we should and will document jsr223.

openhab-5iver: IMO it looks good and provides useful information. The actual content can be cleaned up later when the nonHL docs are moved from the helper library docs to the OH docs. The tabs are great!

PR #1090, opened 2019-11-25
blocks are not supported in switch

kewld:
blocks or subpanel is only supported in group and text but not switch.


Confectrician:
Hey @kewld,

I have tested that and you are right, so the improvement was correct.
Any special reason for closing this pull request?

BR
Jerome

PR #1091, opened 2019-11-25
Aded some content to test this label bot

Confectrician:

### DIFF
--- a/addons/index.md
+++ b/addons/index.md
@@ -4,7 +4,7 @@ layout: documentation
 
 {% include base.html %}
 
-# Add-ons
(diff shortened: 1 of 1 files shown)


Confectrician:
@mueller-ma it seems that my colon escape didn't work.
Did you any furhter investigation already on how to debug this?

Probably it would be the easiest virant to rename the labels over here.

mueller-ma:
I don't think that you have to escape them when using quotes.
One user suggested to use `&colon;`: https://stackoverflow.com/a/14873305

> Probably it would be the easiest to rename the labels over here.

I think so.

PR #1092, opened 2019-11-26
Add JsonPath icon

mueller-ma:
and remove some unneeded x-bit

Signed-off-by: mueller-ma <mueller-ma@users.noreply.github.com>


mueller-ma:
Do I need to set the icon somewhere?

ghys:
No, it should be fine, thanks!

mueller-ma:
The icon doesn't show up on https://www.openhab.org/addons/ yet.

PR #1114, opened 2019-12-21
Changed related to rename of openHAB2-addons

Hilbrand:
Also remove some legacy development files and fixed links to ide documentation in other pages.


Confectrician:
We have to do the same changes in https://github.com/openhab/website, after review of this PR.

The vuepress related files are just here for the preview part.

PR #1123, opened 2019-12-31
Update units-of-measurement.md

clinique:


concepts/units-of-measurement.md
```diff
@@ -133,7 +133,9 @@ SI:
 | ElectricPotential      | Volt                             | V      |
 | ElectricCapacitance    | Farad                            | F      |
 | ElectricCharge         | Coulomb                          | C      |
+| ElectricCharge         | Ampere Hour                      | A h    |
```
Markkuuss: Analog to the implementation also in the documentation the correction without spaces.

Confectrician:
>How should we deal with that?

No action needed.
Documentation is always meant for the "latest" state.
This means we are equal to the latest snapshots, which are OH3 now.

I have created https://github.com/openhab/website/issues/221 to discuss the general presentation of the docs while preparing OH3.

PR #1131, opened 2020-01-04
Fix website build and introduce 2.5.x based docs

Confectrician:
This will change the fetching of external docs to a build that is based on 2.5.x branches (where available).
I have also added some small improvements to the bash script and increased the version of the pulled feature xml.

Docs will now be build based on the 2.5.1-Snapshot artifacts.

Maven Build succeeded locally already.

cc @kaikreuzer 
This should solve https://github.com/openhab/website/issues/221 too.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -153,7 +153,7 @@
               <outputDirectory>${basedir}/_addons_persistences/oh2</outputDirectory>
               <resources>
                 <resource>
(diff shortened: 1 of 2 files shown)


Confectrician:
I did 3 local builds now, which ran very smooth.
Just some missing logos (as usual) and some warnings about ios (who have been there for ages).

Issue #1133, opened 2020-01-05
openHAB 3 Documentation  - ToDo


Confectrician:
# Goal

This issue is meant as a To-Do list for tasks that we have to accomplish for the openHAB 3 release.
This is mainly adressed to the internal parts of this repository and the website appearance.
External content can be fetched from the OH3 branches of the corresponding repositories with some effort.
The contributors of those repositories should check and improve contents themselves in the corresponding repositories.

## Accomplished (so far)

> Concepts => https://github.com/openhab/openhab-docs/issues/1310#issuecomment-744047078
> Installation Guide => https://github.com/openhab/openhab-docs/issues/1310#issuecomment-744047092
> Configuration Guide => https://github.com/openhab/openhab-docs/issues/1310#issuecomment-744047100
> Interfaces and Ecosystem => https://github.com/openhab/openhab-docs/issues/1310#issuecomment-744054946
> Administration Guide => https://github.com/openhab/openhab-docs/issues/1310#issuecomment-744047105
> Developer Guide/Appendix => https://github.com/openhab/openhab-docs/issues/1310#issuecomment-744049193

ghys:
I would say let's create a subdomain like v3.openhab.org, dev.openhab.org or next.openhab.org for the 3.0 docs and deploy a site with its own workflow. The main openhab.org would for now display the 2.5.x docs, maybe with a link to switch to the "next version" site. The home page's "hero" image would be due for a change as well when 3.0 is released, maybe we can also prepare a v3 demo server in the meantime, depending on the burden of maintaining it.

Also, eventually I'd like to revisit the version switcher widget which appears on each page and abandon the principle of staying on the "same" page when switching versions - we overlooked how quickly it can become a PITA when the structure changes (and I suppose it will change significantly for 3.0) and we end up with a lot of dead links. I'd rather put a combobox in the left column with a list of versions and navigate back to the intro page (/docs/) when switching.

kaikreuzer:
Sounds good to me. Would you create a new site in Netlify? I would then add an dns entry for a sub-domain (next.openhab.org sounds good, although I am fine with any other option as well).

Confectrician:
I would vote for `next.openhab.org` too. Sounds catchy.

bwosborne2:
If you use v3.openhab.org that will be obvious and "future-proof" whenever v4 arrives. You could even just redirect the main page to the latest version.

ghys:
[...]

PR #1151, opened 2020-02-06
tip if you installed from openHABian

erichgoldman:
I did a quick setup on rPI using openHABIAN per --> https://www.openhab.org/docs/installation/openhabian.html#raspberry-pi-prepackaged-sd-card-image . Coluld not access via SSH using these instructions, found this and was able to get in to start editting files remotely. Maybe additional information is needed for this note?

### DIFF
--- a/administration/console.md
+++ b/administration/console.md
@@ -64,6 +64,10 @@ On slower systems, such as Raspberry Pi or Pine64, this first SSH connection may
 If this happens, simply try connecting again until successful.
 :::
 
+::: tip Note
+If you are using the openHABian installation image, the [default login username/password](https://www.openhab.org/docs/installation/openhabian.html#raspberry-pi-prepackaged-sd-card-image) is **openhabian:openhabian**
+:::
+
 ## Using the Console
 
 After successful connection and authentication, the console will appear:


ThomDietrich:
Wouldn't it be easier and clearer if you added this note directly to line 58?

BClark09:
I don't think this section talks about the same thing you've added.

In the docs you've edited, it is talking about accessing the openHAB console via SSH and *not* the machine that has openHAB installed. Same protocol, but accessing different things.

The page you linked to does contain the information you seek:

> Ready for more? Connect to your Raspberry Pi SSH console using the username openhabian and password openhabian. You will see the following welcome screen:

But perhaps this can be made clearer?

ThomDietrich:
Ben is totally right. Ignore my previous comment.

I feel like this paragraph (with the visual highlighting of the password ) is already pretty clear. But that might just be the proud author speaking 😋

erichgoldman:
New to openHAB so not sure exactly sure what is supposed to be here. I believe I was looking to connect to upload a file or view a log. When searching for SSH I found this page. 

Should those instructions and logins work on openHABian as well? That port wasn't open and those credentials did not work for me. Is this something which isn't part of the openHABian install or should I be able to access this console on my installation with these creds?

erichgoldman:
To follow the docs, I sshed with openhabian, then ran `openhab-cli console` so I think that then puts me at the same place.

ThomDietrich:
Hey Eric,
you simply ended up in the wrong article.
However, that we should avoid by making clear that what you found is not what you are looking for. Looking at the console page I see this:

image

I would propose you come up with a solution that reduces the risk this is wrongfully understood as the way to login to your system via SSH. I'd suggest a headline like "Remote Connection" instead of "Connecting via SSH" and starting the article with something like "The openHAB console offers the option to:". This and maybe 2-3 more changes should make clear that this article is specific to the console (an advanced interface of openHAB) rather than the SSH console of your host system (e.g. openHABian).

While at it, Could I ask you to move the note block before the code block with `ssh -p 8101 openhab@localhost`?

ghys:
Note: changed the base branch to `2.5.x` which is the default for now, and will appear on the production website.
`master` is now the branch for openHAB 3.x docs.

@openhab/documentation-maintainers 
Once this is merged to `2.5.x` please remember to merge (for example, cherry-pick the commit) this to `master` as well if appropriate.
See https://github.com/openhab/openhab-docs/issues/1133 for details.

mstormi:
Obsoleted by changes made in the meantime
@Confectrician pls close

Confectrician:
> Obsoleted by changes made in the meantime
> @Confectrician pls close

PR #1157, opened 2020-02-13
Update rules-dsl.md

MisterT2208:
Example for heating 

"val Number Temperature" was missing


openhab-5iver:
This is incorrect. In this example, `Temperature` is an Item, the use of which is what the example is illustrating.

kaikreuzer:
@openhab-5iver is right.

PR #1164, opened 2020-02-25
Added note about broken action

cweitkamp:
- Added note about broken action

Probably it will not be working before OH 3.0 - even if it is already fixed in our code base.

See https://github.com/openhab/openhab-core/issues/1374


Confectrician:
I thinks this doesn't get noticed that well.

image

Maybe we can add some custom container for the note.

```
::: warning
Warning goes here ....
:::
```

cweitkamp:
@Confectrician Thanks for the feedback. Is it possible to place the warning inside a table? Or should I put it underneath the table?

Confectrician:
Give it a try and see what the preview will throw out. 😄 
I have never done that before in a table.

cweitkamp:
Doesn't work nice:

grafik

I added a "warning" emoji now:

grafik

Confectrician:
I think the current solution is fair enough.
Thanks. 👍

PR #1166, opened 2020-03-03
Update eclipse.md

billfor:
The pom.xml insert reference includes a version that may need to be updated from the version bundled with eclipse, if the developer has checked out a newer version of openhab, otherwise trying to run the bundle results in an empty dialog with no error message.


mstormi:
please update or close

Confectrician:
I will try to add this to the 2.5.x branch and then cherry pick it into the main oh 3 docs.
But i first wanted to see, if it already covers the actual state or if it would be better to rewrite it due to the big gap between the inital pr and now.

Confectrician:
Too many files affected. I think cherry pick would be difficult here and we will have added some changes already through the year.

Closing for now.

PR #1175, opened 2020-03-20
Reference 2.5.x branch and clarify paths for pom.xml files

bobadair:
Signed-off-by: Bob Adair <bob.github@att.net>


bobadair:
These changes won't apply once master becomes the default branch again.

PR #1177, opened 2020-03-23
Replace outdated zulu.org link.

wasv:
As of 3/23/2020 zulu.org has an SSL cert that expired on 9/28/2019. Changed link to azul.com/downloads, since that appears to be the new official source.


wasv:
> Hi,
> 
> Maybe we should change those links to `https` too.
> The page seems to support it fine. :)

I also checked for any other http links that could be replaced with https links, and updated those as well.

PR #1197, opened 2020-04-18
Installation details

theiding:
Added more details around the installation and configuration process.
Fixed that engine no longer logs "Activated scripting support..."

### DIFF
--- a/configuration/jsr223-jython.md
+++ b/configuration/jsr223-jython.md
@@ -10,8 +10,10 @@ title: JSR223 Jython Scripting
 ## Configuration
 
 [Download](https://jython.github.io/index) the Jython 2.7.0 standalone or installer package, or install through a package repository.
-Newer versions are not yet supported.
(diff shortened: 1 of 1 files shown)


theiding:
Hm, how come my sign off failed? I see it's asking for sign off by theiding <mark.theiding@gmail.com>. In the past I always signed off with Mark Theiding <mark.theiding@gmail.com>.

Confectrician:
How did you make the changes?
GitHub Web editor or on your local device?

I don't know exactly how the dco but works, but it has to compare the sign off email with your account somehow, so i would check your personal email settings for github.
(Just a rough guess)

Confectrician:
I have checked the bot.
It seems to simply compare your github name with the sign-off statement.
That why it expects the sign-off to only be "theiding" in your case.

Personally i would say that is not right and your sign-off should be taken as valid by the bot too.

PR #1179, opened 2020-03-28
Ending an active scan/stopScan

theiding:
I only learned about stopScan details from reviewer feedback when implementing a binding. Before that I had implemented duplicated functionality. I hope that a bit more documentation on this topic will help future binding developers avoid that confusion.

PR #1205, opened 2020-04-26
Add section for documentation contributions

theiding:
Hopefully this will lower the hurdle for people to submit documentation contributions. I know from myself that I didn't submit various documentation improvements, because I didn't know git and thought it would be a much more involved process. 
Ideally there would be a separate documentation section, but submitting this under the development contribution page for now (as per discussion with @Confectrician in https://github.com/openhab/openhab-docs/pull/1179#issuecomment-605642091).
Note that I am addressing the issue of DCO failures wrt specifying the full name that I ran into myself in https://github.com/openhab/openhab-docs/pull/1197#issuecomment-615597308. I found a good discussion of the issue at https://github.com/probot/dco/issues/43.

### DIFF
--- a/developers/contributing.md
+++ b/developers/contributing.md
@@ -103,6 +103,7 @@ or `Fixes #XXX`, which will automatically close the issue when merged.
 openHAB maintainers use the [GitHub review feature](https://help.github.com/articles/about-pull-request-reviews/) to indicate acceptance.
 
 
+<a id="sign-your-work"></a>
 ### Sign your Work
 
 The sign-off is a simple line at the end of the explanation for the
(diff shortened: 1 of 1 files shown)


Confectrician:
Hey @theiding 

Thanks for the work which looks good so far.
Maybe we should additionally add a tutorial i have written in the community.

Can you add a link to the following thread:
https://community.openhab.org/t/dco-check-signing-off-with-github-web-editor-explanation/83330

I think we are good to go afterwards.
Also i think it would be nice to put the additional information about dco check tool in this tutorial.

theiding:
@Confectrician , thanks for the feedback. I had thought about adding the link to the tutorial when I wrote the section, but decided against it. My thinking was: Tutorials are extended guides that can build upon and bring together documentation, but documentation doesn't point to tutorials. 
Would it make sense to lift the tutorial into the documentation? That way there is no duplication of information and everything is in a central place. What do you think?

Confectrician:
Thanks for your answer.
Your points are perfectly valid. :)

>Would it make sense to lift the tutorial into the documentation?

Maybe we should leave it as is for now then and merge this into a real docs contribution page at some point.

theiding:
Should we create that separate, dedicated documentation page now? Do you know what it takes to do that? That is above my pay grade :-).

Confectrician:
I think it is fine for now. :)
I will merge this one. 

Creating the page is not that difficult.
Bringing it into the navigation and choose a wise place is the challenge.

PR #1216, opened 2020-05-04
Update sitemap.md

GerdZanker:
fixed bad link to sitemaps


Confectrician:
Hey @GerdZanker,

Thanks for the catch!
This is a duplicate of the later proposed #1217 which i had already started working on.
I will merge that one and the fix will be included that way, so i will close this one.

Anyway thanks again for the good catch. :)

PR #1217, opened 2020-05-05
Line 115 broken link - should be:

bugsmurf:
({{base}}/docs/configuration/sitemaps.html#element-types)

was:
({{base}}/configuration/configuration/sitemaps.html#element-types)

PR #1224, opened 2020-05-12
fix sitemap elements link

supersuade:
fix sitemap elements link


Confectrician:
Thanks for the catch @supersuade!

This is a duplicate of #1216 and #1217 and got already fixed in #1217.
Therefor closing this one.

PR #1225, opened 2020-05-16
Core Framework branch clarification

theiding:
Clarifying that it's no longer possible to make changes in the Core Framework for 2.5.x.


theiding:
Yikes, forgot sign off again. Fixing it.

PR #1226, opened 2020-05-17
Update eclipse.md

theiding:
As an IDE beginner I have always struggled with the black magic happening in the IDE. I have learned a few things that simplified how I think about the IDE in conjunction with Maven and Git. Obviously we shouldn't go into details about either Eclipse, Maven or Git in this page, but I hope by adding a bit more structure around the install and execution flow and pointers to what can be manipulated from the outside it will help other developers to get the big picture. 

Signed-off-by: Mark Theiding mark.theiding@gmail.com

### DIFF
--- a/developers/ide/eclipse.md
+++ b/developers/ide/eclipse.md
@@ -7,7 +7,11 @@ title: Eclipse IDE
 
 Eclipse is the development environment used since the inception of openHAB.
 To make development easier an out-of-the-box setup is available that completely configures Eclipse to easily develop for the openHAB projects.
-This guide describes the steps to setup Eclipse and how to debug an add-on in Eclipse.
+This guide describes the steps to setup Eclipse and how to run and debug an add-on in Eclipse.
+
+::: tip Existing Eclipse Installations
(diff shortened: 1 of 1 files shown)


theiding:
I don't understand why the DCO check failed. Sign off is provided, matches Github account and identical sign off worked in https://github.com/openhab/openhab-docs/pull/1205. Any ideas?

Confectrician:
Hey @theiding 

The only thing i could imagine is the "long commit message".
Don't know of DCO Bot cuts that off at some point.

Errors message states a "missing sing-off" but the one i have seen looks good from my point of view.

theiding:
Good point about the length of the message. Forgot that it is really the commit message that I entered. Should have made that short and then add the longer message as a comment here. Will do that next time and see if that fixes the sign off issue as well. 
Forcing another sign off now for this one.

theiding:
Duh, missed the <> in the sign off message.
Signed-off-by: Mark Theiding mark.theiding@gmail.com needs to be
Not sure though why the first comment in https://github.com/openhab/openhab-docs/pull/1205 shows the sign off without the <> (but the actual PR does have them and therefore succeeded).
Now I need to update the docs on how to submit doc changes where I specified the wrong format.

PR #1229, opened 2020-05-22
Remove Apache Commons from Default Libraries

fwolter:
See openhab/openhab-addons#7722

Signed-off-by: Fabian Wolter <git@fabian-wolter.de>


Confectrician:
Hey @fwolter 

Is this change relevant for OH3 only or should it be excluded in OH 2.5.x also?

cc: @cweitkamp

cweitkamp:
OH3. But it does not hurt to get rid of it right now :wink:.

Confectrician:
OK we will migrate it then and it is included now already.

PR #1232, opened 2020-05-25
[v3] Add Javadoc to sidebar

ghys:


ghys:
Handled by #1233 - turns out it works only in VuePress 1.x

PR #1262, opened 2020-09-01
[developers] Added Discovery that is bound to a Thing section

Hilbrand:


developers/bindings/index.md
```diff
@@ -929,7 +929,47 @@ It uses the `getThingUID` method to create the thing UID of the result.
 
 ### Discovery that is bound to a Thing
 
-TODO
+When the discovery process is dependent on a configured bridge the discovery service must be bound to the bridge handler.
```
cweitkamp, line 932: I suggest to use a different wording here. Currently it sounds like a mandatory design pattern but imo it's not.

Why? This way of implementation has a limitation I struggled into when trying to apply it to e.g. the OWM binding. The `AbstractDiscoveryService` for example provides access to other OSGi services (providers) like `Transtlation` and `Locale`. IIRC there's no way to inject those services when using the `ThingHandlerService` pattern. The OHC implementation doesn't take care of that.

Confectrician:
Anything new on this?
@Hilbrand do you want to respond to @cweitkamp s suggestion?

If not it would be ok for me to merge this one in its current form and cherry pick it into the oh3 docs.

Confectrician:
Hey @cweitkamp would you agree in merging it as is for now?
I would say that we have some valuable content here which is not merged because of one small discussion.
So i would like to get the main work merged already.

We should then open an issue and discuss this topic there until we have a conclusion.

Wdyt?

cweitkamp:
Yes, I would agree on merging this PR. We already have an pending PR for adding the feature. Thus it will not take much time after we can continue here.

Confectrician:
Could you file a short issue, so we can track the discussion over there?

Confectrician (approved):

PR #1264, opened 2020-09-14
[developers] Channels with options may be Number

Hilbrand:


wborn:
Perhaps you can also add some docs on when to use strings and when to use numbers? I think strings are still preferred to hide  "magic" values from users that are for instance used in protocols. Such "magic" values are very error prone as you always have to look up what they mean. It should be OK to use Numbers for values that represent "quantities".

wborn (approved):
LGTM!

PR #1267, opened 2020-09-21
Add groovy script to process thing types

ghys:
Gather all thing types XML and parse them into a JSON file,
similar to the way the website's crawlthings.ruby does it.


ghys:
Merging to check if Jenkins likes it...

PR #1274, opened 2020-10-07
Change to 2 ms instead of 10ms

pawlobanano:
2 ms is visible on screenshot. And I'm not exactly sure if description describes the exact situation (controlls) on the screen.
image

*I just started my adventure with OpenHab and it seems great!


Confectrician:
Thanks for your input.

Anyway i am closing this one, because we removed the complete tutorial in preparation of openHAB 3.

PR #1280, opened 2020-10-30
Add missing serial configuration page to sidebar

cpmeister:
I noticed that the serial port configuration guide is missing from the sidebar.
I hope that this change is sufficient to add it there.


Confectrician (approved):
Thanks. 👍

PR #1283, opened 2020-11-05
Added 'Number' item for possible type to use state options

cweitkamp:
- Added 'Number' item for possible type to use state options

See https://github.com/openhab/openhab-addons/issues/8543#issuecomment-721684801


wborn:
There was already a PR for adding this and some more details. See https://github.com/openhab/openhab-docs/pull/1264. But I think it somehow did not make it? Maybe you can also add the extra details.

cweitkamp:
Done.

ghys (approved):
Yes there were some problems with the rebasing when we considered that new docs contributions would be considered for the main branch instead of the 2.5.x branch. Thanks for taking care of it!
LGTM

PR #1285, opened 2020-11-09
No combinations of prefix+base unit symbol in the SI table

schmidmuc:
For some units there were symbols mentioned which are built from prefix+unit (e.g. Mbit/s was mentioned), for other units just the base unit was included in the table (e.g. only A was mentioned, but no kA oder mA). This is misleading/misunderstandable.

From my perspective the table SI should only show valid base units symbols and no combinations of prefix + base unit.

### DIFF
--- a/concepts/units-of-measurement.md
+++ b/concepts/units-of-measurement.md
@@ -80,7 +80,7 @@ The `org.eclipse.smarthome.core.library.dimension` and `javax.measure.quantity` 
 All units which are currently supported by default are listed in the tables below.
 
 
-Imperial:
+Imperial (base unit symbols):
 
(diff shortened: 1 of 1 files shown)


ghys (approved):
I think I agree with your assessment.
Can you review the contributing rules and add a sign-off line (https://github.com/openhab/openhab-docs/blob/main/CONTRIBUTING.md#sign-your-work) as a comment (no need to modify the commit message) here for merging?

schmidmuc:
Signed-off-by: Thomas Schmid <schmid.muc@gmail.com>

Confectrician (approved):

PR #1294, opened 2020-11-28
Updated rule dsl docs to openHAB 3 for datetime and implicit var changes

kaikreuzer:


ghys (approved):
Thanks!

PR #1303, opened 2020-12-10
Update logging instruction for OH3

mstormi:
Signed-off-by: Markus Storm <markus.storm@gmx.net>


administration/logging.md
```diff
@@ -127,89 +131,49 @@ log:set DEBUG org.openhab.binding.zwave
...
+In order to define custom log patterns, log to network sockets and so on we can prepare a logging configuration in `log4j2.xml`.
+There are several things that you might want to change in the configuration:
 
 - the log level for a logger;
 - the pattern of an appender;
 - redirect the log to a text file.
 
-The configuration file for openHAB is placed in the [openhab-distro/lauch/home/logback_debug.xml](https://github.com/openhab/openhab-distro/blob/master/launch/home/logback_debug.xml) file. We have added a few comments on this file in order to attract your attention on some significant points:
+<!--
+The configuration file for openHAB is placed in the [openhab-distro/lauch/home/logback_debug.xml](https://github.com/openhab/openhab-distro/blob/master/launch/home/logback_debug.xml) file.
```
wborn, line 175: Logback is only used during development (which isn't very clear from the current docs).

The logback config file is now: https://github.com/openhab/openhab-distro/blob/master/launch/app/runtime/logback.xml

It's XML configuration format is very similar to Log4j, see also:

* https://logging.apache.org/log4j/2.x/manual/configuration.html
* https://logback.qos.ch/manual/configuration.html

So for Log4j we probably want a similar section but with the Log4j XML syntax see: 

https://github.com/openhab/openhab-distro/blob/master/distributions/openhab/src/main/resources/userdata/etc/log4j2.xml

It might be an idea to move the Logback configuration details to a separate page because it might become confusing for normal users. They will likely start using Logback configuration syntax in Log4j files. :neutral_face:

administration/logging.md
```diff
@@ -127,89 +131,49 @@ log:set DEBUG org.openhab.binding.zwave
...
 
 ```
-2016-06-04 16:28:39.482 [DEBUG] [.e.model.script.heating-control.rules] Bedroom: Temperature 21.3°C, Mode NORMAL
+2016-06-04 16:28:39.482 [DEBUG] [org.openhab.core.model.script.heating] Bedroom: Temperature 21.3°C, Mode NORMAL
 ```
 
 Note that, in the last example above, inclusion and formatting of values is done using [Java Formatter String Syntax](https://docs.oracle.com/javase/7/docs/api/java/util/Formatter.html).
 
-## Logging into Separate File
+## Logback configuration and logging into separate files
```
wborn, line 162:
```suggestion
## Log4j configuration and logging into separate files
```

administration/logging.md
mstormi: So if logback is not in use by users (just devs as you say) it's probably safest to delete all of this from the docs.
I guess you yourself don't need this page any more, do you ;-) ?
So will delete what I commented out.

Confectrician: >It might be an idea to move the Logback configuration details to a separate page because it might become confusing for normal users. They will likely start using Logback configuration syntax in Log4j files. 😐

Maybe the development area would be a good place for this.

Confectrician (approved):
LGTM from my side. Any additions @wborn?

wborn:
Yes I think the users will now be able to reconfigure their logging with these docs! :-)

PR #1260, opened 2020-08-30
Fix typo and use lambda expression

fwolter:
Signed-off-by: Fabian Wolter <github@fabian-wolter.de>

PR #1306, opened 2020-12-13
Fix typo and use lambda expression

fwolter:
@Confectrician as requested in #1260 against `main`, now.

Signed-off-by: Fabian Wolter <github@fabian-wolter.de>


Confectrician (approved):
Thanks for moving this one. 👍

PR #1324, opened 2020-12-19
update of the Windows installation page

edolis:
Some updates to the Windows installation page for OH3

-changed the parameters of the wrapper.conf file, whose previous value caused the windows service to fail to start (ref. https://community.openhab.org/t/oh3-installation-service-fails-to-start-cannot-create-java-virtual-machine/110430/26 )
- made more clear the value to fill in the JAVA_HOME variable (il left to an old Java installation, installation fails) replacing the old path
- changed a bit the structure to make it more consequential and integrated with the docflow
- updated the screenshots 
- added reference to some maintenance task like uninstallation, which is important if the user wants to change version, repair a broken installation etc.
@hmerk
@rlkoshak
you review is appreciated thanks.
I will delete the previous branch patch-2, trying to fix the signing process it got too messy to fix.


edolis:
if someone can help me with the signatures. It says verified but DCO says no.. I can't understand.

PR #1331, opened 2020-12-19
Edits to administration index.md

bobadair:
Minor edits to administration index.md


Confectrician (approved):
Thanks. 👍 

Alltough we are trying to avoid karaf in the docs it should be ok for the first explanation here.

Issue #1339, opened 2020-12-19
[Armbian] Check if a refactoring is needed


Confectrician:
We should check if there are further updates needed to the armbian documentation.
I have already seen some outdated screenshots that should get a replacement in the nearer future.

cc: @igorpecovnik could you do us a favor and make a revision of your article, when openHAB 3 has been released?

PR #1343, opened 2020-12-20
Update Armbian install guide

igorpecovnik:
Close https://github.com/openhab/openhab-docs/issues/1339


Confectrician (approved):
Thanks for the quick solution. ❤️

PR #1349, opened 2020-12-20
Delete jsr223-jython.md

kaikreuzer:
I assume this is outdated here and not referenced anymore and thus should be removed.
Note: I have moved parts of it to the add-on with https://github.com/openhab/openhab-addons/pull/9447.


Confectrician:
I had left it over for exactly this usecase (move to addons repo), while already deleting the existing links. 🙂
With openhab/openhab-addons#9447 in place it should be safe to get deleted here.

Confectrician (approved):

PR #1353, opened 2020-12-21
[WIP] Initial refactoring of the rules-ng page

Confectrician:
Just some initial refactorings.

- Removal of old Paper UI content
- New screenshots
- Some formatting

To clarify: (cc: @ghys and @kaikreuzer )

- Should we explain the JSON here or maybe cut that out into a special article?
- Should we instead go for yaml rules which can be used in the ui directly?


Confectrician:
Dropping this for now, since we need a complete rules areaa refactoring

kaikreuzer:
This is outdated and dev-related content. End user rule documentation should all go into the "normal" rule section. 👍

PR #1354, opened 2020-12-21
Secure links should be standard, where available.

Confectrician:
No further explanation needed.


Confectrician:
Some more information:
I checked about 98% of the domains by hand und had to revert som https changes unfortunately.
It's a pitty to not provide a ssl certificate for a public website in 2020.

PR #1360, opened 2020-12-22
Exec actions timeout is now in seconds.

Confectrician:
Reference: openhab/openhab-core#1700


Confectrician:
Hey @cpmeister did i miss something else for the executil?

Confectrician:
Hey @wborn may you have a look too.

I have read about problems with exec ations in oh 3 with the message `Type mismatch: cannot convert from int to Duration`

What is the correct way to define a duration within rules?
We should add an explanation for this.

Confectrician:
I have found no additional information so far and want this to be included in the latest website build.
So we can catch up late if somethigns still missing.

PR #1361, opened 2020-12-22
Add migration tuorial

Confectrician:
Initial start for a migration tutorial.


Confectrician:
Merging for now to get a first version online.
Most text is copied from the release notes anyway so it should fit 90% already.

PR #1368, opened 2020-12-22
Fix typo in header

marbon87:


Confectrician (approved):
Thanks for the catch. 👍

PR #1372, opened 2020-12-23
getting_started, typo: "anda" --> "and a"

scotthraban:


scotthraban:
Will create a PR more in line with contrib guidelines :)

PR #1380, opened 2020-12-25
Make all openhabian docs available on the homepage.

### LINKED ISSUE #1201: [openhabian] Add amanda and debug docs to gathering script
Ref: https://github.com/openhab/openhabian/pull/851

Confectrician:
This will make all openhabian docs available during external content gathering.

@mstormi 
I have excluded the `NEWSLOG.md` file for now.
I don't think that one is necessary for now.
`READNEWS` will get sorted out because it has no markdown ending.

What about `exim.md`?
Should it be tranferred or is it not relevant for the website?

Fixes #1201

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -245,9 +245,12 @@
                 <resource>
                   <directory>${basedir}/.external-resources/openhabian/docs</directory>
                   <includes>
-                    <include>openhabian.md</include>
+                    <include>*.md</include>
                     <include>images/*.png</include>
                   </includes>
+                  <excludes>
+                    <exclude>NEWSLOG.md</exclude>
+                  </excludes>
                 </resource>
               </resources>
             </configuration>


Confectrician:
I want to have another build today or tomorrow. there fore merging this one for a test already.
If needed we can exclude the exim-md file with one line later on.

mstormi:
> What about `exim.md`?

exim is an active optional tool and exim.md up to date so if your intention is to have complete openHABian docs online then yes keep it

Confectrician:
Perfect. Thanks for the feedback.
You could also use relative links from now on.

mstormi:
@ecdye fyi

mstormi:
@Confectrician so we can now change all absolute links I've put into the .md files into absolute ones ?

Confectrician:
I think you have meant relative urls and yes you can. Docs are getting ported already.

Proof of concept available here

https://openhab.org/docs/installation/openhabian-amanda.html

Should be safe for newly added files too.

It would be nice to prepend an `openhabian-` to all files that should be used in the docs.
Makes it easier for us when we are looking into the repo directly.

PR #1381, opened 2020-12-26
Update New User tutorials first page.

Skinah:
I felt it was a bad situation that the first thing we hit a brand new user with on the very first page is the confusing situation of TEXT VS UI. Hence the change to encourage users to skip ahead.
Great to have it for V2 users upgrading, but confusing for the brand new user with zero openHAB experience.

Added links and reworded to make it more personal through the use of 'you' instead of 'the user'.

Signed-off-by: Matthew Skinner <matt@pcmus.com>

### DIFF
--- a/tutorials/getting_started/index.md
+++ b/tutorials/getting_started/index.md
@@ -10,7 +10,7 @@ title: Getting Started - Introduction
 Welcome to openHAB.
 We hope that your journey with openHAB is a rewarding one.
 If you encounter problems or simply cannot figure something out on your own, please do not hesitate to post a question on [the forum](https://community.openhab.org).
-There are many fellow openHAB users ready to help out, and polite constructive feedback always helps us to improve our documentation for the next new user.
+There are many fellow openHAB users ready to help out, and polite constructive feedback can help us to improve this documentation for the next new user.
 Most requests for help usually receive a response within a couple of hours, but please use the search on the forum to check if the topic has already been discussed.
 
 This new user tutorial assumes that you have at least a basic understanding of the [concepts of openHAB](https://www.openhab.org/docs/concepts/), and already have a working installation of openHAB.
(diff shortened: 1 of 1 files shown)


Confectrician:
First of all thanks for the contribution.

As this is a highly (partly emotional) discussed topic i would like to think about this first too.
I completely understand your motivations and would totally agree in the fact that we should not bother new users too much with "historical decisions" on the first page.

Anyway this tutorial is also adressed to openHAB 3 beginners migrating from openHAB 2 since we cover aspects like semantic modelling. We cant guarantee that only completely new users will read it.
So we should carefully cover all needed information and address possible concerns a migrating user can have.

Confectrician:
Reading a bit further in your changes and having a bit more overview i have the following additions.

## Situation
- We have a section Text vs. UI in the tutorial overview
- We have a section Text vs. file based in the configuration overview

For me this is redundant information that needs to be maintained in two places und furthermore may confuse a completely new user (which was the motivation of this Pull Request).

## My suggestions

I have two possible solutions in mind to solve this on long term.

### Move those contents completely to the configuration area and combine the sub articles

We could then add some short explanation and a proper hint with link to the configuration article where a interested user can find pros/cons/culprits/...

### Cut it out of both articles and make a new one

We could pay some attention to this highly discussed topic and give it a place in its own doc page.
We could make sure to add proper explanations and links to both of the above mentioned articles. (Less Maintenance needed.)
We could also explain different situations in this context more in Detail.



I would love to have some opinion on this from @rkoshak as initial poster of this article and from @mstormi cause he has refactored the configuration index page for openHAB 3.

Skinah:
Yes it is a sensitive topic I understand. Yes moving it out of the first page would be another approach. I am only thinking of the first impression of a brand new user as they say first impressions last. If the purpose of the guide is for an upgrading user then it is the right place for it to be.

the other issue I see which is connected to this, is that all the readme files for addons talk about Full Examples and how to do textual config but don’t mention the new UI. So we do need to cover this to prevent users from getting confused by the addon examples. If they are aware there are two ways they hopefully will understand.

As I read through half the bindings readme files recently, this stuck out like a sore thumb.

mstormi:
I'd think that we should keep these two locations.
That's redundant information yes but also very important information so it does not hurt to repeat.
A tutorial is barely the same thing as a reference page.
The two address different people to read this coming from different starting points such as upgraders vs. new users, beginners vs advanced users, readers to systematically browse the docs vs. Googlers, reader seeking guidance vs. those that just want to lookup stuff etc. We don't know the background of a reader and we need to have our docs be prepared for any of them.
We need to double-check there are no contradictions and we used keywords in the very same manner to also mean the same.
When we carefully do, that should not result in any user confusion whatsoever
(at least none to be attributed to our writing - contents and user expectations being different stories).

Confectrician:
Ok then leaving it as is may be the best for now.
I will check if we are able to cut out identical parts (like a pro and con list) and dynamically add them to both pages during the website build.
This way we could make sure that all areas where such contents are present are on the same version.

@Skinah 
MAbe we should then edit the encourament away from "skip this" to a "check out our UI based tutorial first and come back to have a look at text vs after doing the first steps".

Skinah:
Ok I made that change as suggested to go through the tutorial and then to come back later.
I also opted to remove programmer talk like 'parsing' out of the tutorial for more basic everyday language.

Confectrician (approved):
Thanks. 👍

PR #1384, opened 2020-12-26
Remove reference to importing .things file configuration

digitlength:
.things files cannot be bulk imported

### DIFF
--- a/configuration/index.md
+++ b/configuration/index.md
@@ -132,7 +132,7 @@ In openHAB 1.x, smart home configuration was done via configuration files only.
 openHAB 2 added the general administrative web interface "Paper UI", and openHAB 3 now further streamlines input capabilities with "Main UI".
 Things and Items can still be defined either in configuration files or via the GUI.
 We highly recommend adding them to the [system database](/docs/administration/jsondb.html) via Main UI, though.
-Note there is an option in Main UI to bulk create Things and Items by copy and pasting the contents of existing .things/.items files.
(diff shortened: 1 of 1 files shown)


Confectrician:
Hey @ digitlength this has already been solved by #1355, you have possible had an outdated version of the docs repository in your fork.

digitlength:
> Hey @ digitlength this has already been solved by #1355, you have possible had an outdated version of the docs repository in your fork.

How often do the docs get updated? All I did was click the link at the bottom of the relevant page, which at the time still had the issue.

Confectrician:
Sorry you are totally right.
That happens when one doesn't look close enough.

In #1355 there was a fix for line `41` of the file.
You are editing line `135`.
I simply didn't see that this was another occurrence of the wrong explanation.
I just thought "hey i have merged this some days ago" and started searching.
Finding the exact identical change without looking at the line number the case was pretty clear to me.

Sorry again and thanks for giving feedback again.

About your question:
Docs are not being transferred to the website on demand.
This happens on different triggers, like when openHAB Distro is build we are starting to fetch the latest addon doc changes automatically.

There is a high chance that a website build will be some days after the actual doc change has been merged here.
Usually that's not a problem, but especially in times of a fresh release many people loock (and/or contribute) to the docs in parallel.

digitlength:
Ah, that makes sense: I was starting to doubt whether I'd actually sobered up from Christmas...

Thanks for all you do!

Confectrician (approved):

PR #1387, opened 2020-12-27
Installation/index: add section for directory info

Rosi2143:
often in these pages directories are required and there
is no info how to get them yet.

e.g. where does $OPENHAB_CONF point to.

Signed-off-by: Michael Roßner <Schrott.Micha@web.de>

### DIFF
--- a/installation/index.md
+++ b/installation/index.md
@@ -87,6 +87,37 @@ Before you can start, two decisions have to be made:
 
 Please follow the instructions in the installation article matching your platform (see the menu to the left).
 
+### Getting installation info
+
+For many actions you need to know the pathes where openHAB checks for files, or where you need to install additional packages.
+To find out where these locations are, e.g. where `$OPENHAB_CONF` points to, use the [console](https://www.openhab.org/docs/administration/console.html):
+
+```shell
+openhab-cli info
+```
+ will result in e.g. for an [openHABian](https://www.openhab.org/docs/installation/openhabian.html) installation
+
+
+```shell
+Version:     3.0.0 (Build)
+
+User:        openhab (Active Process 7466)
+User Groups: openhab tty dialout audio bluetooth gpio
+
+Directories: Folder Name      | Path                        | User:Group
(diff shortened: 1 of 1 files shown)


Confectrician:
Hey @Rosi2143 thanks for starting this.
We have an issue for this #189 and a basic table in https://github.com/openhab/openhab-docs/issues/189#issuecomment-360506100

We should check if we cann/should combine this table with your edits.
For sure the table will be outdated for the openhab 2 paths.

Confectrician:
Having a preview available i would say this suits most of the cases already, at least when openhab.cli is available.
I would doubt that on windows installations but windows has its own file locations table in place.

So we maybe should at least add some clarifiaction on where to use the openhab-cli command (think of a beginner reading the docs first time)

installation/index.md
```diff
@@ -87,6 +87,36 @@ Before you can start, two decisions have to be made:
 
 Please follow the instructions in the installation article matching your platform (see the menu to the left).
 
+### getting installation info
```
Confectrician, line 90: Should be `### Getting installation info`

installation/index.md
```diff
@@ -87,6 +87,36 @@ Before you can start, two decisions have to be made:
...
+
+For many actions you need to know the pathes where openHAB checks for files, or where you need to install additional packages. To find out where these locations are, e.g. where `$OPENHAB_CONF` points to, use the [console](https://www.openhab.org/docs/administration/console.html):
```
Confectrician, line 92: Please use one line per sentence.

installation/index.md
Rosi2143: Done.

installation/index.md
Rosi2143: Done

Rosi2143:
Changes done

Confectrician (approved):
Thanks. 👍

Confectrician:
I would say we can close #189 too with this PR in place.

PR #1388, opened 2020-12-27
jsr223: correct pathes still pointing to OH2

Rosi2143:
Signed-off-by: Michael Roßner <Schrott.Micha@web.de>


Confectrician (approved):
Thanks for the catch. We seem to have missed that one. 👍

PR #1392, opened 2020-12-27
Updates to administration / console

bobadair:
Replace example output with 3.0.0 output and other minor fixes.


Confectrician (approved):
Thanks. 👍

PR #1393, opened 2020-12-27
Update  administration / bundles

bobadair:
Update examples to show output from 3.0 and remove reference to Paper UI.


Confectrician (approved):
Thanks 👍

PR #1395, opened 2020-12-27
Add section about Authentication to restdocs

### LINKED ISSUE #1391: [OH3] Authentication is not mentioned in Restdocs/subscription examples do not work any longer
The [docs for the REST API](https://www.openhab.org/docs/configuration/restdocs.html) do not mention authentication in any way and starting with openHAB 3 when you set up openHABian, your system will be password-protected by default. This means that some of the examples given in the article will not work any longer, but you will see a 401 Authentication Required error instead. As far as I have found out, this applies to all `/things`-related queries, for example, i.e. also [the `ThingStatusInfoChangedEvent` subscript example](https://www.openhab.org/docs/configuration/restdocs.html#:~:text=/things/).

---

## Information I have collected so far as a non-domain expert

Possible ways to modify the `curl` commands:
- Add argument `-u {USER_NAME}`, then enter password interactively
- Add argument `-u {USER_NAME}:{PASSWORD}`
- Add argument `-u '{API_TOKEN}:'` (usually recommended approach)

Furthermore, basic authentication needs to be enabled in the `org.openhab.restauth` settings unless an API token is used, and maybe the preference "Implicit user role for unauthenticated requests" matters, too (todo: check this - I'm not sure about this, but it does not seem to be the case for the above curl variants).

---

**Original context:** [[OH3] REST API Authentication Help RESOLVED - Setup, Configuration and Use / Scripts & Rules - openHAB Community](https://community.openhab.org/t/oh3-rest-api-authentication-help-resolved/105933/10?u=linqlover)

If no one is keen writing the docs herself or himself, please only inform me that all information listed above is correct and I can open a PR myself. :-)

BTW: In what sense is [this section about authentication](https://www.openhab.org/docs/installation/security.html#authentication-and-access-control) (stating that "openHAB does not (yet) support restricting access through HTTP(S) for certain users") still up to date for OH3, given the fact that now there is a user system in the Main UI and we can manage API tokens?
And a further question - just for curiosity: Is it already possible to create multiple users or to require authentication for every access to OH and its component, i.e. also items etc.? :-)

LinqLover:
Closes #1391.

`Signed-off-by: Christoph Thiede <christoph.thiede@outlook.de>`

### DIFF
--- a/configuration/restdocs.md
+++ b/configuration/restdocs.md
@@ -28,37 +28,37 @@ Nevertheless, here is some examples using [curl](https://en.wikipedia.org/wiki/C
 
 - Switching ```My_Item``` OFF by issuing an http [POST](https://en.wikipedia.org/wiki/POST_(HTTP)) request:
 
-```java
-curl -X POST --header "Content-Type: text/plain" --header "Accept: application/json" -d "OFF" "http://{openHAB_IP}:8080/rest/items/My_Item"
-```
+  ```bash
+  curl -X POST --header "Content-Type: text/plain" --header "Accept: application/json" -d "OFF" "http://{openHAB_IP}:8080/rest/items/My_Item"
+  ```
 
 - Setting a Contact item  ```My_Item``` to CLOSED by issuing an http PUT request to ```My_Item/state```:
 
-```java
-curl -X PUT --header "Content-Type: text/plain" --header "Accept: application/json" -d "CLOSED" "http://{openHAB_IP}:8080/rest/items/My_Item/state"
-```
+  ```bash
+  curl -X PUT --header "Content-Type: text/plain" --header "Accept: application/json" -d "CLOSED" "http://{openHAB_IP}:8080/rest/items/My_Item/state"
+  ```
 
 - Retrieving a list of all Items and Groups by issuing a GET request:
 
-```java
-curl -X GET --header "Accept: application/json" "http://{openHAB_IP}:8080/rest/items?recursive=false"
-```
+  ```bash
+  curl -X GET --header "Accept: application/json" "http://{openHAB_IP}:8080/rest/items?recursive=false"
+  ```
 
 - Retrieving a list of all sitemaps by issuing a GET request:
 
-```java
-curl -X GET --header "Accept: application/json" "http://{openHAB_IP}:8080/rest/sitemaps"
-```
+  ```bash
+  curl -X GET --header "Accept: application/json" "http://{openHAB_IP}:8080/rest/sitemaps"
+  ```
 
 - Subscription to events:
 
-```java
-// ThingStatusInfoChangedEvent - The status of a thing changed.
-curl "http://{openHAB_IP}:8080/rest/events?topics=smarthome/things/{thingUID}/statuschanged"
+  ```bash
+  # ThingStatusInfoChangedEvent - The status of a thing changed.
+  curl "http://{openHAB_IP}:8080/rest/events?topics=smarthome/things/{thingUID}/statuschanged"
 
-// ChannelTriggeredEvent - A channel has been triggered.
-curl "http://{openHAB_IP}:8080/rest/events?topics=smarthome/channels/{channelUID}/triggered"
-```
+  # ChannelTriggeredEvent - A channel has been triggered.
+  curl "http://{openHAB_IP}:8080/rest/events?topics=smarthome/channels/{channelUID}/triggered"
+  ```
 
 The commands above have been copied from the REST API documentation for illustration.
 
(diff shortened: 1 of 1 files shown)


configuration/restdocs.md
```diff
@@ -67,6 +67,27 @@ The commands above have been copied from the REST API documentation for illustra
...
+**Note:** This approach uses basic authentication which needs to be enabled in the `org.openhab.restauth` settings (accessible under Main UI > Settings > API Security).
+
+### With an API token
+
+This method is often recommended in order to keep your passwords safe and avoid to store them without encryption in any public places.
+To authenticate with an API token, **apaddpend `-u '{API_TOKEN}:'` to the commandline.**
+
+You can manage all access tokens in your profile settings in the Main UI.
+
+
```
Confectrician, line 90:
```suggestion
```

configuration/restdocs.md
```diff
@@ -67,6 +67,27 @@ The commands above have been copied from the REST API documentation for illustra
...
+To the complete command, **add `-u {USER_NAME}`,** then enter your password as prompted.
+You can also do this in a single, non-interactive command, in this case **add `-u {USER_NAME}:{PASSWORD}` instead.**
+
+**Note:** This approach uses basic authentication which needs to be enabled in the `org.openhab.restauth` settings (accessible under Main UI > Settings > API Security).
+
+### With an API token
+
+This method is often recommended in order to keep your passwords safe and avoid to store them without encryption in any public places.
+To authenticate with an API token, **apaddpend `-u '{API_TOKEN}:'` to the commandline.**
```
Confectrician, line 86: Are you sure about the colon appended to `{API_TOKEN}` ?

As stated here https://community.openhab.org/t/visual-code-and-openhab-3-0/109687/31?u=confectrician , vscode extension will omit the colon, so i am a bit unser about it.

Confectrician (commented):
Thanks for the quick PR.
I didn't expect it so fast, just assigned you for doing it in general.

See some small comments below.

Also would you mind to add a sign off statement to your first post?

configuration/restdocs.md
ghys: If you don't specify a colon curl will ask you for the password (which must remain blank so just hit Return). A colon with nothing after it represents a blank password.

Confectrician: Ah thanks for the explanation. 👌

Confectrician:
Hey @LinqLover and @ghys 

Maybe we should take the chance here and also add some explanation on how to generate an api token manually.
I would like to use this for the vscode workaround explanation.

@LinqLover i would assume that you have used the github web editor for proposing this change?

LinqLover:
Thanks for the review!

> Maybe we should take the chance here and also add some explanation on how to generate an api token manually.

Nah, I would rather put this on a separate page. Or are API tokens designed for the REST API exclusively?

> Also would you mind to add a sign off statement to your first post?

How can I do this without affecting the ancestry of the branch?

> @LinqLover i would assume that you have used the github web editor for proposing this change?

For 212858d1ee942c94f20592a2984dac6b6171ace7, yes. GitPod for the other commit ...

Confectrician:
> Nah, I would rather put this on a separate page. Or are API tokens designed for the REST API exclusively?

I have added #1398 to fix this independent from this PR. 🙂 

> How can I do this without affecting the ancestry of the branch?

You can add it manually to the first pull request post here, i will use it for the merge commit then.
Usually those statements should be included in every commit as last line.
That's how the DCO check will recognize them.

LinqLover:
Done - ready to merge :-)

Confectrician (approved):
Thanks 👍

configuration/restdocs.md
```diff
@@ -67,6 +67,27 @@ The commands above have been copied from the REST API documentation for illustra
 You can try and validate rest api calles from within the openHAB UI.
 Just log in with an admin user, navigate to `Developer Tools -> API Explorer` and start exploring.
 
+## Authentication
+
+Starting with version 3, openHAB supports password protection for sensible contents such as parts of the semantic model.
```
datMaffin, line 72: Why was the "supports password protection for sensible contents **such as parts of the semantic** model" added? I can not find anything about changing the protection of **parts** of the semantic model.

As far as I understand, the authentication works in the following way:
1. Sensitive things that can only be changed by an "Admin" user need authentiction of any "Admin" user.
2. Not sensitive things are by default accessible without any authentication.
3. If the "API Security -> Implicit User Role" is disabled, then authentication is needed by any "Admin" or "Users" user.

I.e., Users will always have access to all the Items. This statement also seems consistent with various explanations on the forums.

So why was this "**parts** of the semantic" model added here? As the semantic model contains items, and items can always all be accessed, why was this written? Should it be changed?

LinqLover: Probably it should be revised; I'm out of openHAB for two years and can't even remember writing this PR 😅 Feel free to submit a more precise explanation ...

datMaffin: No worries :)

I will probably post something in the forum and get this revised.

PR #1396, opened 2020-12-28
Update security.md

xiberg:
Line 28 still had a path for openhab2. Changed to openhab.


Confectrician (approved):
Thanks for the catch. 👍

PR #1400, opened 2020-12-29
Update items.md

olewetzel:
homekit tag was in old version present in item documentation.


Confectrician:
Hey @olewetzel i think that is not correct.
Tags can be used by a variety of bundles/integrations/addons and not all of them need a `homekit=` in place.

For HomeKit this is explained separately.
See: https://openhab.org/addons/integrations/homekit/#item-configuration

PR #1414, opened 2021-01-03
working example for executeCommandLine in ECMA scripts

Boby71:
The executeCommandLine command doesn't work in ECMA scripts as described.
I added an additional, working example to make it easier for beginners to understand what to do in this case.

### DIFF
--- a/addons/actions.md
+++ b/addons/actions.md
@@ -87,6 +87,11 @@ executeCommandLine("path/to/my/script.sh", itemState1, itemState2);
 // When you need the output in your further rule processing
 var ScriptResponse = executeCommandLine(Duration.ofSeconds(60), "path/to/my/script.sh", itemState1, itemState2);
 ```
+When using ECMA rules, you need to make the Java functions available like in this exampe:
+```text
+var Exec = Java.type("org.openhab.core.model.script.actions.Exec");
+Exec.executeCommandLine("path/to/my/script.sh");
+```
 
 ### HTTP Actions
 


kaikreuzer (changes requested):
I would not add this to the docs.
Why is `sendHttpGetRequest` working (is it), while this method is not?

ALL actions should imho be treated the very same way. We should hence rather find a solution for https://github.com/openhab/openhab-core/issues/2056 and THEN document it here, if necessary.

Confectrician:
I think sendHttpGetRequest is not working neither with ECMA Script.
I just answered a question in community forum about this.

The problem with this article is a more common one, because nearly all main articles have "Rules DSL" in mind as a base.

We have some issue around here which is exactly adressed to this.
The idea to solve this on long term, is introducing tabbed examples for different rule languages.
Now with openHAB 3 released this seems to be more urgent than ever.
(I stopped all activities there, since we only had dsl and some helper libraries at that time.)

So we should aim for a mid-term/long-term doc solution here instead of fixing single articles.

kaikreuzer:
@Confectrician I referred to the http action, because this is explicitly mentioned for JS [here](https://github.com/openhab/openhab-docs/pull/1414/files#diff-c75f1c61eae10e044b9005b26e8fdf71944e7dc27b289dd936abac5f30f6723fL109-L111). I assume that it hence at least worked that way in the past. And calling `Java.type` to access actions is imho not how it should be, so we should consider it as a bug that should be fixed instead of documented.

Confectrician:
I had no time to test it in deep to be honest.
We should provide some working ECMA Script Explanation in one of the first steps.
That's what is hardly missing currently. And the old jsr 223 docs are unusable.
The link to lewies old js helper lib after 3 sentences.

I have no real knowledge of what is really possible or even best practice myself.
Saying this as a user who is using 100% ecmascript rules.
Most of my rules are using included java types currently.
It is working pretty well but we should define a best practice if this is not the best way to achieve special includes.

Boby71:
> @Confectrician I referred to the http action, because this is explicitly mentioned for JS [here](https://github.com/openhab/openhab-docs/pull/1414/files#diff-c75f1c61eae10e044b9005b26e8fdf71944e7dc27b289dd936abac5f30f6723fL109-L111). I assume that it hence at least worked that way in the past. And calling `Java.type` to access actions is imho not how it should be, so we should consider it as a bug that should be fixed instead of documented.

I already opened an issue for this problem, but thought it's more a documentation problem:
https://github.com/openhab/openhab-core/issues/2056
I'd really appreciate if this could be treated as bug, because with adding the function Duration.ofSeconds(60) I had to add another line to get it working: 
`var Duration = Java.type("java.time.Duration");`

As a non Java-speaking person it's extremely hard to find the right libraries/classes to add, so if the ECMA rules would automatically provide these functions already referenced, it would be very helpful in the migration from 2.5 to 3.x

Confectrician:
Closing here for now.
We should first get some results in openhab/openhab-core#2056 to see where and how we need to improve things.

kaikreuzer:
@Confectrician As https://github.com/openhab/openhab-core/issues/2056 has been closed without a fix, should we consider reopening this PR or has the documentation evolved too much by now?

addons/actions.md
```diff
@@ -87,6 +87,11 @@ executeCommandLine("path/to/my/script.sh", itemState1, itemState2);
 // When you need the output in your further rule processing
 var ScriptResponse = executeCommandLine(Duration.ofSeconds(60), "path/to/my/script.sh", itemState1, itemState2);
 ```
+When using ECMA rules, you need to make the Java functions available like in this exampe:
```
kaikreuzer, line 90: typo: exampe

Confectrician:
I would go for reopening, but the branch was deleted already so we have to introduce the changes on our own in a new pr

rkoshak:
See issue #1855. Much of what is discussed in this PR is being reworked by me and @florian-h05 under that issue. We are reworking the entire structure of Rules documentation so they are no longer Rules DSL, or any language specific. Our overall approach is to first present the generic rules concepts (triggers, actions, conditions) and those things that are the same across all the languages (e.g. the types of triggers). Then it's the responsibility of each rules engine to provide reference documentation on how to syntactically implement the rules concepts in that language.

As part of this we plan on also reworking the Actions page as well using the same overall approach because Actions really are a rules thing and, at least from the end user's perspective, not something that exists on their own.

So if you all can be patient, we might fix this as part of that work as a matter of course. If it's important to get this fix out fast, then I'd recommend including not just ECMAScript 5.1 but also Blockly in the examples. There are already examples documented as part of the JS Scripting reference guide. The rest of the automation add-on's readmes are pretty anemic so I've no idea about them.

Confectrician:
For me it is not that urgent.

I will close my PR and leave the branch there, if we get to an urgent point before the rules docs are refactored.

PR #1420, opened 2021-01-04
[installation/rasspi.md] General overhaul

edolis:
some minor formatting and a couple of advices added to the logging section

Signed-off-by: Emanuele Dolis <edoliscom@yahoo.it>


edolis:
@Confectrician  I knew it! I made a mess... if it's hard to fix and you can delete, I''ll try to use my main instead of this branch.  Be patient,pls. 
I know it's a death by thousand cuts :)

Confectrician:
> I knew it! I made a mess...

Where did you made a mess?

Usually a second branch is the way to got.
Using `main` may technically work in some cases but should be avoided if possible.
(I later wanted to tell you this, but as you have another branch in place anyway now i will do it a bit earlier.)

Are things from your other PR missing?
Generally it looks good alltough i havent compared both PRs yet.

edolis:
OH, I did something right not meaning it then. LOL. I was horrified by the duplication - thought should have flowed in the other one.
Yes, should be including the previous amendments if I got everything right. Just sorry that, thinking it would flow in the other pull, I overlooked the description and then you get that sloppy raspi2 title.

Confectrician:
That's the cool thing about pull requests. :)
You only can mess up your own copy of the repo by accident.
Mostly there is a solution available and if it is broken too much, you can just delete your copy/fork and make a new one.

Of course i know that you would like to avoid that, but that simply happens when you are new to this. 😄 
( For legal reasons i can't tell you how much forks i have deleted after messing them up in the past. 👅 )

Back to some serious stuff for your information:

When you want to add something while the pull request is opened,
you have to make sure that the following things are correctly set up.

The branch you are using for this pull request 

image

has to match the one in vscode

image

you can check that at the vscode status bar.

You can make additional changes, save them and do another commit.
After that you only have to synchronize the vscode repo with github.

I think you know all of this already (and even a bit more since about 1 hour), but maybe someone reading here will ifnd this information useful too. :)

edolis:
Thanks,
I am still trying to streamline the process. I use VS code and previously (for the digital signature nightmare which seems to be pointless) GIT desktop which I'll ditch since I am not sure if helps anyway or is redundant burden with VScode
Still confused on the overall process.
How do you work efficiently usually in VScode? So far I understood this (maybe wrong
- start from openhab/openhabdoc + main (to be sure not be working on old stuff)
- create branch say foo
- -do your edits and commit to edolis/openhabdoc foo (but here first doubt: edolis would be potentially not updated: Should I every time rebranch edolis/openhab and start from it instead??)
- Repeat edits and commits as required
- then? For the pull can you make it in VScode or you have to go through the GIT web interface? I'd try but then I'd come littering around with test pulls.
Please suggest (and if you can explain this in some bulletpoint post on the community, you'll rake around so many karma tokens you can't imagine!)

Confectrician:
If you have some time to wait i would love to write that down structured.

That's exactly what i want to explain here in the future:

https://community.openhab.org/t/wiki-how-to-contribute-to-the-openhab-documentation/111308/3?u=confectrician

edolis:
:'(
The last point please... just an hint!. Or I'll have to start _project crazy pull/push attempts_ and see what happens.

I know,  I know, you perfectionist, I continuous improvement - let's meet somewhere closer to me jusst once!

Confectrician:
If its just for the last point:

There should be a little synchronize button, if you set up your repo correct:
image

When edits are proper commited it may show a number of commits to upload.
With synchronizing through the button the commits should appear hear.
(Assuming that you editet the correct branch like explained above.) 🙂

Confectrician:
Hey @edolis,

Looking at the pr again and not the talk about contribution i would say that it looks good as is.
Do you want to edit something from your side or are we good to go for merging?

edolis:
well, at the moment for this page I think that is it on my side, I have got other idean but not pertaining rasspi.md. 
Except that maybe I would pack up a page for absolute beginnersn in Linux )aimed at win users like me=. But I think it should be discussed, and not know where yet. Let us give it a go.

Confectrician (approved):
Thanks 👍

PR #1426, opened 2021-01-09
Remove reference to non-existent tutorial

digitlength:


Confectrician (commented):
Hi @digitlength 

Thanks for your contribution 👍

I would suggest to not remove the reference but point it to our new beginner tutorial instead.

We have https://openhab.org/docs/tutorial/things_simple.html which should be a good replacement.

The correct link would be:
`[beginer tutorial]({{base}}/tutorials/things_simple.html)`

Do you agree and could you change this?

PR #1428, opened 2021-01-10
Add info how to disable authentication

spacemanspiff2007:
Signed-off-by: Sebastian Gerber <github@sgerber.com>

### DIFF
--- a/configuration/restdocs.md
+++ b/configuration/restdocs.md
@@ -87,6 +87,18 @@ To authenticate with an API token, **apaddpend `-u '{API_TOKEN}:'` to the comman
 
 You can manage all access tokens in your profile settings in the Main UI.
 
+### Disable authentication
+
+It is possible to disable authentication.
+Stop openhab and add this part to ```userdata/etc/org.apache.karaf.features.xml``` before ```</featuresProcessing>```:
+
+```
+    <blacklistedFeatures>
+        <feature>openhab-*-auth</feature>
+    </blacklistedFeatures>
+```
+Once openhab is restarted authentication will be disabled.
+
 ## Additional Considerations
 
 The REST API also supports server-push supporting subscriptions on change notification for certain resources.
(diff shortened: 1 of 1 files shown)


configuration/restdocs.md
```diff
@@ -87,6 +87,18 @@ To authenticate with an API token, **apaddpend `-u '{API_TOKEN}:'` to the comman
 
 You can manage all access tokens in your profile settings in the Main UI.
 
+### Disable authentication
+
+It is possible to disable authentication.
+Stop openhab and add this part to ```userdata/etc/org.apache.karaf.features.xml``` before ```</featuresProcessing>```:
```
Confectrician, line 93: We should consider using `$OPENHAB_USERDATA` as term placeholder here like done in other articles.

```
$OPENHAB_USERDATA/etc/org.apache.karaf.features.xml
```

Confectrician (commented):
Hi @spacemanspiff2007 and thanks for the contribution,

Alltough i would call this an advanced rest topic, i have no better place in mind for this part.
So we can leave it here for now. 

You sign-off, alltough added correct, seems to have issues.
Do have conencted this email adress with github already?
At least it is not public available from your profile.

Just to be clear:
That does **not** affect your PR, since the sign-off is ok from my side and i will override the check in case its needed.
I am just curious, to understand the reason why check is failing here to learn for the future.

spacemanspiff2007:
> You sign-off, alltough added correct, seems to have issues.
> Do have conencted this email adress with github already?
> At least it is not public available from your profile.

It's because of the github privacy feature that strips mail addresses from commits.
I have to manually set the mail address to override it but because I created this commit from the web gui I am not able to.
The only way is to create the PR locally and then push it to Github.

I have made the changes as you requests. :+1:

Confectrician:
But it should have worked for your first commit was correctly signed off and the DCO check failed too.

It is missing for the second commit, that's correct.
(You can simply add it in the commit message yourself and it will already work.)

For the first commit, GitHub expected something different than you have send.

image

I would guess that this is related to the profile settings.

Confectrician (approved):
Thanks 👍

PR #1449, opened 2021-01-13
Add link check action.

### LINKED ISSUE #103: Check for broken links regulary

### LINKED ISSUE #565: Continuous checking for broken links and the remark style
Hey everyone.

I was thinking about a way to automatically generate **broken links** and **markdown lint** statistics, e.g., after every commit to gh-pages. I've been looking at various Github integrated [Code Quality](https://github.com/marketplace/category/code-quality) tools, however none of them seems to offer this kind of freedom. Most of them are specialized to e.g. run a python linting. I am looking for an online tool because I specifically do not want to set up a dedicated service for this kind of continuous thing.

The result of such an integration could e.g. be some report plus a nice little badge in the README, informing about the count of errors.

Does anyone know of a good solution for the use case?

Alternative: We have a process in place to retrieve and include external content, like binding READMEs. Executing the above tests and including their output as part of that process would be easy. The check results could e.g. be piped into a file in the repo. It's something... Wdyt?

Linked to #564  #103 #101

Confectrician:
Fixes #103
Fixes #565 


Confectrician (commented):
Ignoring Failed build, since there where no changes made to contents or docs related parts.

PR #1455, opened 2021-01-15
Update items.md

sancho-sumy:
Example with item Switch Kitchen_Light in the example is not correct for current binding version.


Confectrician (approved):
Thanks for the catch 👍

No [sign-off](https://github.com/openhab/openhab-docs/blob/master/CONTRIBUTING.md#sign-your-work) needed due to [small patch exception](https://github.com/openhab/openhab-docs/blob/master/CONTRIBUTING.md#small-patch-exception)

PR #1461, opened 2021-01-19
Add hint that /etc/default/openhab2 is not considered by the upgrade

BertramVielsack:
When you use a raspberry you may have used the file /etc/default/openhab2 to changed JAVA options.
After upgrade to openHAB 3.0 this file is no longer used. Instead the file /etc/default/openhab must be used.

See also: https://community.openhab.org/t/settings-from-etc-default-openhab2-are-gone-after-upgrade-to-release-3-0/114645

### DIFF
--- a/configuration/migration/index.md
+++ b/configuration/migration/index.md
@@ -30,13 +30,18 @@ If you are working with an [openHABian](https://www.openhab.org/docs/installatio
 
 ### Package-based Installations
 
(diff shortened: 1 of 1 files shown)


Confectrician (approved):
Thanks 👍

I have added some line breaks, since we are doing one line per sentence here.

No [sign-off](https://github.com/openhab/openhab-docs/blob/master/CONTRIBUTING.md#sign-your-work) needed due to [small patch exception](https://github.com/openhab/openhab-docs/blob/master/CONTRIBUTING.md#small-patch-exception)

PR #1468, opened 2021-01-26
Update NGINX configuration for OH3

Pedals2Paddles:
Added new directives related to authorization required for use with OH3.  Also added directives that are included in openHabian automatic configuration but were missing from the docs.  So now, the docs should more closely match what openHabian automatically generates.

### DIFF
--- a/installation/security.md
+++ b/installation/security.md
@@ -138,7 +138,8 @@ If this doesn't fit for you, you just have to replace `proxy_pass http://localho
 server {
     listen                                    80;
     server_name                               mydomain_or_myip;
-    
+
+    # Cross-Origin Resource Sharing
     add_header 'Access-Control-Allow-Origin' '*' always;
     add_header 'Access-Control-Allow_Credentials' 'true' always;
     add_header 'Access-Control-Allow-Headers' 'Authorization,Accept,Origin,DNT,X-CustomHeader,Keep-Alive,User-Agent,X-Requested-With,If-Modified-Since,Cache-Control,Content-Type,Content-Range,Range' always;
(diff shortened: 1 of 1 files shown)


ghys:
thanks for this, but I think the `add_header` should be added in the server context, and the `proxy_set_header` to the location context, see https://community.openhab.org/t/oh3-with-nginx-reverse-proxy-and-authentication/106528/53.

Pedals2Paddles:
Updated to reflect that now.

installation/security.md
```diff
@@ -209,6 +215,15 @@ Open the configuration file and **add** the following lines underneath the proxy
         auth_basic_user_file                  /etc/nginx/.htpasswd;
 ```
 
+##### Add authorization and cookie directives in NGINX Configuration
+
+This is an important new requirment in openHAB 3.0 and later versions.  This is not required prior to openHAB 3.0. In the `location /` block, you must add the following two directives underneath the `add_header` and `proxy_set_header` items respectively:
```
Confectrician, line 220: Please use one line per sentence here.
The website will take care of a correct formatting.

Confectrician (commented):
Hi @Pedals2Paddles and thanks for taking care of this. 👍

I had to look up the meaning of those headers and wondered if i am the only one here.
So my suggestion would be to add some comments to the config examples as well for unexperienced users.

For the genere headers we could add a `# Cross-Origin Resource Sharing.` and `# openHAB api authentication related` to the second block.

I think i have already added some comments in my personal content, so i will look them add and add some more explanations, when this pr got merged.

I have also added some small comment below.
(Don't worry if the preview fails after changing. It seems to be broken, but not by this pr.)

Confectrician (approved):
I have added some small twerks myself.
To small to prevent this information from being merged any longer...

Pedals2Paddles:
Not sure what "one line per sentence" is for, and I've never seen or heard of anything being written or coded that way. But thanks for merging.

Confectrician:
All of our docs are written that way.
Just means that one sentence has one line in that file. (With some minor exceptions.)
This is done to better keep track of changes here.

Our website then makes proper paragraphs during the build.

PR #1472, opened 2021-01-27
Resolve merge conflicts / Fix CI builds

Confectrician:
@ghys may you have a short look?

This should sync the branches again and resolve the existing merge conflicts.


ghys:
@Confectrician I suppose?
What caused the conflicts?
When merging I think you should rebase & merge in this case.

Confectrician:
> What caused the conflicts?

I have removed paper ui, openhabian and habpanel pages.

First one is deprecated for oh3 and had only been removed from the sidebar so far.
Both of the other pages are being maintained in their code repos for a longer time.
Seems that this caused some conflicts, since the pages are getting gathered from ci too.

We have to do a manual merge for the gathering part too probably.

Confectrician:
> When merging I think you should rebase & merge in this case.

Rebase and merge is not possible in this case 
image

Since we have the history available in the main branch, squash should be suitable to.
Wdyt?

ghys:
I think the last time I made it with the git command line. If you want I can try again.

ghys (approved):
Ah, you know what, let's squash, the history's on main anyway as you said.

PR #1475, opened 2021-01-29
Small clarification in modeling equipment section

kolcun:
Updating to use the same language as presented in the UI.


Confectrician (approved):
Thanks 👍

Also thanks for doing the sign-off statement. 👍🏼

Confectrician:
Additionally no need to worry about the failed page preview.
This is caused by a different problem.

PR #1477, opened 2021-01-31
Fix Typo

Trysupe:
Signed-off-by: Felix Schneider <fs@felix-schneider.org>

### DIFF
--- a/configuration/rules-dsl.md
+++ b/configuration/rules-dsl.md
@@ -345,7 +345,7 @@ These are methods of Objects that can accept a variety of types.
 
 Contrary, the Actions `sendCommand(MyItem, "<new_state>")` and `postUpdate(MyItem, "<new_state>")` can only accept strings as arguments.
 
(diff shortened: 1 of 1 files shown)


Confectrician (commented):
`reasons lies` sound wrong to mee too. 🤔

Trysupe:
> `reasons lies` sound wrong to mee too. 🤔

It's a common phrase: https://www.linguee.de/englisch-deutsch/uebersetzung/the+reason+lies+within.html

Confectrician:
Yeah i know that phrase but the current text ist `reasonS` `lies` i think you forgot to delete the `s` behind `reason` then.
Since its now clear what was meant just remove that and we can merge. :)

Trysupe:
2 be honest: I read `reason` and not `reasons`. Of course `reasons lie` is correct

Confectrician:
I am fine with both versions, so leave as is it will be for now.
Thanks anyway for staying active here. :)

PR #1482, opened 2021-02-05
Update the Time rule to a complete example and extend the explanation of .cancel()

Trysupe:
Source https://community.openhab.org/t/difference-between-timer-null-and-timer-cancel/116363/2

 Signed-off-by: Felix Schneider <fs@felix-schneider.org>

### DIFF
--- a/addons/actions.md
+++ b/addons/actions.md
@@ -131,7 +131,7 @@ then
     if (YourItem.state == ON) {
         if (myTimer !== null) {
             logInfo("rules", "Timer rescheduled")
-            timer_heizdecke.reschedule(now.plusMinutes(5))
+            myTimer.reschedule(now.plusMinutes(5))
         } else {
             myTimer = createTimer(now.plusMinutes(5), [ |
                 logInfo("rules", "Timer activated")
--- a/addons/actions.md
+++ b/addons/actions.md
@@ -121,18 +121,38 @@ val output = sendHttpGetRequest("https://example.com/?id=1", headers, 1000)
 
 For example:
 
-```javascript
-var Timer myTimer = createTimer(now.plusMinutes(5), [ |
-    logInfo("rules", "Timer activated")
-])
+```php
+var Timer myTimer = null
+
+rule "timer example"
+when
+    Item YourItem changed
+then
+    if (YourItem.state == ON) {
+        if (myTimer !== null) {
+            logInfo("rules", "Timer rescheduled")
+            timer_heizdecke.reschedule(now.plusMinutes(5))
+        } else {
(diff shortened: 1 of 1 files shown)


addons/actions.md
```diff
@@ -121,18 +121,38 @@ val output = sendHttpGetRequest("https://example.com/?id=1", headers, 1000)
...
+var Timer myTimer = null
+
+rule "timer example"
+when
+    Item YourItem changed
+then
+    if (YourItem.state == ON) {
+        if (myTimer !== null) {
+            logInfo("rules", "Timer rescheduled")
+            timer_heizdecke.reschedule(now.plusMinutes(5))
```
Confectrician, line 134: I think you forgot your **heizdecke** one time. 😛

Confectrician (commented):
Thanks Felix 👍

One small comment below.

I will fix the sign-off error before merging.
Also don't worry about the failing preview build. That is in our responibility.

Trysupe:
> One small comment below.
> 
> I will fix the sign-off error before merging.
> Also don't worry about the failing preview build. That is in our responibility.

I know. I didn't find out on how to sign the commits in the web editor. I couldn't clone the repo and do it locally since my vscode has many extension that change all lines (removing trailing spaces, etc...)

I will fix the "Heizdecke" later this evening

Confectrician:
>I didn't find out on how to sign the commits in the web editor.

The statement should be placed in the last line (description input) for each commit.



Reference:
https://community.openhab.org/t/dco-check-signing-off-with-github-web-editor-explanation/83330?u=confectrician

I have a [Beeftext-Combo](https://beeftext.org/) for this, when adding small changes in web directly.

Trysupe:
> The statement should be placed in the last line (description input) for each commit.

Okay. Thanks. I'll try later when editing my original commit

Trysupe:
I've removed the `heizdecke`. As far as I can see the signing didn't work. I added my signature in the last (and only) line of the description of the commit...

Confectrician:
No problem, i will override that.

Some more information for explanation.
According to https://github.com/openhab/openhab-docs/pull/1482/checks there area two problems left.

- First commit is not signed off. (That was expected.)
- Second commit is signed off, but the DCO Check expected your github-noreply adress
  This could be caused, when your sign-off email is not connected to your GitHub-Account as far as i know.
  (Could even be possible that it has to be public visible, but i am not sure about that.)

So the next question would be: Do you use the sign-of mail in your github account and/or is it public visible?

Anyways for me that's enough, because we will have a proper sign-off statement in the merge commit. 🙂

Confectrician (approved):
Thanks 👍

PR #1485, opened 2021-02-06
Fixed the item type of the timestamp profiles and moved it next to the other State profiles

kaikreuzer:


Confectrician (approved):
Thanks 👍

Confectrician:
Generally i very dislike this table for a longer time.
To me it just brings a expanded list of what is available, but mostly i don't get the information on how to use it properly.

I have already thought about separating the table into topic related tabbed parts with simple examples.
But that's not topic for this change. Maybe i can come up with a better suggestion.
Profiles are really powerful and in my view not advertised/explained well and enough yet.

kaikreuzer:
Yes, I fully agree that this whole piece should be moved out to a dedicated section.
The full power of the profiles isn't yet even covered in the UI either, see https://github.com/openhab/openhab-webui/issues/886, https://github.com/openhab/openhab-webui/issues/884 and https://github.com/openhab/openhab-webui/issues/885.

Confectrician:
Moved to #1486 to keep track of it.

PR #1489, opened 2021-02-13
Update security.md

Urmel:


Urmel:
In place edit - sorry for forgetting to sign off. Corrected in https://github.com/openhab/openhab-docs/pull/1491.

PR #1490, opened 2021-02-13
Note on cert exchange in security.md

Urmel:

Signed-off-by: Alexander Behring <alex@behrings.eu>


Urmel:
My normal sign off seems not to be accepted by this project. Corrected in https://github.com/openhab/openhab-docs/pull/1491.

PR #1491, opened 2021-02-13
Note on cert exchange in security.md

Urmel:

Signed-off-by: Alexander Behring github.com@alexander-behring.eu

### DIFF
--- a/installation/security.md
+++ b/installation/security.md
@@ -33,6 +33,8 @@ On the very first start, openHAB generates a personal (self-signed, 256-bit ECC)
 This process makes sure that every installation has an individual certificate, so that nobody else can falsely mimic your server.
 Note that on slow hardware, this certificate generation can take up to several minutes, so be patient on a first start - it is all for your own security.
 
+If you wish, you can import your own certificate into this keystore. Please ensure that you remove the old certificate and give the new certificate the same alias as the old one (otherwise, the App might still be presented the old certificate).
+
 ## Authentication and Access Control
 
 openHAB does not (yet) support restricting access through HTTP(S) for certain users - there is no authentication in place, nor is there a limitation of functionality or information that different users can access.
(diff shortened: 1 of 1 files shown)


Urmel:
Unclear, why sign off is not detected. Its there - at least it looks like it - and I adjusted it to the value expected by the check.

Urmel:
The other netlify check errors seem not to relate to this commit, but to more general errors, esp. regarding:
Vue packages version mismatch:
12:55:44 PM: - vue@2.6.11
12:55:44 PM: - vue-server-renderer@2.6.12

Confectrician:
Hey @Urmel 

From a quick look it seems that DCO Check expects the angle brackets around the email address, which seem to be missing in your commit.

Example:
`Signed-off-by: Jerome Luckenbach <github@luckenba.ch>`

But no need to do anything, i will override the failed check.
The Netlify preview is a bit buggy last weeks (which is a bit annoying to me) and this has indeed no root in your specific changes.

installation/security.md
```diff
@@ -33,6 +33,8 @@ On the very first start, openHAB generates a personal (self-signed, 256-bit ECC)
 This process makes sure that every installation has an individual certificate, so that nobody else can falsely mimic your server.
 Note that on slow hardware, this certificate generation can take up to several minutes, so be patient on a first start - it is all for your own security.
 
+If you wish, you can import your own certificate into this keystore. Please ensure that you remove the old certificate and give the new certificate the same alias as the old one (otherwise, the App might still be presented the old certificate).
```
Confectrician, line 36: Please use one line per sentence for better tracking.
The website build will take care of a proper presentation.

Confectrician (commented):
Hey @Urmel thanks for contributing here. 👍

Just one small comment below.
DCO check will still fail, even with a correctly signed off second commit. Don't worry about that. I will take care of it.

Confectrician (approved):
Thanks 👍

Urmel:
> Thanks 👍

Thank you for quick response and maintainig this!
Danke.

PR #1492, opened 2021-02-16
[docs] Cover naming of Channel ID

Skinah:
Make the Java naming convention more clear, as 100% of the new binding submissions that I have looked in the last week have all failed to know this. In most cases it is the Channel ID that they get wrong as:

1. Some merged bindings do it "differently".
2. The linked url does not cover what to do for channel IDs.

Signed-off-by: Matthew Skinner <matt@pcmus.com>


Confectrician (approved):
Thanks 👍

PR #1498, opened 2021-02-25
Additional Information

Tim-Ayling:
After line 103 it would appear many (including myself) have had to add a line indicating the location of the JAVA installation.

This is not too difficult to find through the excellent Forum contributions, but it would be nice to see it here instead of having the Service installation fail, then having to look elsewhere for the answer.

In my case the new line 104 is 'set.default.JAVA_HOME=C:\Program Files\Zulu\zulu-11', followed by a new blank line.

### DIFF
--- a/installation/windows.md
+++ b/installation/windows.md
@@ -101,6 +101,7 @@ By installing the openHAB process as a service in Windows, you can:
     set.default.KARAF_DATA=%OPENHAB_USERDATA%
     set.default.KARAF_ETC=%OPENHAB_USERDATA%\etc
     set.default.PATH=%PATH%;%KARAF_BASE%\lib;%KARAF_HOME%\lib
+    set.default.JAVA_HOME=C:\Program Files\Zulu\zulu-11
 
     # Java Application
     wrapper.working.dir=%KARAF_BASE%


mdeanda:
I don't run windows but wouldn't windows just read the JAVA_HOME value from the path for the service as well if its defined as a system variable (vs user)? Did you confirm this was needed even after a reboot?

Tim-Ayling:
Hi Miguel,

I rebooted several times – but couldn’t be specific as to when.

However, I have just examined my system using a command prompt.
Windows only maintains one list of environment variables (and that is still a throwback to MSDOS) – PATH.
‘C:\program files\Zulu\zulu-11’ is on the path, which must be how the package self-installs on Windows.
There is no reference to JAVA in the path, which would explain why the additional line is needed in the .conf file.

Kind regards

Tim Ayling

Confectrician:
I see one problem here.
Generally it would be fine to document an explicitly set up java home location,
but an unexperienced user will maybe not use zulu and still copy this config, so it will end up in an error afterwards.

Tim-Ayling:
Hi Jerome, 

I totally agree, which is why I made a point of saying “My line....”

My point was purely that mention could be made that if  ‘Error 1067’ occurred upon installing the (working) system as a service, this is a likely cause.

For the purposes of my last reply, I interrogated my PATH environment variable. I don’t think I’ve done that for about 7 years.

I realise that the documentation does its best to be generic, but an awful lot of the documentation is geared to the Linux fraternity, who are used to THEIR flavour of Linux running THEIR personal flavour of sub-systems. This appears very geekish to Windows users.

I started with DOS and am happy with a DOS (command) prompt and have been coding for nearly 50 years. The vast majority of current Windows users have no clue what that is, yet alone the appalling (  ) concept of actually typing in commands! They certainly are not aware of the PATH variable – it’s just not needed by Windows anymore.

Windows install packages interrogate the system to see if there is a Java build present. If so, adjust internal pointers to match it, or download their writers choice. Better yet, we are now seeing packages like tMM that have been re-written to contain their own engine removing all external requirements.

Just sayin’ if you want a broader take up in the Windows arena, you might try to make it a bit easier for them to at least get started.

Regards

Tim

Confectrician:
We should then point to a possible error with a generic example on how to add this config instead of just adding one uncommented line in the example.

Tim-Ayling:
That would be sufficient.

Perhaps even a wiki-type page dedicated to possible causes for each ‘Error xxx’ that may occur.

Then the documentation could be directed to that page stating possible reasons for ‘Error xxx’.

That page would then be under your control, containing the distilled knowledge gleaned from the trial-and-error discussions of the forum.

Confectrician:
@Tim-Ayling are you going to work on this?

I am not doing anything here currently, since i am not able to proof or test it, because i am not running a windows system.
Maybe i got your last message wrong.

Tim-Ayling:
I'm sorry if that was the impression I gave.

I have absolutely no idea where to begin or to supplement the help pages I'm afraid. If they were written, I would happily vet them and suggest alterations amendments.

That may take some time, however, because I have not yet settled on which HA system I want to use. None of them are Windows based, and thus pose hurdles for the non-Linux user.

The main problem I have with openHAB at the moment is that the only Windows compatible Zigbee device is the Conbee II. This connects via the deConz application. The deConz Web App no longer works to open the network for new device pairing since the last f/w update... and I'm not throwing away all my Xiaomi sensors &  switches.

Confectrician:
I think 3 or 4 sentences in thisarticle would be enough.

something like

"you may get this error <error message here"
"this has the following reason(s)"
"this is how you can solve it with the example java installation"
"it can be different for other java instalaltions"

I think those would already give plenty of infomration without focussing too much on one java provide which was the case first.

Do you need any help for the docs part and how to do changes there?
If you tell me how you made the current change i am happy to give some information on how to proceed technically with the docs. :)

Confectrician:
Closing for now. If you want to get back to this, we can reopen and i can help out with hints.
Can't write anything useful to the topic as stated above.

PR #1513, opened 2021-03-16
Update transformations.md

Rossko57:
Small correction to replace formatter in Item label example, from %.1f to %s
Transform returns a string, not possible to format with %.1f

Signed-off-by: Ross Kennedy rossko@culzean.clara.co.uk


Confectrician (approved):
Thanks 👍

PR #1515, opened 2021-03-20
Update logging.md to correct the log:set statement in the debug example

WolfgangSn:
Both, the text description and the example output talk about logging level debug for the heating rule but the statement to set the log level used level info instead of level debug.
To match the description and the example output it has to be **log:set debug org.openhab.core.model.script.heating**

Signed-off-by: Wolfgang Schlueschen github@ncc1701d.33mail.com


Confectrician (approved):
Thanks 👍

Makes sense. :)

PR #1522, opened 2021-03-22
Added emphasis that not all Items and Groups belong in the model and how to model Items outside the model

rkoshak:
There is a pervasive misunderstanding about how to use the model and the limitations of the model. I added some text to emphasize that not everything belongs in the model, restrictions of the model, and some instructions for how to manually manage Items and Groups. 

Signed-off-by: Rich Koshak <rlkoshak@gmail.com>


ghys:
I'll take this one ;)

Graet changes, but I wonder if we're not putting too much in the "getting started" tutorial which IMHO shouldn't overwhelm newcomers with too much information, when some details could perhaps be better suited to the general docs?

But as this is all valuable information, so no need to debate too much and not have it published.

rkoshak:
I agree that the tutorial is getting long, but I'm mainly letting the questions that come up over and over on the forum drive new contributions to the existing pages. This one comes up so often I think it must be in the tutorial. Other stuff can be taken out as the reference docs are filled in.

PR #1538, opened 2021-04-16
[lcn] Add HiDPI logo

fwolter:
Is there a recommended size for addon logos?

Signed-off-by: Fabian Wolter <github@fabian-wolter.de>


kaikreuzer:
> Is there a recommended size for addon logos?

Nothing documented, really. The original logos were scaled to a height of 100px (with variable width), with transparent background (where possible) and the size should not exceed 20KB.

With this in mind, I'd say that the existing LCN logo is just fine. On which page did you feel that there should be a HiDPI version of it?

fwolter:
The current logo looks blurry on my 4K display on https://www.openhab.org/addons/ and on https://www.openhab.org/.

The logo on the addons page is scaled to 182x91px, which would need a logo of twice the width to look crispy on a 4K display with 200% scale. I guess many 4K users use a scale between 125% and 200%. That's why I chose a width of 364px for the HiDPI logo.

The velbus logo for example has 400x228px, which looks crispy on both sites.

I just noticed that the logo is also displayed in the readme. It's scaled to a width of 210px there. So, to look crispy there, too, it should have a width of 420px, IMHO.

kaikreuzer:
We could go for 512x512 pixel size (or smaller in one dimension, if it's not a square logo).
I would limit the max allowed size to 30kB nonetheless. Wdyt?

fwolter:
I'm okay with it. Took me some time to figure out how to produce such small PNG file sizes. I'll document it.

kaikreuzer (approved):
Excellent, thank you!

PR #1540, opened 2021-04-18
Add logo section to dev documentation

fwolter:
I'd like to add a sentence about licensing. Do one need to ask the company for permission adding the logo or is that covered by the LICENSE file?

### DIFF
--- a/developers/bindings/index.md
+++ b/developers/bindings/index.md
@@ -1055,7 +1055,7 @@ These are the requirements for logos:
 File size is key as the website displays hundreds of small logos on the same page.
 To shrink the file size, save your logo with Palette-Based Colors (sometimes called "Indexed-RGBA").
 Also, JPEG compression artifacts from prior conversions or halo around the logo increases file size dramatically.
-There are online converters to convert your True Color PNG logo to Palette-Based Colors. E.g. https://compresspng.com/.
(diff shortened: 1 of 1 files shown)


Confectrician:
From my view we should at least add a disclaimer for this.
Most of the bigger vendors have guidelines for this.

I have left out the pixometer logo, when adding my binding, because i was not able to get a valid statement from the vendor.

Confectrician:
Hey @fwolter maybe we should adapt the requirements, when @ecdye finished the svg icon tests.
Currently it is lookiung pretty good and we may can provide svg icons too, so no need to limit it to png files.

fwolter:
I added the SVG format and also @ecdye's zopflipng command.

Confectrician:
Currently we have a limit to png files indeed, but i am not yet sure if this was for a reason.
I am trying to evaluate this and then finalize here.

Confectrician:
Hey @fwolter i am a bit behind my plans currently and svg won't be possible that fast.
I would suggest that we remove the svg part again for now and then bring this into the "next" and current version docs.

I think we are holding back too much useful information just for the svg part.

Do you agree?

fwolter:
Yes, sure. I thought it was already possible. I removed the SVG reference.

Confectrician (approved):
Thanks 👍

Issue #1040, opened 2019-09-01
Compress images to improve loading times


mueller-ma:
You might want to set up[ this bot](https://github.com/marketplace/imgbot) with a [schedule ](https://imgbot.net/docs/#schedule) to compress all images in this repo to improve loading times of the website.

Confectrician:
Looks fine for me. :)

If there are no doubts from @kaikreuzer or @ghys i would add it here.

PR #1544, opened 2021-04-19
[OmniLink] Add HAI by Leviton logo

ecdye:


fwolter:
You could reduce the size of the logo by -68% when using palette-based colors. See https://compresspng.com/

Confectrician:
We can also add imgbot here on midterm (#1040) to do improve this automatically from time to time for all images.

Confectrician (approved):
Thanks 👍

ecdye:
> You could reduce the size of the logo by -68% when using palette-based colors. See https://compresspng.com/

My thought on this subject is that image compression is good to an extent, I don't think that it is necessary in most cases to go to great lengths to compress an image super tiny when it is already fairly small. In the long run it is not going to make nearly enough difference to bend ourselves over backwards to try and improve image compression. 

ImgBot should do just fine.

PR #1557, opened 2021-04-27
Update macOS TTS logo with svg source

ecdye:
This is an experimental PR that will test if using an svg source for addon logos will work properly with the website. It may need to be reverted if the outcome is not correct.


Confectrician:
>This is an experimental PR

So should we add it and see what the outcome next website build has?
Checks  looks good from my side and ci is working currntly.

ecdye:
That was my thought, I don't think it should harm anything.

Confectrician (approved):

ecdye:
@Confectrician It seems to not work, it has reverted to the default logo on the latest website build. Any ideas as to why?

Confectrician:
Not without looking into it. I will have a look what the build is producing finally.

Confectrician:
I am not sure but probably it is because of this file here (cc @ghys  for conformation)

https://github.com/openhab/website/blob/main/.vuepress/components/AddonLogo.vue

This seems to just add an image tag with a proper image source.
That won't work for svg and we may have to adapt the vue component to differentiate between svg and file based image formats.

Confectrician:
And another find:
https://github.com/openhab/openhab-docs/blob/final/process_addons.groovy#L63

This script processes all addon readmes and provides frontmatter texts.
The highlighted line checks for png logos so that is another reason for the svg to fail.

kaikreuzer:
@Confectrician It is not about an svg, but about a json file that is simply referenced for download on the README. It is not an image that should be rendered on the page.

Confectrician:
I think you are mixing up some issues currently?!

This one here is definetely about an svg logo we are trying to get to work in the website.

kaikreuzer:
Oops, indeed, sorry! I'll just delete my comment again...

PR #1555, opened 2021-04-27
Optimize all png images using zopflipng

ecdye:
The command
``` bash
for i in $(find . -name '*.png'); do zopflipng -m --filters=0me --lossy_8bit --lossy_transparent -y "$i" "$i"; done
```
was used to accomplish this optimization, all images should have lost no
accuracy as a result of the compression applied.

Signed-off-by: Ethan Dye <mrtops03@gmail.com>

PR #1559, opened 2021-04-29
Revert zopflipng on openHABian images

ecdye:


ecdye:
Actually, @ghys these images should be all migrated from the openHABian repo, so shouldn't we delete them?

Confectrician:
>Actually, @ghys these images should be all migrated from the openHABian repo, so shouldn't we delete them?

Yes. See: https://github.com/openhab/openhab-docs/pull/1555#issuecomment-830070579

We have to do one manual merge anyways afterwards.
But we could at least try to get the build running again for now with the reverted images.

Confectrician (approved):
Thanks 👍

Let's see if it willl do the trick for the build now. (Temporary.)

PR #1561, opened 2021-05-02
Improve sinks documentation

wborn:
* Remove references to the `javasound` sink since it no longer exists since OH 2.5
* The only UI supporting the `webaudio` sink is currently HABPanel

---

See:
* https://github.com/openhab/openhab-webui/issues/743
* [OH 3 Audio from rule using webaudio](https://community.openhab.org/t/oh-3-audio-from-rule-using-webaudio/121602?u=wborn)


wborn:
I also noticed that the [Speech-to-Text section](https://www.openhab.org/docs/configuration/multimedia.html#speech-to-text) mentions an Android App issue which is closed (fixed?).

> The Android app is [not yet adapted to the openHAB 2 voice features](https://github.com/openhab/openhab-android/issues/242).
> To still use the voice feature you will need to have a String item with the name "VoiceCommand" defined, to which the app sends any recognized text as a string command.


Do you know if those lines can be removed or should be updated @openhab/android-maintainers?

Confectrician (approved):
Thanks 👍

mueller-ma:
@wborn This info is quite outdated. I proposed a PR to remove it: #1562

PR #1562, opened 2021-05-02
Remove outdated info about voice recognition in Android app

mueller-ma:
This has been fixed in 2018 already.

Signed-off-by: mueller-ma <mueller-ma@users.noreply.github.com>


Confectrician (approved):
Thanks for the catch. 👍

PR #1573, opened 2021-05-15
Added note and example of new Timestamp Offset Profile

cweitkamp:
- Added note and example of new Timestamp Offset Profile

See https://github.com/openhab/openhab-core/pull/2351


Confectrician (approved):
Thanks 👍

PR #1578, opened 2021-05-18
Add cherrypick action

Confectrician:
This will add an action, that should help us spread contents over several branches, when a change is relevant to more than one docs version.
It will auto-generate pull requests to the relevant branches, to ease up the workflow.


ghys:
I think you forgot this part for the push to work:
```
      env:
        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```

Confectrician:
I understood the action docs as it would default to GITHUB_TOKEN if nothing is mentioned,
since GITHUB_TOKEN is a standard env variable.

But ofc its worth a try^^

Confectrician:
Yep it did the trick. I really thought it is available in standard nowadays. 🤷‍♂️
I will add a Pull Request to the action repo to enhance the example.

PR #1581, opened 2021-05-20
Tabberize docker article a bit.

Confectrician:
Provides tabbed examples for direct usage with Linux and Windows based installation.
There are still OS based explanations left in the article.

### DIFF
--- a/installation/docker.md
+++ b/installation/docker.md
@@ -123,7 +123,6 @@ docker run ^
 
 ::::
 
-
 Where
 
 - `<uid>` is the user ID number for the `openhab` user which you can obtain using the command `id openhab`,
@@ -259,7 +258,6 @@ If you want use an USB stick (for example for Z-Wave network), then it will be n
 In Docker openHAB is running in name of `openhab`, a restricted user.
 The stick will work if you run the following command right after docker image is started.
 
-
 :::: tabs
 
 ::: tab Linux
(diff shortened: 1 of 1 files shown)


M123-dev:
Looks pretty good already, but in the Windows install code we should add `-p 8080` and `-p 8443` to expose the correct ports. 

The question is if we leave the linux user creation area in and if not, how we display it.

Confectrician:
We can cut out complete parts including Text and example code boxes into tabs, that should be no problem.

Can you provide a Pull request with your input to my fork and PR branch?

M123-dev:
@Confectrician, I just checked a few thing, I dont think the docker image is setup to be runnable on windows machines, I'll look at it more closely and then maybe open an issue again to improve the documenatation about it. 
Until then, the pr is so in any case sufficient. If someone has desire to deal with the current problems, he may be happy about the ^ code variant.

Confectrician:
Ok I will move the user creation into the Linux tab then and merge this one. :)

Confectrician:
I have checked the article again and will provide a little Note box, that some contents may only apply to linux usage.

For moving the user part, we would have to refactor many parts.
That should make more sense in a complete rework with your findings.

rkoshak:
I spent some time trying to help someone get it to work in Docker. And it *can* work but Windows doesn't support passing through any USB devices which eliminates a whole host of technologies from support. On Windows one is far better off running OH natively or in a VM instead of Docker.

Confectrician:
Maybe we should then add a general notice to the article?!

PR #1599, opened 2021-06-07
Update jsr223.md

hypetsch:
Remove HTML escaping


Confectrician (approved):
Thanks 👍

Good catch.

PR #1603, opened 2021-06-17
fix: correct mvn command

hypetsch:
Corrects a typo in mvn command.


Confectrician (approved):
Thanks for the catch. 👍

PR #1602, opened 2021-06-14
Remove requirement for compact profile 2 as those do not exist anymore for Java 11

kaikreuzer:
Signed-off-by: Kai Kreuzer <kai@openhab.org>

PR #1614, opened 2021-07-06
Add default libraries for SSE

BjoernLange:
During the review of the Miele Cloud Binding I had a [discussion with @kaikreuzer regarding the use of JAX-RS as a library for SSE](https://github.com/openhab/openhab-addons/pull/9146#issuecomment-848951048). It seems to be common practice to use JAX-RS for SSE within openHAB and as the restriction to [compact profile 2 has been lifted](https://github.com/openhab/openhab-docs/pull/1602) I think this should be added here.

Signed-off-by: Björn Lange <bjoern.lange@tu-dortmund.de>


kaikreuzer (approved):
Thanks for this addition, @BjoernLange!

PR #1622, opened 2021-07-26
Ensure that UTF-8 is used as default for all text files on Windows

PatrikG8:
The same as https://github.com/openhab/openhab-docs/pull/1173


Confectrician (approved):
Thanks 👍

No sign-off needed due to small-patch-exception.

PR #1626, opened 2021-08-04
Spelling. Loops notation

fringefilmsoz:
Corrected error in spelling of fragments (from fragements).

Changed notation of state of item in a loop to: items[loop.item.name].state (was loop.item.state). 
This will ensure items update state with state changes.


Confectrician:
Hi,

The signutare add commit seems to have made a mess.
I will see if i can fix that myself. But this will need some days because i am abroad currently.

Confectrician (approved):

PR #1629, opened 2021-08-09
Update expire examples

bredmich:
Added a description that the expire parameter can be combined with items that are connected to a channel.
Added an example to show the correct syntax when combining channels and expire.

Signed-off-by: Michael Bredehorn michael@bredehorn.nrw (github: bredmich)

### DIFF
--- a/configuration/items.md
+++ b/configuration/items.md
@@ -631,7 +631,7 @@ The first example shows a symbiosis of the LG webOS Binding and the Wake-on-LAN 
 
 While the `channel` parameter is used to link an item to a channel of a thing, it is possible to add further parameters for additional features.
 Parameters are provided as a comma separated list.
-The order of parameters doesn´t matter and is interchangeable.
+The order of the parameters does not matter.
   
 {: #autoupdate}
   
--- a/configuration/items.md
+++ b/configuration/items.md
@@ -629,7 +629,7 @@ The first example shows a symbiosis of the LG webOS Binding and the Wake-on-LAN 
 
 #### Parameters
 
-While the `channel` parameter is used to link an item to a channel of a thing, it´s possible to add additional parameters for more features.
+While the `channel` parameter is used to link an item to a channel of a thing, it is possible to add further parameters for additional features.
 Multiple parameters can be divided by a comma and a space.
 The order of parameters doesn´t matter and is interchangeable.
   
--- a/configuration/items.md
+++ b/configuration/items.md
@@ -630,7 +630,7 @@ The first example shows a symbiosis of the LG webOS Binding and the Wake-on-LAN 
 #### Parameters
 
 While the `channel` parameter is used to link an item to a channel of a thing, it is possible to add further parameters for additional features.
-Multiple parameters can be divided by a comma and a space.
+Parameters are provided as a comma separated list.
 The order of parameters doesn´t matter and is interchangeable.
   
 {: #autoupdate}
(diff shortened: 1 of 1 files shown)


configuration/items.md
```diff
@@ -649,7 +649,8 @@ This parameter allows to post an update or command to an item after a period of
 
 The expiration timer is started or restarted every time an item receives an update or a command *other than* the specified "expire" update/command.
 Any future expiring update or command is cancelled, if the item receives an update or command that matches the "expire" update/command.
-
+The expire parameter can also be combined with items that are connected to a channel.
```
kaikreuzer, line 652: I'd say that this is the normal case and thus should not be specifically mentioned.

configuration/items.md
```diff
@@ -669,6 +670,7 @@ Player MyPlayer   { expire="1h,command=STOP" } // send STOP command after one ho
 Number MyChannel  { expire="5m,state=0" }      // update state to 0 after five minutes
 String MyMessage  { expire="3m12s,Hello" }     // update state to Hello after three minutes and 12 seconds
 Switch MySwitch   { expire="2h" }              // update state to Undefined two hours after last value
+Switch MyTrigger  { channel="openhab:speedtest:binding:trigger_test", expire="5s,command=OFF" } // update state to OFF after 5 seconds
```
kaikreuzer, line 673: We should maybe follow the [same syntax as above](https://github.com/openhab/openhab-docs/pull/1629/files#diff-f35476973205a2c93d83050c5d645b4db05826b0aeb9eafe3fa355c2c41988d6R641) with the `autoupdate` feature.
Instead of adding another Switch, I'd suggest to hence simply add `channel="xxx", ` to the three entries above.

kaikreuzer (changes requested):

configuration/items.md
bredmich: Agreed, i followed the example of `Text File Linking` with real values.

configuration/items.md
bredmich: Maybe it´s not the best way to describe it.
I think it´s worth to mention this possible combination of channels and parameters.
Could it be an idea to add a general **Parameters** description above **Parameter `autoupdate`** and **Parameter `expire`**?
Like the general description of **Tags**.

kaikreuzer: Yes, I guess it should go somewhere to the [binding configuration](https://next.openhab.org/docs/configuration/items.html#binding-configuration) section - after all, the `channel` and `autoupdate` and `expire` parameters are provided all the same way. Explaining this and thus introducing the additional parameters in a general Parameters section above `autoupdate` definitely makes sense!

bredmich:
I added a general `Parameters` section above `autoupdate` and `expire`.
I´m not 100% satisfied with the wording.
Maybe someone can have a look.

configuration/items.md
```diff
@@ -643,13 +651,14 @@ Switch Garage_Gate {channel="xxx", autoupdate="false"}
...
 
-#### Parameter `expire`
+##### Parameter `expire`
 
 This parameter allows to post an update or command to an item after a period of time has passed.
 
 The expiration timer is started or restarted every time an item receives an update or a command *other than* the specified "expire" update/command.
 Any future expiring update or command is cancelled, if the item receives an update or command that matches the "expire" update/command.
-
+The expire parameter can also be combined with items that are connected to a channel.
```
kaikreuzer, line 660: With the above section added, this can imho be removed.
```suggestion
```

kaikreuzer (changes requested):
Thanks @bredmich, I made some small text change suggestions.

kaikreuzer (approved):
Thanks!

PR #1630, opened 2021-08-09
provide additional example to highlight transform

mtraxiot:
transform must be string type addition example to highlight this


Confectrician (approved):
Thanks 👍

PR #1644, opened 2021-09-12
Fix copy&paste typo in example code

juergencodes:


kaikreuzer (approved):
Thanks!
(merging as small patch exception without DCO)

PR #1648, opened 2021-09-22
Mikrotik binding logo

duhast:
Mikrotik binding logo as a followup for https://github.com/openhab/openhab-addons/pull/10014#pullrequestreview-760062104


duhast:
I can make it even smaller. What is the optimal width in pixels?

ghys (approved):
I think it's fine - AFAIK there aren't strict requirements for logos yet except "use common sense" (a couple kB at max and reasonable dimensions).

PR #1650, opened 2021-09-25
Update linux.md

skenhall:
The actual command "samba-tool" was missing to run the "testparm" subcommand.


ghys (approved):
Thanks, no DCO required (small patch exception).

PR #1651, opened 2021-09-27
Update "Default Tags" documentation

wborn:
There used to be the following "Default Tags" which are no longer used in the current semantic model:

* Lighting
* Switchable
* CurrentTemperature
* TargetTemperature
* CurrentHumidity

Related to openhab/openhab-addons#11313

---

Let's see how the (already big) system-wide channel types table ends up with yet another column. :thinking:

### DIFF
--- a/developers/bindings/thing-xml.md
+++ b/developers/bindings/thing-xml.md
@@ -123,28 +123,28 @@ The following XML snippet shows a system channel type definition and thing type 
 
 There exist system-wide channel types that are available by default:
 
-| Channel Type ID       | Reference typeId             | Item Type            | Category         | Description                                                                                                                                                                                                             |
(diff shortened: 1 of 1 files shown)


wborn:
> Let's see how the (already big) system-wide channel types table ends up with yet another column. :thinking: 

It turned out pretty well I think.

If we would also add descriptions to [SemanticTags.csv](https://github.com/openhab/openhab-core/blob/main/bundles/org.openhab.core.semantics/model/SemanticTags.csv) we might also be able to generate even nicer docs for the available tags regardless of what is tags are added or removed in the future. :slightly_smiling_face:

ghys (approved):
Thanks!

> If we would also add descriptions to SemanticTags.csv we might also be able to generate even nicer docs for the available tags regardless of what tags are added or removed in the future. 🙂

Agreed; it would be nice to have this list in the docs in the future.

PR #1674, opened 2021-11-11
Add nuvo logo and fix background of oppo logo

mlobstein:
Add nuvo logo and fix background of oppo logo


Confectrician (approved):
Thanks 👍🏼

PR #1676, opened 2021-11-19
Update sitemaps.md

Redsandro:
As a new user, I was frustrated to find this out far too late. I really did read, twice, the page about sitemaps. I didn't understand why I couldn't see the results as in the screenshots. Where is it? Am I crazy?

There is so much documentation, a new user is not going to read every page. They learn, read or hear that they would need a sitemap, so that's what they are going to read in the documentation. They are not going to read a [different page](https://www.openhab.org/docs/ui/) first, where the above warning is present.

I advise all documentation writers to keep, move or duplicate all important information surrounding a subject _on the page about that subject_.

If you don't like the wording, just change it to your liking. You get the global idea. :heart: 
image

### DIFF
--- a/ui/sitemaps.md
+++ b/ui/sitemaps.md
@@ -5,7 +5,11 @@ title: Sitemaps
 
 # Sitemaps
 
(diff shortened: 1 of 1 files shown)


Confectrician:
Hey @Redsandro 

Thanks for your contribution.
I have added a info box around your explanations top make it more visible.
I think we are good to go then.

Confectrician (approved):

Confectrician (approved):

PR #1689, opened 2021-12-05
Add updated 2.5.x openhabian information

### LINKED ISSUE #1608: OH2 install instructions
Just came across #1607 and that reminds me that the openHABian instructions need to be changed for OH2, too.
Essentially in order to get OH2, users need to set `clonebranch=stable` in `openhabian.conf` AFTER flashing but BEFORE you boot, else it'll install openHAB3. 
See https://www.openhab.org/docs/installation/openhabian.html#openhabian-conf how to edit `openhabian.conf`.

Also correct the link in "either by plugging an Ethernet cable or configuring the Wi-Fi, and boot!"
* remove the word "the" from the link text
* remove the dash from wi-fi of the link target so it becomes
https://v2.openhab.org/docs/installation/openhabian.html#wifi-based-setup-notes

Confectrician:
Fixes #1608 

@mstormi
i couldn't find all places you wanted edits in. (For example the wifi link.)
May you check if anything is missing?

### DIFF
--- a/installation/openhabian.md
+++ b/installation/openhabian.md
@@ -149,10 +149,15 @@ All packages will be downloaded in their newest version and configured to work w
 The whole process will take a few minutes, then openHAB and all other needed tools to get started will be ready to use without further configuration steps.
 openHABian is designed as a headless system, you will not need a display or a keyboard.
 
+::: warning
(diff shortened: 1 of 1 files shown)


mstormi:
Assuming that change just applies to OH2 docs I think we're fine with that as the rest applies to all OH versions so the most recent openHABian docs are fine to use.

Confectrician:
Yep. This does only affect the OH2 docs.

PR #1717, opened 2022-01-01
[publictransportswitzerland] Add logo

jeremystucki:
Icon "Zug_r" from https://digital.sbb.ch/en/brand-elements/pictograms

We can use this icon because of the [right of use policy](https://digital.sbb.ch/en/brand-elements/rights-of-use), which states that:
> Use of SBB pictograms, timetable icons and SBB clock:
> - Projects related to public transport in Switzerland.

---

Relevant PR: https://github.com/openhab/openhab-addons/pull/8540


Confectrician:
Thank for providing both png and svg file. 🙂 
SVG coes not yet get processed in our docs but it is good to have it in place already.

Confectrician (approved):

PR #1734, opened 2022-01-13
Added image for IPP binding

cweitkamp:
- Added image for IPP binding


Confectrician:
Should we also bring this in the stable docs @cweitkamp ?

Looks like ipp is a bit longer available as binding already.

cweitkamp:
I do not have any arguments against doing so.

Confectrician:
OK i have prepared it for cherry picking then. :) Thanks.

Confectrician (approved):

PR #1737, opened 2022-01-14
Update persistence.md

digitlength:
Make clear that openHAB ships with RRD4J installed and running by default.
Make clear that the default persistence service is used by charts and rules.


Confectrician (approved):
Thanks. 👍🏼

PR #1741, opened 2022-01-17
Document new "System reached start level XX"-trigger in DSL-Rules

Flole998:
Documents openhab/openhab-core#2658

@kaikreuzer As you requested ;)


Confectrician (approved):
Thanks. 👍🏼

PR #1744, opened 2022-01-28
Update timer example

marcelerkel:
Make sure that `myTimer` is not null when invoking the cancel() method


Confectrician (approved):
Thanks. 👍🏼

PR #1745, opened 2022-01-29
Fixed file name for USB symlinks

RealMalWare:
documentation pointed at etc/udev/rules.d/99-com.rules instead of etc/udev/rules.d/99-usb-serial.rules


Confectrician (approved):
Thanks. 👍🏼

PR #1747, opened 2022-02-01
Update linux.md

digitlength:
Changed 'less' to 'more' as the Testing Release comes out more frequently than the Stable Release just described.


Confectrician (approved):
Thanks. 👍🏼

PR #1748, opened 2022-02-02
Added addon logos

Hilbrand:
Logos for:
- enphase
- feican
- foobot
- melcloud
- spotify

All logo's are resized to 512x512 and compressed using zopflipng.


Confectrician (approved):
Thanks. 👍🏼

PR #1764, opened 2022-02-20
fix inbox list command in documentation

dalori:
Following the current documentation `openhab:inbox` will not list all current inbox entries.
It will show the command usage instead.

```
openhab> openhab:inbox
Usage: openhab:inbox list - lists all current inbox entries
```


Confectrician (approved):
Thanks. 👍🏼

PR #1767, opened 2022-02-20
Add note about textual configuration when implementing a discovery service

fwolter:
There are steadily coming new binding PRs which are not configureable via text files.

CC: @openhab/add-ons-maintainers


Confectrician (approved):
Thanks. 👍🏼

PR #1786, opened 2022-03-19
jsr223.md: obvious inconsistency

dilyanpalauzov:
The original text looks wrong, but the original text might be correct.


Confectrician (approved):
Seems like a copy paste failure.
Thanks for the good catch. 👍🏼

PR #1791, opened 2022-03-31
Clarifying lowercase requirement

ulfwin:
The "-state" part of the custom dynamic icon name must be lowercase. Clarifying this where the state parameter is explained.

See: https://community.openhab.org/t/custom-icon-does-not-respond-to-item-value/134669


Confectrician (approved):
Thanks. 👍🏼

PR #1797, opened 2022-04-13
Fix broken "System State Channel Types" table

wborn:
Looks like the table got broken with #1788.


Confectrician (approved):
Good catch.

Thanks. 👍🏼

PR #1814, opened 2022-05-12
item-type (channel): list all available in core

lolodomo:


Confectrician (approved):
Thanks. 👍🏼

PR #1815, opened 2022-05-12
[Binding] Add logo for evcc

florian-h05:
As the evcc Binding has been recently added in PR https://github.com/openhab/openhab-addons/pull/12611 (merged today), I have added the icon for evcc.

Signed-off-by: Florian Hotze <florianh_dev@icloud.com>


jlaur:
@florian-h05 - thanks. I'm not an expert on graphics, but would it be possible to get a larger version using more of the 512 pixel width available? It seems like the logo itself is much smaller than 512 pixels width, so it will be scaled down and probably appear quite small. As I read the instructions the width requirement is 512 pixels exactly (not maximum), so if you can't find a larger version, I guess there's nothing to do about this since it can't be cropped to a width smaller than 512 pixels. Perhaps someone with more knowledge can help/guide/elaborate.

florian-h05:
@jlaur 

No problem, I can create any size I want to as I have the icon as svg.
I read the 512px as maximum, but I can provide a higher resolution.

florian-h05:
What about 1500 x 671 pixels, file size would be 8,11 kB.

jlaur:
> Oh okay, so I will create a version with a width of 512 pixels but crop of the border around the icon, right?

Exactly, so the logo itself would be 512 in width. At least I'm seeing this:
image

jlaur (approved):
LGTM

Confectrician (approved):
Thanks. 👍🏼

PR #1819, opened 2022-05-15
Add Group, Thing and System triggers to JSR223 docs

jimtng:
These triggers are missing from the current JSR223 docs. Ideally this needs to automatically load from core `bundles/org.openhab.core.automation/src/main/resources/OH-INF/automation/moduletypes`


Confectrician (approved):
Thanks. 👍🏼

Confectrician:
> Ideally this needs to automatically load from core

Please add an issue for this, to keep separate track of the idea.

PR #1820, opened 2022-05-17
[renault] Replace off white background from renault logo with transparency

dougculnane:
The logo looks ugly with the off white background this update should fix it.

Signed-off-by: Doug Culnane <doug@culnane.net>


Confectrician (approved):
Thanks. 👍🏼

PR #1843, opened 2022-06-20
[blockly] finalize blockly docs for M3.3

stefan-hoehn:
- update all documentation and images towards final M3.3 blockly blocks
- add blockly context menu description
- add zoom description

Signed-off-by: Stefan Höhn <mail@stefanhoehn.com>


Confectrician (changes requested):
I have checked all the pictures and that ok so far.

Please remove all markdownlint exception from the files for now.
I want to see where the linter errors occur before blindly accepting dozenz of linter executions.

If an exception makes sense we can add it by hand but i won't accepot that amount blindly.

PR #1845, opened 2022-06-27
Changed apt installation instructions gpg import

rkoshak:
apt-key is deprecated and shouldn't be used nor should the trusted keystore in /etc/apt. Instead the gpg keys should be stored somewhere else and the list file that adds the openHAB repo should point at that one key. Thus, apt will only trust that key for that repo instead of trusting that key even if it's used by some other repo.

There were problems with the redirect from `gpg` which makes the command fail sometimes even when using sudo which is why I save it locally and then move and chmod it after.

Signed-off-by: Richard Koshak rlkoshak@gmail.com

### DIFF
--- a/installation/linux.md
+++ b/installation/linux.md
@@ -69,10 +69,13 @@ Alternatively resort to the [manual installation approach](#manual-installation)
 
 {% include collapsible/body.html %}
 
-First, add the openHAB repository key to your package manager:
+First, add the openHAB repository key to your package manager (note `/usr/share/keyrings` may already exist):
 
 ```shell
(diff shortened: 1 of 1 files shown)


Confectrician:
@rkoshak could you check the installation instroctions on the download page too?

The are located in the website repository, so we may have to change that in parallel.

Confectrician (approved):

rkoshak:
Yes, those need to be updated too. Both the Raspberry Pi and Linux instructions need to be updated. I'll open an issue and submit a PR over there too.

PR #1846, opened 2022-06-27
[JSR223] Mention difference between NashornJS and GraalJS

florian-h05:
Fixes https://github.com/openhab/openhab-js/issues/139.

This explains the difference between the two available JavaScript options and recommends to use JS Scripting.

Signed-off-by: Florian Hotze <florianh_dev@icloud.com>

### DIFF
--- a/configuration/jsr223.md
+++ b/configuration/jsr223.md
@@ -14,7 +14,13 @@ and are comfortable working with the command line prompt of the operating system
 
 [JSR223](https://docs.oracle.com/javase/8/docs/technotes/guides/scripting/) ([spec](https://jcp.org/aboutJava/communityprocess/pr/jsr223/index.html)) is a standard scripting API for Java Virtual Machine (JVM) [languages](https://en.wikipedia.org/wiki/List_of_JVM_languages).
 The JVM languages provide varying levels of support for the JSR223 API and interoperability with the Java runtime.
-Currently the following languages are known to work well for openHAB scripting: [**Apache Groovy**](/addons/automation/groovyscripting/) (JVM scripting language), [**JavaScript**](/addons/automation/jsscripting/), [**Ruby**](/addons/automation/jrubyscripting/), and [**Jython**](https://www.openhab.org/addons/automation/jythonscripting/) (Python on the JVM).
+Currently the following languages are known to work well for openHAB scripting: [**Apache Groovy**](/addons/automation/groovyscripting/) (JVM scripting language), **JavaScript/ECMA - 262 Edition 5.1**, [**JavaScript/ECMAScript 262 Edition 11**](/addons/automation/jsscripting/), [**Ruby**](/addons/automation/jrubyscripting/), and [**Jython**](https://www.openhab.org/addons/automation/jythonscripting/) (Python on the JVM).
+
+Please note that **JavaScript/ECMA - 262 Edition 5.1** (whose setup is described here) is using an older engine called NashornJS, which has been removed from JDK 15 and will be unavailable one day. 
+Moreover, you need to work with the raw openHAB Java APIs in this JavaScript environment which is not easy.
+In contrast to this, [**JavaScript/ECMAScript 262 Edition 11**](/addons/automation/jsscripting/) supports the latest **ES6** features like `let` and `const` by using a new engine called GraalJS. 
+It is recommended to use the [**JavaScript/ECMAScript 262 Edition 11**](/addons/automation/jsscripting/), also known as [JS Scripting](/addons/automation/jsscripting/), 
+because of the [openHAB JavaScript library](https://github.com/openhab/openhab-js/) which provides convenient access to the openHAB APIs just with plain JavaScript - no messing with Java is needed.
 
 Although JSR223 scripts can be used as a general-purpose extension language for openHAB, it is most commonly used for the creation of rules, and within scripted Actions or Conditions.
 Currently, openHAB allows JSR223 scripting to access all packages, which may not be included in the official APIs.
@@ -25,7 +31,7 @@ New APIs are planned to be implemented in the future, which will provide standar
 
 :::: tabs
 
-::: tab JavaScript
+::: tab JavaScript/ECMAScript - 262 Edition 5.1
 
 ```js
 'use strict';


configuration/jsr223.md
```diff
@@ -14,7 +14,13 @@ and are comfortable working with the command line prompt of the operating system
 
 [JSR223](https://docs.oracle.com/javase/8/docs/technotes/guides/scripting/) ([spec](https://jcp.org/aboutJava/communityprocess/pr/jsr223/index.html)) is a standard scripting API for Java Virtual Machine (JVM) [languages](https://en.wikipedia.org/wiki/List_of_JVM_languages).
 The JVM languages provide varying levels of support for the JSR223 API and interoperability with the Java runtime.
-Currently the following languages are known to work well for openHAB scripting: [**Apache Groovy**](/addons/automation/groovyscripting/) (JVM scripting language), [**JavaScript**](/addons/automation/jsscripting/), [**Ruby**](/addons/automation/jrubyscripting/), and [**Jython**](https://www.openhab.org/addons/automation/jythonscripting/) (Python on the JVM).
+Currently the following languages are known to work well for openHAB scripting: [**Apache Groovy**](/addons/automation/groovyscripting/) (JVM scripting language), **JavaScript/ECMA - 262 Edition 5.1**, [**JavaScript/ECMAScript 262 Edition 11**](/addons/automation/jsscripting/), [**Ruby**](/addons/automation/jrubyscripting/), and [**Jython**](https://www.openhab.org/addons/automation/jythonscripting/) (Python on the JVM).
```
rkoshak, line 17: It's probably worth listing the Jython version here as 2.7. I say that because Python 2.7 is in fact a couple years since end of life and knowing that this is the version supported might give some users pause.

configuration/jsr223.md
```diff
@@ -14,7 +14,13 @@ and are comfortable working with the command line prompt of the operating system
...
+
+Please note that **JavaScript/ECMA - 262 Edition 5.1** (whose setup is described here) is using an older engine called NashornJS, which has been removed from JDK 15 and will be unavailable one day. 
+Moreover, you need to work with the raw openHAB Java APIs in this JavaScript environment which is not easy.
```
rkoshak, line 20: This too is not entirely correct. There is a Helper Library for JS. If the problem is that you have to install that library via means other than JRuby, Jython, and Groovy (which does lack a helper library) also should be listed. 

Over all I think it's premature to take a stand on which language to use over all (if we ever get to that point) but I also think it's important to stere those who want to use JS to use JS Scripting instead of Nashorn.

So maybe rework this section a little bit along the lines of:

> Please note that **JavaScript/ECMA - 262 Edition 5.1** which is referenced here is using an older engine called NashornJS.
> This version of ECMAScript is end of life and lacks many important features that have been added to the language since 2009.
> If you plan to use ECMAScript as your rules language, please use [**JavaScript/ECMAScript 262 Edition 11**](/addons/automation/jsscripting/), also known as [JS Scripting](/addons/automation/jsscripting/) which includes the latest **ES6** language features, supports third party node libraries, and comes with a well documented helper library that makes interacting with all the APIs documented on this page much simpler and straight forward.

I think that makes it a bit more clear that we are not pushing someone who loves Ruby, for example, towards using JS Scripting but instead limiting the advice to those who are drawn to JS in the first place.

rkoshak (commented):
I like this. I added a few comments based on the current state of Nashorn.

Over all, I think our whole approach to the rules documentation needs an overhaul but I think we should wait to see if Rules DSL gets spun out into a separately installable add-on.

configuration/jsr223.md
florian-h05: > This too is not entirely correct. There is a Helper Library for JS. If the problem is that you have to install that library via means other than JRuby, Jython, and Groovy (which does lack a helper library) also should be listed.

Out of interest: Where can I find those helper libraries? Do you refer to https://openhab-scripters.github.io/openhab-helper-libraries/JavaScript/Reference.html?

florian-h05: Okay thanks, I remember having a look at this helper library some time ago. 
One thing that I missed back then was the documentation for that one.

florian-h05:
@rkoshak 
This is now ready for re-review, I also added an example rule for JS Scripting.

rkoshak:
The text looks great!

I'm ambivalent about the JS Scripting example.

On the one hand it's useful and clearly illustrates the differences between them. And there are examples for all the other languages there already (except Rules DSL for obvious reasons).

On the other hand, JS Scripting isn't actually JSR223 so there is a strong argument that an example like this doesn't even belong here, especially since just about everything that is discussed on this page is hidden behind the JS Scripting Helper Library which is documented elsewhere. Were the example to only use the JSR223 interface as it's discussed on this page, it would look almost exactly the same as the Nashorn example with the addition of a `require("@runtime");`.

Given that the topic of this page is the JSR223 runtime interface, any JS Scripting example should probably use that interface directly, perhaps with a note that using the Helper Library results in a much simpler and straight forward implementation (which is also true for jRuby which probably should have a similar note ).

florian-h05:
Reagrding JS Scripting:
What about adding a note to the code example that this example is using the official helper library to demonstrate it's functionality?

florian-h05:
I update the JS Scripting example to have one using the raw Java APIs and to have the same using the helper library (with a comment that clearly indicates that it is using the helper library).

However, I was not able to get the example for the raw APIs working with `@runtime`, I instead used `Java`.

Just look at the docs preview.

Confectrician (commented):
Looks good from docs side.
Please give me a ping, when it is ready for review/merge from the content side too.

rkoshak:
So this didn't work? Or did you try to import some other way? 

I've not done file based rules with JS (either version) so I don't know if there is something different. But when I was looking in how to make Nashorn compatible with JS Scripting, the easiest thing was to pull in `@runtime` and then any call to the OH API classes would go through the `runtime` variable.

```
var runtime = require("@runtime");

runtime.scriptExtension.importPreset("RuleSupport");
runtime.scriptExtension.importPreset("RuleSimple");

var sRule = new runtime.SimpleRule() {
    execute: function( module, input) {
        print("This is a 'hello world!' from a JavaScript rule.");
    }
};

sRule.setTriggers([
    runtime.TriggerBuilder.create()
        .withId("aTimerTrigger")
        .withTypeUID("timer.GenericCronTrigger")
        .withConfiguration(
            new runtime.Configuration({
                "cronExpression": "0 * * * * ?"
            })).build()
    ]);

runtime.automationManager.addRule(sRule);
```

If that or something like that won't work, I'm good with the example as it is. Even if that does work, I'm OK with the example as it's currently written so I leave it up to you if you want to change it to be consistent with the other examples (there is some value in that) or keep it different and therefore showing an alternative way to do it.

configuration/jsr223.md
```diff
@@ -54,6 +60,51 @@ automationManager.addRule(sRule);
 
 :::
 
+::: tab JS&nbsp;Scripting
+
+```js
+const { automationManager } = require('@runtime/RuleSupport');
+const Configuration = Java.type('org.openhab.core.config.core.Configuration');
+const SimpleRule = Java.extend(Java.type('org.openhab.core.automation.module.script.rulesupport.shared.simple.SimpleRule'));
```
digitaldan, line 68: Is there a good reason to show using the low level API's here ?  Should we just show openhab-js examples since thats what ships with it, and new users will likely end up here and be confused ?

rkoshak: Because this page is explicitly documenting the low level API and it's confusing to show an example that doesn't use the low level API on this page. If using just openhab-js there would be nothing on this page explaining how that example works and conversely there is nothing that the example would be demonstrating that is documented on this page.

digitaldan:
Sorry i'm late to the party here, but i  wonder if we want to keep this page about "JSR223", or rather just make it "Scripting" and add in the rules DSL to it.  As a user i would think its super confusing why its spread around like this across multiple pages, and at the end of the day do they really care about the underlying Java extension , or do they care about the rule languages  (Javascript, ruby, DSL, etc...)  themselves?

rkoshak:
> Sorry i'm late to the party here, but i wonder if we want to keep this page about "JSR223"

Ultimately I think we do need some page to document the actual Java openHAB rules API because there will always be something that a helper library doesn't do or the way it works needs to be worked around for some reason. But it should become an advanced topic and not something that most users ever need to read since they are not using this API directly but instead using the various helper libraries.

However, once we start going down that path we would ultimately end up with a complete reworking of our entire approach for rules documentation in general. I've actually been thinking about this for some weeks now but lack the time to really work on it much.

But were I to start on something like this I'd use the following structure:

1. Rules Concepts: borrow from the Getting Started tutorial perhaps with more detailed discussion, examples cover most of the supported languages, maybe this belongs in the concepts section
    a. triggers
    b. conditions
    c. actions
    d. comprehensive examples using the respective helper libraries (not the raw API)
    e. rule templates
2. Actions: These really shouldn't be their own thing, they should be a part of the rules documentation, maybe it should be part of 1. maybe move the persistence actions here too.
3. UI Rules Reference: Currently not documented anywhere, may be short depending on what's covered in 1. I'd like to see some of the JSON stuff from https://www.openhab.org/docs/developer/module-types/#tie-everything-together-define-rules
4. Blockly: as currently written
5. Rules DSL: modified to be more of a reference than explaining rules concepts, goes away when/if this becomes and add-on since it will be part of the add-on's docs
6. openHAB's raw API: this page with some additions, (e.g. the `event` Object is missing)
7. Writing Rule Templates: new content

We need to extract the generic stuff from the current Rules page and move the Rules DSL stuff to its own page. We need to push people to use their respective helper libraries instead of the RAW API. Over all, the Rules docs need to look more like the UI section in the docs. 

> at the end of the day do they really care about the underlying Java extension , or do they care about the rule languages (Javascript, ruby, DSL, etc...) themselves?

Several of the languages including Groovy and JRules *only* work through the raw API. Other languages have a third party helper library (i.e. not part of the openHAB project itself) including Jython, Nashorn, and jRuby and users may not want to or be able to install these and instead would rather use the language without the helper library. And not all the helper libraries do as good of a job abstracting the raw API.

Anyway, for this specific PR though, since we are not looking to do a complete reworking of the rules docs and not reworking this specific page, I strongly believe that if we include a JS Scripting example at all, it should use the raw API instead of the helper library. Otherwise it's kind of a pointless example given what this page is about.

When, if the rules docs do get reworked, then there will be lots of other places to include "proper" examples using openhab-js, not to mention the examples in the JS Scripting add-on itself.

florian-h05:
@rkoshak

> I think that wouldn't be sufficient because I think the presets get imported to the runtime, not the wider script so stuff like SimpleRule and TriggerBuilder wouldn't be available. But that's just an assumption on my part, I've not tried it.

```javascript
Object.assign(this, require('@runtime'));
Object.assign(this, require('@runtime/RuleSimple'));
Object.assign(this, require('@runtime/RuleSupport'));
```
… imports everything needed for the creation of a rule, but I get the following error:
```
TypeError: invokeMember (create) on org.openhab.core.automation.util.TriggerBuilder failed due to: Unknown identifier: create
```
So `TriggerBuilder` is imported, but it somehow cannot be used (according to the JavaDoc there is a `create()` method.
When I import `TriggerBuilder` with 
```javascript
const TriggerBuilder = Java.extend(Java.type('org.openhab.core.automation.util.TriggerBuilder'));
```
everything is working.

FYI: Using `scriptExtension.importPreset(…)` is not possible as `scriptExtension` is not available on `@runtime`, so your example did not work.

I updated the example to use less `Java.type`.

florian-h05:
@rkoshak 
Regarding your approach of how to rework the rules docs: Looks really good to me, maybe you can create an issue with your post above as I would focus on getting this PR merged.

And can you please have a look at my current example? If it‘s good, we are ready to merge.

rkoshak:
Looks good to me. Thanks for doing this. @Confectrician this looks ready to merge and I'm going to open a new issue to talk about reworking the rules documentation.

Confectrician (approved):
Thanks. 👍🏼

PR #1857, opened 2022-07-06
Fix typo of filename for openhab public key

realbigflo:
The filename /usr/share/keyrings/openhab.pgp should be used consistently in the documentation about setup


Confectrician (approved):
Thanks. 👍🏼

Issue #1855, opened 2022-07-06
Rework Rules Documentation


rkoshak:
The rules documentation is in sore need of reworking. All of the basic rule concepts documentation is embedded in the Rules DSL reference docs. The JSR223 page is too prominent and incomplete. Some of the add-on rules languages do not have much documentation at all.

I think the Getting Started Tutorial makes an OK first crack at explaining rules and rules concepts but as a tutorial, it's incomplete.

I envision a set up more like the UI section of the docs and propose an outline along these lines. Everything below would be under a `Rules` heading in the sidebar.

Done | Section | Title | Contents | Notes
-|-|-|-|-
 ✅  | 1  | Rules Concepts | Explain how rules work, parts of a rule, etc | Language independent, expand on the Getting Started Page
 ✅  | 1a | What are rules | Even driven, made up of parts, other rules concepts |
 ✅  | 1b | Triggers | Explain triggers and list all the triggers | Most of this content can come from the Rules DSL page
 ✅  | 1c | Conditions | What are they, where are they supported, etc. | New content
 ✅  | 1d | Script Actions | What are they, where are they supported, etc. | Expand on the content in the Getting Started Tutorial
 ✅  | 1e | Implicit Variables | What gets passed to the rule when it's triggered | Probably needs a new name
 ✅  | 1f | Rule Templates | Where to find them, how to install and configure them | Expand on content in the Getting Started Tutorial
 ✅  | 1g | Helper Libraries | What are they, where to find them | 
 ✅  | 1h | Comprehensive Examples | Use tabs and show most of the popular languages | Includes UI and text based, emphasis on use of helper libraries
[ ]  | 2 | Rule Actions | List built in rule actions, describe binding actions | Most of the content already exists, may want to rework order, move persistence actions to here, move the cloud notifications to the add-on's docs? Include examples in multiple languages
[ ]  | 2a | Event Bus | |
[ ]  | 2b | Audio & Voice | |
[ ]  | 2c | Logging | |
[ ]  | 2d | Exec | |
[ ]  | 2e | HTTP | |
[ ]  | 2f | Semantics | |
[ ]  | 2g | Timers | |
[ ]  | 2h | Thing Status
[ ]  | 2i | openHAB Subsystem | |
[ ]  | 2j | Ephemeris | |
[ ]  | 2k | Persistence | | Move the content from the Persistence page to here
[ ]  | 2l | Binding Actions | Where do they come from, how to call them | 
[...]

PR #1860, opened 2022-07-08
[Ready for Review] Reworking Rules docs

rkoshak:
Getting started implementing #1855.

This first commit has some placeholders for the files we'll need but doesn't include any content changes yet.

Signed-off-by: Richard Koshak <rlkoshak@gmail.com>


Confectrician:
Hey @rkoshak,

i know this is a big one and you have many things on your list usually.
Is there any chance that we split the changes/to-do's from #1855 and do a partly merge?

From my view any refactored documentation is a win for us, even if it isn't 100% and includes all asked points. :)
There are already good improvements here, that are laying around for weeks now.

Is a partly change possible or would that have a negetaive effect for your workigns on this?

rkoshak:
We can but even what we have now isn't really ready. If at least like to get the concepts page done before the first merge. Then I think we can treat each page after that as a separate PR. Though if like to continue to track the progress on the same issue if that's ok.

florian-h05:
I agree on @rkoshak ‘s opinion.
Let’s get the concepts page done, and then have the first merge.

Confectrician:
> Though if like to continue to track the progress on the same issue if that's ok.

Yes of course. :)

florian-h05:
@Confectrician 
FYI: The concepts page should be finished soon, and can then be merged.

Rkoshak reviewed my contribution of the concepts page, so there shouldn‘t be need for a super detailed review from your side, I think there are just some styling issues (e.g. the ToC) where we need your help regarding Jekyll.

rkoshak:
I think I broke it. 

I've been trying to fix the premerge checks like linting and such and now it's not running those checks nor is it deploying the preview. I'm not sure what I can do at this point.

rkoshak:
@Confectrician , this is ready for review.

None of the other concepts pages have a TOC but I think it's useful to have it for the rules page. Should we be consistent and remove it (or add one to the other pages) or keep it in even if it's inconsistent? Also, is there a way to set the TOC to only show level 2 headings? I don't think we need to go to level 3 and 4 for this page. I tried changing the levels from 2..4 to just 2 but it's still showing them all.

The MD linting errors are not from our changes (and frankly I think they are false positives).

@florian-h05, It seems we cannot have a space in the title for the tabs in our examples. So I renamed the tabs to UI, DSL, JRuby, and JS.

Also, now that I see it, I wonder if later on we ought to replace the JS examples with rule builder. As it is right now, the JS examples look really wordy compared to even Rules DSL. We can leave it as is for this PR though but I think the JS examples will look more comparable using the builder instead of JSRule.

Finally, @Confectrician, the next PR will create a new "Rules" section in the sidebar (similar to the UIs section). Is there anything else I need to modify beyond the .vuepress/config.js file? It seems I remember needing to edit another file too but I can't figure out what.

Confectrician:
> @Confectrician , this is ready for review.

🎉

> None of the other concepts pages have a TOC but I think it's useful to have it for the rules page. Should we be consistent and remove it (or add one to the other pages) or keep it in even if it's inconsistent? Also, is there a way to set the TOC to only show level 2 headings? I don't think we need to go to level 3 and 4 for this page. I tried changing the levels from 2..4 to just 2 but it's still showing them all.

If you refer to the curly braced configuration you can throw it away anyways.
This is an old way of configuring the TOC, we used when the docs were served via guthub pages.
From what i read we can configure the toc only for the complete website.

> The MD linting errors are not from our changes (and frankly I think they are false positives).

They are correct, but i am already working on #1906 and the should be solved then.


I will need some time for review, but would try to get it done over the weekend.

Confectrician (approved):
Sorry for the delay.
Finally had some time to chek this.
Just som minor things i will solve myself.

PR #1872, opened 2022-08-19
Adapt to core changes (addon.xml)

J-N-K:
Depends on https://github.com/openhab/openhab-core/pull/3050


Confectrician:
> 1800 changed lines are not looking wrong.

Can you check if something went wrong with the commits on your side?

J-N-K:
Looks like line-ending changes. According to https://github.com/openhab/openhab-docs/blob/main/.editorconfig CRLF is expected, what is exactly what I committed. Can you comment what is the correct line-ending (it seems it was LF before, according to my local git client).

Confectrician:
It is ok for me the change the line endings to crlf.
I have no strong opinion on this and the file hasn't been changed in any valuable part of the suggestions.

Maybe it isn't used that often and we missed those files for a long time.
If you are fine with having that much changes in here, that would be ok for me as it is for a reason.

But ofc those changes are a bit misleading to the original intention and pr title.

J-N-K:
I can revert that, not an issue. It wasn‘t even intentional, my IDE just picked up what the repo provided as suggestion. I believe in most other repos we use only LF.

Confectrician:
Then i should check this in egenral and adapt the config to the generic approach, where LF would be fine at all.

In this case reverting would be good for th epr to just reflect the intentional changes.

Confectrician (approved):
Sorry, this went out of my focus for way too long.
I didn't realise, that you already fixed everything with the commits.

I have just one small finding, which i will solve myself.

developers/addon/index.md
```diff
@@ -1,14 +1,29 @@
...
+- `misc` for other integrations like Apple's HomeKit or the openHAB cloud
+- `persistence` for services that can be used to store and read historical data
+- `transform` for add-on that provide a way to manipulate values
+- `ui` for bundles that provide UIs or UI-additions (like widgets)
+- `voice` for add-ons that provide voice services
+
+## Structure of an Add-on
+
+Every add-on needs to define a `addon.xml` file.
+Find more information in the respective [Add-on XML reference](addon-xml.html).
```
Confectrician, line 24: This url will result in a 404.
Probably `addon.html` could be meant, since the file was added, but i am not sure.

```suggestion
Find more information in the respective [Add-on XML reference](addon.html).
```

developers/addon/index.md
```diff
@@ -196,7 +206,7 @@ If a thing will be created for this thing type then its properties will be autom
 Therefore the `org.openhab.core.thing.Thing` interface also provides the `getProperties` operation to retrieve the defined properties.
 In contrast to the `getProperties` operation of the thing type instance the result of the thing´s `getProperties` operation will also contain the properties updated during runtime (cp. the thing handler [documentation](thing-handler.html)).
```
Confectrician, line 207: The link will result in a 404.
Do you simply want to jump to the corresponding doc part of this page?

```suggestion
In contrast to the `getProperties` operation of the thing type instance the result of the thing´s `getProperties` operation will also contain the properties updated during runtime (cp. the thing handler [documentation](#the-thinghandler)).
```

developers/addon/index.md
```diff
@@ -1024,19 +1034,19 @@ This is done by implementing the `getServices` method in your bridge handler:
     }
 ```
 
-## Frequently asked questions / FAQ
+### Frequently asked questions / FAQ
 
 Various binding related questions are answered in our [Binding development FAQ](faq.html).
```
Confectrician, line 1039:
```suggestion
Various binding related questions are answered in our [Add-on development FAQ](faq.html).
```

Confectrician (commented):
Again some small comments/questions, from checking the preview after solving the conflicts.

developers/addon/index.md
J-N-K: I have no idea where this should point to. It's already a 404 in the current version. But probably your suggestion is fine.

J-N-K:
Please make sure that this is only merged after the core-PR has been merged, otherwise we'll have an inconsistent documentation.

Confectrician:
I have added the dependency label and subscribed to the core pr. :)

.vuepress/docs-sidebar.js
```diff
@@ -142,7 +142,7 @@ module.exports = [
     children: [
       ['developer/', 'Overview & Introduction'],
       'developer/guidelines',
-      'developer/bindings/',
```
kaikreuzer, line 145: I would definitely suggest to keep the "bindings" menu in place, just as we have dedicated entries for other add-on types like  automation addons, transformation services, voice add-ons etc.
What would imho make sense here is to add a new menu entry for everything that generally applies to all add-ons and this new page could then link to the different specific add-on pages as well.

kaikreuzer (changes requested):

.vuepress/docs-sidebar.js
J-N-K: The problem is that we don't have content for that page. Duplicating what is written on the general page makes no sense.

kaikreuzer: Well, a few parts could be moved from the binding page to the add-on page, like the description about the addon.xml, the info on how to add a logo for the add-on and similar.

J-N-K: After I did that, there was no content left :-)

kaikreuzer: Hm, I see your point, but I am a bit puzzled, because on https://www.openhab.org/docs/developer/bindings/ I see a lot more content about how to develop a binding. You seem to have moved all the files that contribute to this page to `/developer/addon`.

J-N-K: Sometimes I'm a bit slow... Is this version better?

developers/addons/index.md
```diff
@@ -0,0 +1,43 @@
+---
+layout: developersguide
+title: Add-ons
+---
+
+# Developing an add-on
```
kaikreuzer, line 6:
```suggestion
# Developing an Add-on
```

.vuepress/docs-sidebar.js
kaikreuzer: Yes, thanks - that look better!

kaikreuzer (commented):
Do you have any idea, why Github does not properly show the changes that you made, but rather shows the full files to be changed (e.g. developers/bindings/index.md)?
This makes reviewing hard, but I just trust that there were no irrelevant changes.

developers/addons/index.md
J-N-K: Yes, that's due to an incorrect `.editorconfig` in this repo (it defines CRLF instead of LF which is actually used).

J-N-K: This should be fine now.

kaikreuzer (approved):
Wonderful - that's how it should look like! 👍

PR #1873, opened 2022-08-21
add executeCommandLine in ECMA scripts example

Confectrician:
Follow Up for openhab/openhab-docs#1414


rkoshak:
Oops, I commented over on that closed thread before I saw this. I can move the comment over here if it needs more discussion. Tl;dr, as part of issue #1855 the original issue should get handled in a more universal way.

Confectrician:
Closing for now and leaving the branch here.
We can than fastly reintroduce this if needed and extend it with blockly content.

PR #1874, opened 2022-08-21
Update rules_advanced.md

2delarosa:
line 140 fails if actions. is not prepended


rkoshak:
Good catch. LGTM.

Confectrician (approved):
Thanks. 👍🏼

PR #1905, opened 2022-10-28
Update introduction.md

nostroff:
Replace references to sitemap with page. Pages description row in concepts table could be improved - perhaps "are web site pages generated by openHAB (software) to provide a user interface that presents information and allows interactions.

### DIFF
--- a/introduction.md
+++ b/introduction.md
@@ -81,7 +81,7 @@ The following table gives a top-level description of the most important concepts
 | Channels | are the openHAB (software) connection between “Things” and “Items” (see below) | _see below_ |
 | Items | are the openHAB (software) generated representation of information about the devices | [click for more info on Items]({{base}}/concepts/items.html) |
 | Rules | that perform automatic actions (in its simplest form: if "this" happens, openHAB will do "that") | [click for more info on Rules]({{base}}/configuration/rules-dsl.html) |
-| Sitemap | is the openHAB (software) generated user interface (web site) that presents information and allows for interactions | [click for more info on Sitemaps]({{base}}/ui/sitemaps.html) |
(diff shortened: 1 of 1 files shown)


Confectrician:
I am not sure, if we should delete Sitemaps completely from the table.
Sitemaps may not be the main UI Interface anymore, but Basic UI and (at least) the Android App are still using sitemaps.

So they are a valid and available approach of building a ui.
Also Sitemaps are not marked as deprecated at any point currently.

nostroff:
The motivation for the change is to align the introduction section with the Main UI. As a newbie, having sitemaps included as a key concept and no mention of the Main UI/pages was a source of confusion.

rkoshak:
> I am not sure, if we should delete Sitemaps completely from the table.
> Sitemaps may not be the main UI Interface anymore, but Basic UI and (at least) the Android App are still using sitemaps.

I think it depends on whether we want sitemaps to be a part of our first impression. That table is not and can not be comprehensive. So are sitemaps part of what we want to be in our first impression? I'd say no, and instead we should point at MainUI Pages instead. I'm not sure the Getting Started Tutorial even mentions sitemaps except in passing. 

That doesn't mean that sitemaps are deprecated or no longer supported or anything like that. It just means that MainUI is what we want to use as a first impression. 

I think the changes made in this PR are reasonable.

nostroff:
Please do not consider the proposed text as being sacred. More than anything, the PR has the perspective of someone new to OpenHAB and tries to be specific/constructive in identifying one area where the documentation can be improved (IMHO).

Confectrician:
Valid points.
Thanks for the discussion. 🙂

Confectrician (approved):
Thanks. 👍🏼

PR #1937, opened 2022-12-02
Documentation for JSR223 cache preset

J-N-K:
Related to https://github.com/openhab/openhab-core/pull/2887

### DIFF
--- a/configuration/jsr223.md
+++ b/configuration/jsr223.md
@@ -370,7 +370,7 @@ scriptExtension.importPreset("RuleFactories")
 
 The `cache` preset does not provide a default import and needs to be imported explicitely.
 
-::::
+:::: tabs
 
 ::: tab Groovy
 ```groovy
--- a/configuration/jsr223.md
+++ b/configuration/jsr223.md
@@ -415,7 +415,8 @@ scriptExtension.importPreset("cache")
 It provides two different caches:
 
 - `sharedCache`: This cache is shared over all languages and all scripts (file-based and UI). Usage of entries is tracked and entries will be removed if the last script that ever accessed an entry is removed.
-- `privateCache`: This cache is private to a script engine, usually that means it is private to a script or rule, depending on the implementation of the scripting language. It is cleared when the script engine is unloaded (i.e. usually when the script is unloaded).
+- `privateCache`: This cache is private to a script engine, usually that means it is private to a script or rule, depending on the implementation of the scripting language.
+It is cleared when the script engine is unloaded (i.e. usually when the script is unloaded).
 
 In both cases values that are either a `ScheduledFuture<?>` or a `Timer` are cancelled by calling `.cancel()` on the object if the object is removed automatically.
 
--- a/configuration/jsr223.md
+++ b/configuration/jsr223.md
@@ -376,6 +376,8 @@ The `cache` preset does not provide a default import and needs to be imported ex
 
 ```groovy
 scriptExtension.importPreset("cache")
+
+sharedCache.put("x", "y")
 ```
 
 :::
@@ -384,6 +386,8 @@ scriptExtension.importPreset("cache")
 
 ```js
 scriptExtension.importPreset("cache")
+
+var valueX = sharedCache.get("x")
 ```
 
 :::
@@ -392,6 +396,8 @@ scriptExtension.importPreset("cache")
 
 ```js
 var { sharedCache, privateCache } = require('@runtime/cache')
+
+sharedCache.remove("x")
 ```
 
 :::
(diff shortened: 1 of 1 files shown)


rkoshak:
Does it make sense to include a Nashorn example? I think it'll be the same as Groovy/Jython.

```
scriptExtension.importPreset("cache")
```

I can see both ways since we really should encourage users to avoid Nashorn, but it's there now and we have the add-on on the marketplace to continue support when we move to Java 17. 

@jimtng, jRuby's import should be similar too, right? I personally would really like to see jRuby represented also.

J-N-K:
I can add it if someone confirms it's working. I don't have a JDK 11 installed, so can't test Nashorn.

rkoshak:
> I can add it if someone confirms it's working. I don't have a JDK 11 installed, so can't test Nashorn.

Actually, I wonder if it's a better test if it's tested with the Nashorn add-on on the marketplace.

I need to update my OH before I can test it and I'm just about out of time for the day. If I don't see anything I can try the test tomorrow.

configuration/jsr223.md
```diff
@@ -366,6 +366,66 @@ scriptExtension.importPreset("RuleFactories")
...
+```
+
+:::
+
+::::
+
+It provides two different caches:
+
+- `sharedCache`: This cache is shared over all languages and all scripts (file-based and UI). Usage of entries is tracked and entries will be removed if the last script that ever accessed an entry is removed.
+- `privateCache`: This cache is private to a script engine, usually that means it is private to a script or rule, depending on the implementation of the scripting language. It is cleared when the script engine is unloaded (i.e. usually when the script is unloaded).
```
Confectrician, line 418:
```suggestion
- `privateCache`: This cache is private to a script engine, usually that means it is private to a script or rule, depending on the implementation of the scripting language.
It is cleared when the script engine is unloaded (i.e. usually when the script is unloaded).
```

Confectrician (commented):
Just one small remark.

From  the discussion it would be best to wait until you have clarified the example?

J-N-K:
The Nashorn example is already there. We can wait for the ruby example.

configuration/jsr223.md
```diff
@@ -366,6 +366,67 @@ scriptExtension.importPreset("RuleFactories")
 | `ConditionHandlerFactory` | `org.openhab.core.automation.module.script.rulesupport.shared.factories.ScriptedConditionHandlerFactory` |
 | `TriggerHandlerFactory`   | `org.openhab.core.automation.module.script.rulesupport.shared.factories.ScriptedTriggerHandlerFactory`   |
 
+#### `cache` Preset
+
+The `cache` preset does not provide a default import and needs to be imported explicitely.
```
wborn, line 371:
```suggestion
The `cache` preset does not provide a default import and needs to be imported explicitly.
```

wborn:
> We can wait for the ruby example.

Maybe @ccutrer can provide it after finishing https://github.com/ccutrer/openhab-jrubyscripting/pull/83

ccutrer:
Yeah, don't wait in Ruby to get this merged. I'll make sure we get a Ruby example in in another PR.

florian-h05:
Do we also want to provide a JS Scripting (GraalJS) example?

If yes, in which way? As this is one the JSR223 page, it would make sense to document the raw access without using the helper libray. If we want to document it using the helper library, this addition is something for a follow up PR, as I currently don't have the new cache implemented in the library.

J-N-K:
@florian-h05 I don‘t think that is necessary since the helper libraries are always installed, correct?

florian-h05:
The helper libraries are always installed.

I'm not sure what you refer to with 

> I don‘t think that is necessary

J-N-K:
An example without the helper libraries. There is already an example for JS Scripting. I can add „GraalJS“ to the tab-title if you want.

florian-h05:
> There is already an example for JS Scripting.

I've just checked it, looks fine 👍

> I can add „GraalJS“ to the tab-title if you want.

Not needed.

configuration/jsr223.md
```diff
@@ -366,6 +366,67 @@ scriptExtension.importPreset("RuleFactories")
...
+:::
+
+::: tab Jython
+
+```python
+scriptExtension.importPreset("cache")
+```
+
+:::
+
```
ccutrer, line 412: ::: tab JRuby

```ruby
# the preset is imported automatically by the helper library.
value_x = shared_cache[:x]
shared_cache.delete(:x)
shared_cache[:x] = "y"
```

:::

Confectrician (approved):
Nothing left from my side and alle examplers are available.

Anything left to wait for from your side?
If not please give a short ping and i will merge.

J-N-K:
IMO you can merge.

PR #1947, opened 2022-12-05
Documentation for using the cache in DSL

J-N-K:
After https://github.com/openhab/openhab-core/pull/3187 has been merged the cache is also available for DSL rules.


Confectrician (approved):
Thanks. 👍🏼

Maybe we should convert this large list into a table at a later point.
It has grown pretty well.

PR #1313, opened 2020-12-13
Update docs for SmartHome namespace change

wborn:
Also removes pages for the "IoT Marktplace" and "REST Docs" add-ons which no longer exists in OH3.

---

These changes will also help a lot @Confectrician!

PR #1951, opened 2022-12-10
Fix topic subscription example

wborn:
This seems to have been overlooked in #1313


Confectrician (approved):
Thanks for the catch.
It is really difficult to catch all of those, since "smarthome" is such a generic word.

PR #1958, opened 2022-12-17
Add NUT Binding logo

florian-h05:
@Confectrician This adds the missing network UPS tools (NUT) Binding logo.


wborn (commented):
Did you manually scale this up? It looks pixelated.

florian-h05:
Yes, I’ve scaled this one up to meet the 512x512 px described in the docs.

wborn:
I would think that only applies to downscaling so the images are not unnecessarily big. It's already upscaled by the browser when necessary which will result in the same pixelated look without transferring unnecessary data.

florian-h05:
Okay, I‘ll then put the original version here.

Confectrician (approved):

PR #1962, opened 2022-12-17
document bundle install and update

stefan-hoehn:
Signed-off-by: Stefan Höhn <mail@stefanhoehn.com>

see discussion https://community.openhab.org/t/correct-approach-of-installing-an-addon-jar-over-an-installed-one-in-the-addon-folder/142097


Confectrician (approved):
Thanks. 👍🏼

PR #1967, opened 2022-12-19
Moved reverse-proxy configuration to it's own page

Hilbrand:
To give a better balance between the different options I moved the NGINX configuration to it's own page. So that the security page is focussed on the different options, and not looks like a configuration page for NGINX as it's the preferred way to do this. Also added the option to run your own openHAB cloud.


Confectrician:
Probably we should also use a dedicated Sidebar Entry for this also.
I will have a look on it after changing the websites versions.

Confectrician (approved):
Thanks for the progress. :)
I will also reintroduce this for the stable docs.

PR #1976, opened 2022-12-23
Fix add-ons stylization

wborn:
This makes the documentation a bit more consistent to use "add-ons" or "Add-ons" where possible.


Confectrician (approved):
Thanks. 👍🏼

PR #1982, opened 2022-12-28
Update items.md

spacemanspiff2007:
Missed the link to the transformation profiles. I think with an entry in the table its much easier to find.


Confectrician (approved):
Thanks. 👍🏼

PR #1992, opened 2023-01-16
Fix .editorconfig

J-N-K:
The editor config specifies CRLF as line-ending while we use LF. This is corrected now.


Confectrician (approved):
Thanks. 👍🏼

PR #1996, opened 2023-01-24
Add sudo command for consistency

felixschndr:
This commands needs `sudo`. The one above has `sudo` in it so this one should have `sudo` as well


Confectrician (approved):
Thanks. 👍🏼

PR #1999, opened 2023-02-01
Fixed typo

rkoshak:
Signed-off-by: Richard Koshak <rlkoshak@gmail.com>

### DIFF
--- a/ui/sitemaps.md
+++ b/ui/sitemaps.md
@@ -504,7 +504,7 @@ See this [Tutorial](https://community.openhab.org/t/13761/1) for more details.
 
 **Technical constraints and details:**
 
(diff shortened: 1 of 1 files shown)


JueBag:
IMHO this sentence isn't true anymore. The use of a samplerate of 20 sec or less (as used by the default configs and as suggested for cusom configs on the rrd4j docs page) is enough to ensure a working database. 
I'd suggest to delete the whole sentence.

rkoshak:
Has this been tested? I'm hesitant to delete without confirmation that using `everyChange` only as the strategy still works to generate charts. I'm only using the default strategy with rrd4j and I'm not sure the best way to go about testing this myself. Maybe create a .persist and apply only everyChange to a test Item perhaps?

JueBag:
God point! 
I'll setup a new test using only "everyChange" on a duplicate of an item that changes often.
Will report...

Confectrician (approved):
Thanks. 👍🏼

PR #2022, opened 2023-03-09
Update windows.md

keef65:
When running as a service an extra folder called log is created in userdata  (Note: OpenHab log folder is userdata/logs with the s on the end). This extra folder is created because the default Karaf log folder is not explicitly set in the wrapper config.  These changes update the wrapper config file to make sure the Karaf log folder is set to the same as the OpenHab Logs folder.

See Community post - https://community.openhab.org/t/oh3-oh4-running-as-service-on-windows-creates-spurious-log-folder-with-openhab-log-empty/145048


Confectrician (approved):
Thanks. 👍🏼

PR #2037, opened 2023-03-22
Update docs for OSGi R8 upgrade

wborn:
This is related to the dependency upgrades due to the Karaf 4.4.3 upgrade.

See: https://github.com/openhab/openhab-core/pull/3252.


Confectrician (approved):
Thanks. 👍🏼

PR #2038, opened 2023-03-22
Document the script-prefix for Nashorn JavaScript

rkoshak:
Added note what the script prefix is for Nashorn JS, noted that Nashorn is legacy and explicitly stated that the syntax is the same for both JS add-ons.


Confectrician (approved):
Thanks. 👍🏼

PR #2040, opened 2023-03-25
[developer] document thing-type update instructions

### LINKED ISSUE #2033: [developer] Thing type update instructions not (yet) documented
Functionality was added in core that makes it possible to update thing definitions by an `instructions.xml` file. While this can greatly improve the user experience, it is not yet documented. So before binding maintainers can actually use it it needs to be documented.

For details see:
https://github.com/openhab/openhab-core/pull/3330#issuecomment-1432033039
https://github.com/openhab/openhab-addons/pull/14206#issuecomment-1467540056

J-N-K:
Fixes #2033 
Core-PR: https://github.com/openhab/openhab-core/pull/3330

### DIFF
--- a/developers/bindings/thing-xml.md
+++ b/developers/bindings/thing-xml.md
@@ -767,79 +767,79 @@ The full Java API for bridge and _Thing_ descriptions can be found in the Java p
 </thing:thing-descriptions>
 ```
 
-| Property                     | Description                                  | |
-|------------------------------|----------------------------------------------|--------|
+| Property                     | Description                                        |           |
+|------------------------------|----------------------------------------------------|-----------|
 | thing-descriptions.bindingId | The identifier of the binding this types belong to | mandatory |
 
 **Bridges and Things:**
 
-| Property                       | Description                                  | |
-|--------------------------------|----------------------------------------------|--------|
-| bridge-type.id / thing-type.id | An identifier for the bridge/Thing type | mandatory |
-| bridge-type.listed / thing-type.listed | Denotes if user interfaces should list the bridge/Thing, e.g. for pairing | optional, defaults to true |
-| bridge-type.extensible / thing-type.extensible | If the bridge/Thing supports a generic number of channels the allowed channelTypeIds can be listed here. This provides a hint for UIs to support adding/removing channels. Channel groups are not supported. | optional |
-| supported-bridge-type-refs     | The identifiers of the bridges this bridge/Thing can connect to | optional |
-| bridge-type-ref.id             | The identifier of a bridge this bridge/Thing can connect to | mandatory |
-| label                          | A human-readable label for the bridge/Thing | mandatory |
-| description                    | A human-readable description for the bridge/Thing | optional |
-| category                       | Category this bridge/Thing belongs to, see categories) | optional |
-| channels                       | The channels the bridge/Thing provides | optional |
-| channel.id                     | An identifier of the channel the bridge/Thing provides | mandatory |
-| channel.typeId                 | An identifier of the channel type definition the bridge/Thing provides | mandatory |
-| label                          | A human-readable label for the channel | optional |
-| description                    | A human-readable description for the channel | optional |
-| channel-groups                 | The channel groups defining the channels the bridge/Thing provides | optional |
-| channel-group.id               | An identifier of the channel group the bridge/Thing provides | mandatory ||
-| channel-group.typeId           | An identifier of the channel group type definition the bridge/Thing provides | mandatory |
-| properties                     | Name/value pairs for properties to be set to the thing | optional |
-| representation-property        | The name of the property that contains a unique identifier of the thing | optional |
-| config-description             | The configuration description for the bridge/Thing within the ConfigDescriptionRegistry | optional |
-| config-description-ref         | The reference to a configuration description for the bridge/Thing within the ConfigDescriptionRegistry | optional |
-| config-description-ref.uri     | The URI of the configuration description for the bridge/Thing within the ConfigDescriptionRegistry | mandatory |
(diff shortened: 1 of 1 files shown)


Confectrician:
Does this solve #2033 completely when finished?

jlaur:
@J-N-K - perhaps examples for adding and removing channels would be useful too? We will often use this reference when reviewing pull requests, so I guess being a bit more verbose could result in fewer questions and mistakes. Currently the XSD has to be checked for finding `add-channel` and `remove-channel`.

developers/bindings/thing-xml.md
```diff
@@ -849,3 +849,84 @@ The full XML schema for Thing type descriptions is specified in the [https://ope
...
+                            xmlns:update="https://openhab.org/schemas/update-description/v1.0.0"
+                            xsi:schemaLocation="https://openhab.org/schemas/update-description/v1.0.0 https://openhab.org/schemas/update-description-1.0.0.xsd">
+
+  <thing-type uid="foo:pool">
+    <instruction-set targetVersion="1">
+      <remove-channel id="water_level">
+      </remove-channel>
+      <add-channel id="chlorine">
+        <type>foo:concentration</type>
+        <label>Chlorine Concentration</label>
```
jlaur, line 900: What is the purpose of the label for `add-channel`?

J-N-K: If you use the same channel-type (e.g. `voltage`) for different channels of the thing (e.g. a three-phase power meter) inheriting the label from the channel-type is not desired. By using `label` you can modify the label to be `Voltage L1`, `Voltage L2`, `Voltage L3` .

J-N-K: In general (i.e. in the XML that defines the thing), if you define a channel, label and description are taken from the channel-type. If you set label or description in the channel, this will override what is given by the channel-type. The same applies here: explicit label/description take precedence over what is given in the channel-type.

jlaur: So when adding a new channels, is the precedence:
1. `add-channel`/`label` -> `channel`/`label` -> `channel-type`/`label`, or:
2. `add-channel`/`label` -> `channel-type`/`label`

?

If the answer is 1) then I don't see the use-case for overriding it (and doing so without I18N support), because the example you mentioned should already be possible to define correctly in the channel labels, which is also used when creating a new thing or using file-based configuration.

If the answer is 2), maybe this should be fixed in core, so that the label would also be picked from the channel definition without any need to customize this in the update instructions? It seems this would even mean that the label will be in English (for managed things).

J-N-K: The label is stored in the JSON database, together with the other channel information. It is put there when the channel is added to the thing, so usually when the thing is created or when the user adds an `extensible` channel. This can't be "fixed" because it is not broken, it works as designed.

When the thing is initially created, then the label is taken from the channel-type, if the XML does not explicitly define another one.

When the channel is added by the update mechanism this is the same: the label is taken from the channel-type, if it is not explicitly set on the update instruction.

jlaur: Okay, thanks for the explanation. So it's option 2) then. Just to be completely sure, does it mean that for:

```xml
<channel-type id="voltage">
	<item-type>Number:Voltage</item-type>
	<label>Voltage</label>
</channel-type>

<channel id="voltage-1" typeId="voltage">
	<label>Voltage L1</label>
</channel>

<add-channel id="voltage-l1">
        <type>foo:voltage</type>
        <label>Voltage L1</label>
</add-channel>
```

With foo.properties:
```
thing-type.foo.bar.channel.voltage-l1.label = Voltage L1
channel-type.foo.voltage.label = Voltage
```

And foo_da.properties:
```
thing-type.foo.bar.channel.voltage-l1.label = Spænding L1
channel-type.foo.voltage.label = Spænding
```

For managed things, the label after the upgrade will be: **Voltage L1**
Whereas for file-based things, or newly created managed things, it will be: **Spænding L1**

And if you omit the label in the update instruction, the result will be:

For managed things, the label after the upgrade will be: **Spænding**
Whereas for file-based things, or newly created managed things, it will be: **Spænding L1**

?

J-N-K: i18n is another issue. I didn't check, but if available it has highest priority.

jlaur: Okay, but still it seems like the only way to get fully correct labels and descriptions, is to recreate the thing? Since this PR is about the documentation, I think the only thing that could be improved here would be explaining that:
- If an added channel overrides the channel type label and/or description, these must also be added in the update instructions, since otherwise the default label/description from the channel type will be used.
- Even when doing so, they may not be used, if the channel type label and/or description are translated. _(unconfirmed)_ This does not change the recommendations though, it's still better to have them. This is just explanation. Maybe too detailed.

jlaur: The should not block the PR which is already a huge improvement over having no documentation at all. 🙂 I will probably investigate this a bit more later.

J-N-K:
Is there anything I can do to get this merged?

jlaur (approved):
LGTM

jlaur:
> Is there anything I can do to get this merged?

I would also like to see this merged. I keep linking to the README in this PR during PR reviews, which is somewhat cumbersome compared to linking directly to the real thing. 🙂

Confectrician:
> The should not block the PR which is already a huge improvement over having no documentation at all. 🙂 I will probably investigate this a bit more later.

Based on this comment i thought there were still open points from your discussions.
Seems i misunderstood this.

From what i read now it is perfetctly fine to be merged from your side and from my side already for a longer time. 🙂

Confectrician (approved):

PR #2042, opened 2023-03-28
Document the new dynamic scripting-language transformation

jimtng:
Depends on https://github.com/openhab/openhab-core/pull/3487

### DIFF
--- a/addons/transformations.md
+++ b/addons/transformations.md
@@ -64,15 +64,15 @@ To keep these examples simple, the contents of the referenced files `window_esp.
 The script transformation is available from the framework and needs no additional installation.
 It allows transforming values using any of the available scripting languages in openHAB (JSR-223 or DSL).
 
-The script needs to be placed in the `$OPENHAB_CONF/transform` folder with the native extension for the chosen language, for example `stringlength.js` for a
-transformation using JS Scripting. The script file name here acts as the `script identifier` for the
-script transformation. A script identifier can also refer to a transformation script created in the UI.
+The script needs to be placed in the `$OPENHAB_CONF/transform` folder with the native extension for the chosen language, for example `stringlength.js` for a transformation using JS Scripting.
+The script file name here acts as the `script identifier` for the script transformation.
+A script identifier can also refer to a transformation script created in the UI.
 
 The input value is injected into the script context as a _string_ variable `input`.
 Additional parameters can be injected into the script by adding them to the script identifier in URL style (`scale.js?correctionFactor=1.1&divider=10` would also inject `correctionFactor` and `divider`).
 These additional parameters are also injected into the script context as _string_ variables.
-The result of the transformation is provided by the script as its return value. It can be `null`,
-a _string_, or a value of a type that properly implements `.toString()`.
+The result of the transformation is provided by the script as its return value.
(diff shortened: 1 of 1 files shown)


jimtng:
The feature has been merged, so this PR can also be merged, after a review.

addons/transformations.md
```diff
@@ -60,29 +59,29 @@ Transformation files need to be placed in the directory `$OPENHAB_CONF/transform
...
-The `SCRIPT` transformation is available from the framework and needs no additional installation.
+The script transformation is available from the framework and needs no additional installation.
 It allows transforming values using any of the available scripting languages in openHAB (JSR-223 or DSL).
 
-The script needs to be placed in the `$OPENHAB_CONF/transform` folder with an extension `.script` regardless of the actual script type.
-When referencing a transformation, the script type must be prepended to the filename (e.g. `dsl:stringlength.script` for the DSL version of `stringlength.script`).
-Please note that you cannot have transformations with the same name and different languages as the file-extension is always `script`.
-The script type depends on the scripting engine used, it is usually either the file extension or the MIME-type of the scripts.
+The script needs to be placed in the `$OPENHAB_CONF/transform` folder with the native extension for the chosen language, for example `stringlength.js` for a
+transformation using JS Scripting. The script file name here acts as the `script identifier` for the
```
stefan-hoehn, line 68: Pls, don't use line breaks in between. Pls, put each sentence in its own line like

... for a transformation using JS Scripting.
The script file name ...

addons/transformations.md
```diff
@@ -60,29 +59,29 @@ Transformation files need to be placed in the directory `$OPENHAB_CONF/transform
...
+script transformation. A script identifier can also refer to a transformation script created in the UI.
```
stefan-hoehn, line 69: Put "A script ..." into a separate line (see above).

addons/transformations.md
```diff
@@ -108,17 +109,104 @@ Note the overall syntax is the same.
...
+
 ::::
 
-Currently the `SCRIPT` transformation is not available as profile.
+### Inline Script Transformation
+
+A simple transformation rule can also be given as an inline script. The inline script should
+start with the `|` character. Quotes within the script may need to be escaped with a `\` when
+used within another quoted string such as in sitemaps or text configurations.
+
```
stefan-hoehn, line 137: ^ see above. Put each sentences into separate line.

stefan-hoehn (changes requested):
Thanks for the amount of the work, you put it to this documentation.

addons/transformations.md
jimtng: Should I put "The inline script should..." also in its own separate line?

stefan-hoehn: Yes, correct: each sentence should start at a new line (the rendering will concatenate them into a one-liner).
FYI, to be honest, I do not know if this is required in the case when you have multiple sentences in table columns nor do I even know if that would work. For that particular case I got in touch with @Confectrician to ask him. As a result you don't have to do that in table columns.

jimtng: OK I've split them up.

jimtng: You can't break them into multiple lines inside a table. All the columns would have to all be in one single line per row.

stefan-hoehn (approved):
LGTM, thanks!

PR #2044, opened 2023-04-02
Update serial.md as openHAB can now use symlinks

dalgwen:
With https://github.com/openhab/openhab-core/pull/3290, users can now use stable symlink instead of setting udev aliases.


stefan-hoehn (approved):
LGTM, thanks for the contribution 👍

PR #2072, opened 2023-05-02
Explain new UoM system

J-N-K:
Related to https://github.com/openhab/openhab-core/pull/3481

### DIFF
--- a/concepts/units-of-measurement.md
+++ b/concepts/units-of-measurement.md
@@ -22,34 +22,75 @@ The metric system with SI units is used for the rest of the world.
 This conversion will convert the given `QuantityType` into a default unit for the specific dimension of the type.
 This is:
 
-| Dimension     | default unit metric        | default unit imperial  |
-|---------------|----------------------------|------------------------|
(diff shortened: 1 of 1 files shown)


stefan-hoehn (approved):
Thanks for the comprehensive explanation, Jan. LGTM.

J-N-K:
I did nothing :-) Maybe GitHub selected that automatically.

lolodomo:
The documentation is clear now, thank you.
I see just one thing missing. We should explain to the user what unit he will get in rules for any item state. It is implicit but I think it will be better to say it.
The following sentence could be enhanced: "The state description is used for display purposes only, it can contain any compatible unit and will not affect what is persisted or used in events."
Proposal: The state description is used for display purposes only, it can contain any compatible unit and will not affect what is persisted or used in events **and rules**.
WDYT ?

stefan-hoehn:
Would it even make sense to change the following in the UI

image

to 

Pattern or transformation applied to the state ONLY for display purposes?

PR #2073, opened 2023-05-02
Explain upgrade tool

J-N-K:
Related to https://github.com/openhab/openhab-core/pull/3268

### DIFF
--- a/installation/index.md
+++ b/installation/index.md
@@ -130,3 +130,33 @@ or by working on your own configuration.
 
 The very active [openHAB Community Forum](https://community.openhab.org) can provide you with many more details, hints, and tips.
 If you run into any problems, use the search function in the forum or open a new thread with your detailed question.
+
+## Upgrading
+
+In some cases upgrading to a new version of openHAB requires additional steps.
+For textual based configuration this can usually be done using a text editor or VS Code.
+For UI configuration updates to the JSON database should not be done manually but by using the upgrade tool.
+
+The upgrade tool is a java application and allows performing different steps.
+Each step can only be executed once (unless you `--force` the tool to perform them again).
+Currently the following steps are supported as arguments to `--command`:
+
+- `itemCopyUnitToMetadata`: With openHAB 4 presentation and internal representation of the unit of an item are separated.
+This step copies the unit set in the state description of an item to the new `unit` metadata.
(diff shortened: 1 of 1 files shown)


installation/index.md
```diff
@@ -130,3 +130,33 @@ or by working on your own configuration.
...
+
+The upgrade tool is a java application and allows performing different steps.
+Each step can only be executed once (unless you `--force` the tool to perform them again).
+Currently the following steps are supported as arguments to `--command`:
+
+- `itemCopyUnitToMetadata`: With openHAB 4 presentation and internal representation of the unit of an item are separated.
+This step copies the unit set in the state description of an item to the new `unit` metadata.
+This is necessary to keep units consistent for data persisted with previous versions of openHAB.
+- `linkUpgradeJsProfile`: The JS Scripting transformation/profile has been removed because of changing to Java 17.
+They have been replaced by the generic script transformation/profile which require different configuration options.
```
stefan-hoehn, line 148: They or this?
which requires...

installation/index.md
```diff
@@ -130,3 +130,33 @@ or by working on your own configuration.
...
+This step rewrites the profiles to the new format.
+
+The upgrade tool needs to know the path to the openHAB userdata folder (e.g. `/var/lib/openhab` on most Debian like systems).
```
stefan-hoehn, line 151: would it make sense to mention 'openhabian' here? (not everyone knows that is Debian-based)

installation/index.md
J-N-K: Transformation and profile are two different things, therefore I decided to use "they/require" instead of "this/requires".

installation/index.md
J-N-K: If openhabian uses that directory, then probably yes. I don't know much about openhabian.

J-N-K:
@openhab/documentation-maintainers  Can you check why this is not showing up on next.openhab.org?

SteWers:
@J-N-K I'm a little confused, because you are referring to the generic script transformation. As I understand, the `script` transformation has gone with PR [#3487 ](https://github.com/openhab/openhab-core/pull/3487). Am I wrong?

J-N-K:
It's still the generic script transformation. the change was to use  the respective prefix (JS, RB, ...) for each language instead of `SCRIPT` for all.

PR #2075, opened 2023-05-07
Fix markdown in installation/index.md

J-N-K:
I think this is the reason why #2073 is not showing up on next.openhab.org. The question is, why this did not show up in the PR builds.

### DIFF
--- a/installation/index.md
+++ b/installation/index.md
@@ -153,10 +153,6 @@ If the tool is not operated from that folder, it can be specified by using `--di
 
 Example:
 
-```
+```shell
 java -jar upgradetool.jar --dir /var/lib/openhab --command itemCopyUnitToMetadata
 ```
-
-
-
-


stefan-hoehn:
wow, how did you find that out? I wasn't aware that this is required not to fail to be published. I am going to take this to @Confectrician, so we can investigate that. At least I know that I'll have a closee eye during the review on that.

I'll merge it and let's see if it solves the problem.

stefan-hoehn (approved):

J-N-K:
Maybe @ghys is right. I just check the build status on the repo-page and that showed these markdown errors as the reason why the build failed.

stefan-hoehn:
@Confectrician is looking into atm

Confectrician:
Just for clarification:

Markdown errors do not prevent the page builds at any time.
This is just internal for keeping the docs as clean as possible.

J-N-K:
But they show up as "failed build":

Bildschirmfoto 2023-05-07 um 19 16 37

Confectrician:
Sorry i should have clarified it a bit more in detail.

The build results in this github repository do not affect the website builds on our external netlify instance where a vuepress build is executed. 🙂

PR #2076, opened 2023-05-07
Fix markdown in UoM documentation

J-N-K:
See #2075


Confectrician:
As mentioned in the linked pr markdwon erros do not prevent any websit builds. 🙂 
Anyway thanks for taking care of a correct markdown syntax.

I will trigger a manual build for now, so we have at least the latest changes from this repository available in the next.openhab.org homepage and then check why the ci is having problems at a special point of the maven config.

Confectrician (approved):

PR #2078, opened 2023-05-08
Typos

dilyanpalauzov:


dilyanpalauzov:
I added an example how to use Group with Number, Dimession and a function.

PR #2081, opened 2023-05-17
configuration/items: write Tags are for Semantic model

dilyanpalauzov:
and remove text that tags are new or hardly used


Confectrician (approved):
Thanks. 👍🏼

PR #2083, opened 2023-05-23
[icons] Improved icons for Hue and PowerView

andrewfg:
Improved icons for Philips Hue and Hunder Douglas PowerView


Confectrician:
Both pictures have no transparent background and the hue one is smaller than the current picture.
I am not sure if this is a improvement at all.

@stefan-hoehn wdyt?

andrewfg:
The problem with the current images is the form factor, which needs to be about 4:3 for best looks on the web pages, without cropping artifacts. But if you also want transparency, and larger size, I will try again.

Confectrician:
We do not use the transparency effectifely currently (especially for the dark theme of the docs, but this may change in the future, so we are prepared best with having it. That's the reason why i "want" it.

I will have another look at both pages to get an impression of the cropping.

andrewfg:
@Confectrician I just uploaded a bigger Hue image

andrewfg:
Both images now have transparency; the Hue has a transparent 'swoosh' across its middle, which is apparently how Philips wants it to be..

stefan-hoehn:
@Confectrician to me the new logos look fine.

stefan-hoehn (approved):

Confectrician (approved):
Yes, looks good.
Thanks for the patience. 👍🏼

PR #2091, opened 2023-06-17
ui/sitemaps.md: emphasize period= is mandatory parameter in Chart

dilyanpalauzov:
I just struggled to configure a Chart in sitemap, until I figured out that `period=` must be set.

This changeset emphasizes that `period=` is mandatory.  It also sets `item=` in `Text` and `Chart` as mandatory, and `item=` in `Group` as optional, as this is what makes sense for me.

### DIFF
--- a/ui/sitemaps.md
+++ b/ui/sitemaps.md
@@ -496,8 +496,8 @@ Video url="https://demo.openhab.org/Hue.m4v"
 ### Element Type 'Chart'
 
 ```java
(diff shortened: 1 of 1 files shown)


dilyanpalauzov:
For `Group` and `Text` I reverted the initial changes, as the system does not accept `Group` without `item=`, but accepts `Text` without `item=`

Confectrician (approved):
Thanks. 👍🏼

PR #2106, opened 2023-07-27
Remove name "Paper UI" from transformations.md and replace it with "Main UI"

WolfgangSn:
Replaced Paper UI by Main UI


Confectrician (approved):
Thanks. 👍🏼

PR #2107, opened 2023-07-27
Blockly uom fix pic

### LINKED ISSUE #2104: Blockly Unit of Measurement
Hi. I think there is a mistake in two pictures in the Blockly Unit of Measurement documentation, namely: 
https://github.com/openhab/openhab-docs/blob/5b66fb0591daf0558b02c6e2eed657ff409d9c53/configuration/blockly/rules-blockly-uom.md?plain=1#L87
https://github.com/openhab/openhab-docs/blob/5b66fb0591daf0558b02c6e2eed657ff409d9c53/configuration/blockly/rules-blockly-uom.md?plain=1#L91

The pictures implies that regular comparison blocks will work (without the leading "Qty"). In my experience they do not. It's a bit tricky to figure out since there is no error message. My guess is that the strings are compared rather than the quantities, giving valid boolean out, but not the result you expect.

The "More examples" picture below shows it correctly:
https://github.com/openhab/openhab-docs/blob/5b66fb0591daf0558b02c6e2eed657ff409d9c53/configuration/blockly/rules-blockly-uom.md?plain=1#L95

stefan-hoehn:
fixes #2104


Confectrician:
@stefan-hoehn you added this to 4.0 milestone.
Is this on purpose? Do you want this to be als ported to the stable docs?

stefan-hoehn:
I was actually thinking if 4.0.1 would be related to 4.0 or 4.1 and I honestly didn't know. It belongs to 4.0.1, so does that mean 4.1 already?

Confectrician:
It is not really about which milestone we choose, i would need to know the answer of my question to give some proper information:

>Do you want this to be als ported to the stable docs?

So do you want it to be ported?

stefan-hoehn:
I should have been more clear. No, it shouldn't have been backported, just to be long the next minor release 4.0.2 but I am fine with 4.1 as well.

PR #2108, opened 2023-07-29
Fix broken links

wborn:
Some documentation files were moved in #1872 which caused these links to be broken.


Confectrician (approved):
Thanks. 👍🏼

PR #2126, opened 2023-08-25
[ide-vscode] Add instructions to tasks.json

Rosen01:
Add 3 tasks for Visual Studio Code IDE :
- mvn Spotless (Fix codestyle)
- mvn update properties (Default)
- mvn update properties (Default) [full plug-in name]

Force code style update before compile.

Signed-off-by: Daniel Rosengarten <github@praetorians.be>

### DIFF
--- a/developers/ide/examples/vscode/tasks.json
+++ b/developers/ide/examples/vscode/tasks.json
@@ -159,7 +159,7 @@
             "type": "shell",
             "command": "mvn",
             "args": [
-                "spotless:apply",
+                "spotless:apply"
             ],
             "problemMatcher": []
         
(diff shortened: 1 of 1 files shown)


stefan-hoehn (changes requested):
First of all, thanks for contribution and I am very sorry it took me so long. I never had the time to really look into it and didn't just want to wave it through.

There is a minor change I like you to do.

On the other hand, I don't have a vsc setup, so it would be a lot of work for me to test it. However, it does look good to me.

I checked who also worked on that file and it seems @wborn may be using it. 
Wouter would you mind having a quick look at that?

Rosen01:
I only develop with VS Code and I use the 2 first tasks to avoid check errors when I push updates ([Velbus module](https://github.com/openhab/openhab-addons/pull/15661)).

stefan-hoehn (approved):

stefan-hoehn:
Thanks, Daniel, I trust you in this that this is a proof it works.

PR #2127, opened 2023-08-25
Solax and Paradoxalarm binding logos added

theater:


theater:
@openhab/documentation-maintainers Could someone review the addition of these two logos please? If they're not OK just let me know.

stefan-hoehn (approved):
LGTM

Sorry for the time, @jlaur , I was on vacation. Now only seeing while taking off 😉

PR #2130, opened 2023-09-02
Update transformations.md with missing commata

ericbodden:
I am not entirely sure but I think you are missing these commata


stefan-hoehn (approved):
Thx, Eric!
After having a second look from @florian-h05 , it looks good to us.

PR #2132, opened 2023-09-08
Fix the Reference typeId of the system.electric-energy

theater:
Small fix in the reference type ID of the electric energy type. 
Took me a while to figure out the typo.


stefan-hoehn (approved):
Thanks for catching it. LGTM.

PR #2175, opened 2023-12-16
Sitemap: extended syntax for icon to define conditional rules

lolodomo:
Related to openhab/openhab-core#3820


ui/sitemaps.md
```diff
@@ -721,12 +721,49 @@ There are also the following keywords that can be used as colors:
...
-![battery-0](/iconsets/classic/battery-0.png "battery-0")
-![battery-30](/iconsets/classic/battery-30.png "battery-30")
-![battery-60](/iconsets/classic/battery-60.png "battery-60")
-![battery-100](/iconsets/classic/battery-100.png "battery-100")
+![battery-0](/iconsets/classic/battery-0.svg "battery-0")
+![battery-30](/iconsets/classic/battery-30.svg "battery-30")
+![battery-60](/iconsets/classic/battery-60.svg "battery-60")
+![battery-100](/iconsets/classic/battery-100.svg "battery-100")
+
+There is also a more powerful way to define a dynamic icon based on the states of different items and allowing you to attach any type of icon regardless of its source (not restricted to icons provided by openbHAB).
```
stefan-hoehn, line 732: openbHAB -> openHAB

ui/sitemaps.md
```diff
@@ -721,12 +721,49 @@ There are also the following keywords that can be used as colors:
...
+A condition including the AND operator will be considered as true if all individual comparisons are considered as true.
+Of course, it is possible to reference a different item in each comparison.
+
+Note that `item_name` and `operator` are both optional.
+If `item_name` is not provided, the Item name will default to the current Item.
+If an operator is not specified, the operator will default to `==`.
+
+Conditions are evaluated from left to right; the first matching condition determines the icon.
+If only a reference to an icon is specified, that condition is considered as true and that icon will be used.
+It can be used as a last condition to set a default icon that will be used in case any of the other previous conditions is true.
```
stefan-hoehn, line 754: The "It" isn't clear to me here (but maybe that is only me...)

stefan-hoehn (changes requested):
Thanks for the comprehensive documentation.
A small type in openHAB and a minor unclear sentence (at least to me)

ui/sitemaps.md
lolodomo: It is in reference to the previous sentence. But I will try to clarify.

lolodomo: It is changed.

stefan-hoehn (approved):
thx, @lolodomo

lolodomo:
As icons are now available in SVG format and no more in PNG format, I made the change in this PR for battery icons.
Unfortunately, the result is not the one expected, the icons are terribly big:

https://next.openhab.org/docs/ui/sitemaps.html#icons

@stefan-hoehn : any idea what I should add to fix a size ?

stefan-hoehn:
@lolodomo Good that you caught this (again something I should look more closely in the future!)

I researched that a bit and played around with it to see what we could do.

Unfortunately AFAIK there is no official way to tell markdown how to amend image sizes in the markdown image tag. There are some unofficial ways like adding the size this way Image which I tried it out but it doesn't work for us.

So the only way that I am aware of to fix this special case is to use the html image tag (which we usually try to avoid)

battery-0

Can you try that out?

lolodomo:
> So the only way that I am aware of to fix this special case is to use the html image tag (which we usually try to avoid)
> 
> battery-0
> 
> Can you try that out?

The icon will appear in the page with this solution without having to click on a link ?

stefan-hoehn:
I would say so, yes, can you try to make a PR and if it doesn't I am happy to look into it more closely.

PR #2181, opened 2023-12-18
[wemo] Increase logo resolution

jlaur:
This increases the resolution from 100x100 to 220x220. Still a bit small, but should look better than (neighbor logo for comparison):

image

Related to openhab/openhab-addons#16077
Also related: openhab/openhab-webui#2205


stefan-hoehn (approved):
Looks better than the old one, indeed. Thanks.

PR #2183, opened 2023-12-19
Update persistence.md

mherwege:
Remove statement about persistence configuration only being text based.


stefan-hoehn (approved):
True! ;-) Thanks.

PR #2189, opened 2023-12-27
Consolidate widget expression and css docs into their own pages

florian-h05:
This merged the widget expression docs from the „Building Pages“ and „Personal Widgets“ pages into a single new page and moves the CSS docs from „Building Pages“ to its own new page.
Both topics are very important for „Building Pages“ and „Personal Widgets“, so they should not be split across both or only present at one of those.

PR #2192, opened 2023-12-28
Update sidebar for #2189

florian-h05:


stefan-hoehn (approved):
LGTM, though we should make sure in the future not to add too many more menu entries, otherwise it could become confusing.

image

PR #2195, opened 2024-01-01
configuration/items: use more specific item types

dilyanpalauzov:
I think having all examples using as specific item types, as possible, utilizing the default UoM, is a good practice.

The wording “If no state presentation and no square brackets are given…” leaves unclear what happens, if state presentation is missing, but square brackets are present.  This is refined here.

Alternative good wordings will be just “If no state presentation is given”, or “If no square brackets are given”.

### DIFF
--- a/configuration/items.md
+++ b/configuration/items.md
@@ -64,9 +64,9 @@ itemtype itemname "labeltext [stateformat]" <iconname> (group1, group2, ...) ["t
 ```java
 Switch Kitchen_Light "Kitchen Light" {channel="mqtt:topic:..." }
 String Bedroom_Sonos_CurrentTitle "Title [%s]" (gBedRoom) {channel="sonos:..."}
-Number Bathroom_WashingMachine_Power "Power [%.0f W]" <energy> (gPower) {channel="homematic:..."}
+Number:Power Bathroom_WashingMachine_Power "Power [%.0f W]" <energy> (gPower) {channel="homematic:..."}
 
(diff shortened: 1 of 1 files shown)


dilyanpalauzov:
The information about what happens, when the square brackets are absent, is not universally correct:
* As explained at https://github.com/openhab/openhab-webui/issues/2251 missing `[%s]` implies not shown values in Basic UI, but
* [The example for `systembining`](https://www.openhab.org/addons/bindings/systeminfo/#example) suggest not to use state presentation/square brackets, and it still shows the current values (I tried only with `CPU_Load5`)

configuration/items.md
```diff
@@ -254,10 +254,10 @@ The state presentation is part of the Item label definition and contained inside
 The state presentation for the Item in the following example is "`%.1f °C`":
 
 ```java
-Number Livingroom_Temperature "Temperature [%.1f °C]"
+Number:Temperature Livingroom_Temperature "Temperature [%.1f °C]"
 ```
 
-If no state presentation and no square brackets are given, the Item will not provide a textual presentation of its internal state (i.e. in UIs no state is shown).
+If no state presentation is given, or there is no text between the square brackets, the the Item will not provide a textual presentation of its internal state (i.e. in UIs no state is shown).
```
stefan-hoehn, line 260: the the

stefan-hoehn (changes requested):

stefan-hoehn (approved):
LGTM now.

PR #2204, opened 2024-01-09
Update icon sources

mueller-ma:
All icon sources will be supported in the next releases of Android and iOS apps.


stefan-hoehn (approved):
LGTM

PR #2212, opened 2024-01-15
[tibber] Added image

soenkekueper:
Tibber binding is available a long time, but icon is missing.


stefan-hoehn (approved):
very good, small and transparent 

LGTM, thanks.

PR #2214, opened 2024-01-20
configuration/items: clarify that items linked to channels have implicit state presentation

dilyanpalauzov:
Per https://github.com/openhab/openhab-webui/issues/2251#issuecomment-1890903468 third bullet, if an item without explicit state presentation is linked to a channel, the State presentation is inherited from the channel.

This sentence is now complicated with its structure `If not A, or not B, and not C, then`. I am open to suggestions how to improve the structure of the sentence, to make it easier to understand.  But on the other side adding the current clarification is an improvement.

### DIFF
--- a/configuration/items.md
+++ b/configuration/items.md
@@ -257,7 +257,7 @@ The state presentation for the Item in the following example is "`%.1f °C`":
 Number:Temperature Livingroom_Temperature "Temperature [%.1f °C]"
 ```
 
-If no state presentation is given, or there is no text between the square brackets, the Item will not provide a textual presentation of its internal state (i.e. in UIs no state is shown).
+If no square brackets are given and the Item is not linked to a channel, the Item will not provide a textual presentation of its internal state (i.e. in UIs no state is shown).
 This is often meaningful when an Item is presented by a non-textual UI elements like a switch or a diagram.
 
 Formatting of the presentation is done applying [Java formatter class syntax](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/Formatter.html#syntax).


dilyanpalauzov:
Eventually I decided to use instead of

> If no state presentation is given, or there is no text between the square brackets, and the Item is not linked to a channel, the Item will not provide a textual presentation of its internal state (i.e. in UIs no state is shown).

just 

> If no square brackets are given and the Item is not linked to a channel, the Item will not provide a textual presentation of its internal state (i.e. in UIs no state is shown).

stefan-hoehn:
@florian-h05 WDYT?

florian-h05 (approved):

stefan-hoehn (approved):
LGTM

lolodomo:
This change is a problem as it undocuments (or hides) a feature which is that using `[]` as pattern allows removing value display.
@dilyanpalauzov @stefan-hoehn : this change has to be reverted or at least updated to restore documentation of this feature.

dilyanpalauzov:
I think that `[]` means the empty string.  Why does it need explictly to be documented?

stefan-hoehn:
@lolodomo Thanks you have noticed it, very much appreciated. This was even originally written by @ghys and I probably should have looked that up in the git annotations and ping him. It is always a bit hard for me to find the right people to double check the content when I am not completely into the topic...

So, as I don't know that topic too much, could you provide an idea how it should be changed instead? Should we revert it or is there a way to add something to what dilyan has provided...

dilyanpalauzov:
https://www.openhab.org/docs/configuration/items.html#state-presentation contains:

> The state presentation is part of the Item label definition and contained inside square brackets.

For me this means that empty string between the square brackets means empty string as State presentation.

lolodomo:
This s not so obvious for any normal user if not clearly documented.

dilyanpalauzov:
https://github.com/openhab/openhab-docs/pull/2221 adds

> No text between the square brackets also implies no textual presentation.

PR #2217, opened 2024-01-22
Use `@Component` annotation instead of XML in Event Bus example code

wborn:
Using these DS Component annotations is more convenient and is used everywhere else in the code nowadays.


stefan-hoehn (approved):
LGTM, thanks

PR #2220, opened 2024-01-29
Override karaf console settings in runtime.cfg

jimtng:
Resolves https://github.com/openhab/openhab-addons/issues/16256

I am not 100% sure whether userdata/etc/users.properties is preserved during an openhab upgrade. On Docker, it seems to be preserved but I'm not sure how it's handled in other installation methods.


stefan-hoehn (approved):
Nice and comprehensive addition to the topic. Thanks.

LGTM

PR #2225, opened 2024-02-03
Changed class for `actions` to the one actually used.

OlegAndreych:


configuration/jsr223.md
```diff
@@ -290,7 +290,7 @@ The `default` preset is preloaded, so it does not require importing.
 | `things`                | Instance of `org.openhab.core.thing.ThingRegistry`                                       |
 | `rules`                 | Instance of `org.openhab.core.automation.RuleRegistry`                                   |
 | `events`                | (internal) Used to send events, post commands, etc. [Details](#events-operations) below] |
-| `actions`               | Instance of `org.openhab.core.thing.binding.ThingActions`                                |
+| `actions`               | Instance of `org.openhab.core.automation.module.script.defaultscope.ScriptThingActions`  |
```
florian-h05, line 293: I don’t think that is correct.

For example do the `AstroActions` implement `org.openhab.core.thing.binding.ThingActions`: 
https://github.com/openhab/openhab-addons/blob/b316096e2992f55acca235257dc07d2ef9f49f42/bundles/org.openhab.binding.astro/src/main/java/org/openhab/binding/astro/internal/action/AstroActions.java#L47

OlegAndreych: As far as I can understand the doc, on the left-hand side of the table are context variables of rules scripts and imported constants, and on the right-hand side are their respective types.

If that's correct, then `actions` is a context variable of a rule script.  
If I understand correctly, a value for this variable is set there `org.openhab.core.automation.module.script.internal.defaultscope.DefaultScriptScopeProvider#DefaultScriptScopeProvider` as the field `thingActions` of type `org.openhab.core.automation.module.script.internal.defaultscope.ScriptThingActionsImpl`, which implements `org.openhab.core.automation.module.script.defaultscope.ScriptThingActions`, that I've specified in the PR.

Talking about `org.openhab.core.thing.binding.ThingActions` like `AstroActions`, they can be obtained by calling `org.openhab.core.automation.module.script.defaultscope.ScriptThingActions#get` on the context variable `actions`. But `org.openhab.core.thing.binding.ThingActions` is not the type of the context variable itself.

florian-h05: I have had a closer look and you are right.
Anway: Instead of just correcting the description, I would prefer that the `get` method returns a Thing action.

OlegAndreych: :thinking: 

I’m not quite understand. Could you please elaborate on the idea a bit more?

florian-h05, line 293:
```suggestion
| `actions`               | Use `actions.get(String bindingName, String thingUid)` to get the actions of a Thing (instanceof `org.openhab.core.thing.binding.ThingActions`)  |
```

florian-h05: Please fix the table's formatting afterwards, you may use http://markdowntable.com for that.

OlegAndreych: I’m convinced that knowing the type of the `actions` variable is useful for a typesafe coding.  

E.g. I’m using Intellij Idea with a custom `.gdsl` file to define types of context variables and imported constants to get useful code suggestions and checks.

I’ve noticed that all other table entries are containing types. The entry you suggested contains a way of use.  

As for me, it seems inconsistent. Maybe https://www.openhab.org/docs/configuration/actions.html would be a better place to explain how to use actions in rule scripts to the user.  

The same table from this doc has an `itemRegistry` entry without usage explanation and it seems fine. I don’t see how `actions` variable is different in that regard.

If you insist on having the usage description for the `actions` variable, I propose to combine both type definition and usage explanation.

OlegAndreych:
I've added the link you provided and formatted the table.

Please look if is that what you meant?

configuration/jsr223.md
OlegAndreych: I've added the link you provided and formatted the table.

Please look if is that what you meant?

florian-h05 (approved):
LGTM, thanks!

stefan-hoehn (approved):
If it looks good to Florian, it LGTM 2 U ;-)
Thanks for the contribution Oleg.

PR #2230, opened 2024-02-09
Minor enhancements

florian-h05:
- Fixes some findings from Vuepress compilation warnings while working on the webite build.
- Improve the prepare docs ruby script similar to https://github.com/openhab/website/pull/445.


stefan-hoehn (approved):
Thanks in particularly finding out that blanks in :::tab need to be done with nbsp...

PR #2238, opened 2024-02-20
Add sqlite icon

jimtng:


stefan-hoehn:
As per definition the icons should be less than 30kB (I usually would also accepts around 50) and use transparency. Thx.

jimtng:
resized

stefan-hoehn (approved):

PR #2245, opened 2024-02-23
Add support for providing svg add-on logos

florian-h05:
This extends the Groovy script used by Jenkins CI to process add-ons with support for svg add-on logos.

I have verified the websites compatibility with SVG logs with the local build.

### DIFF
--- a/process_addons.groovy
+++ b/process_addons.groovy
@@ -60,8 +60,9 @@ def process_addon_type = { features, sources, type, collection, suffix, lblremov
                         }
                         label = label.trim()
                     }
-                    def logo = new File(project.basedir, 'images/addons/' + id + '.png').exists()
-                    if (! logo) log.info("No logo found.")
+                    def logo_svg = new File(project.basedir, 'images/addons/' + id + '.svg').exists()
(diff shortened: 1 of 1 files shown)


jimtng:
Cool! I was just thinking how it would be nice to be able to have svg icons. How would it work with webui addons page?

mherwege:
> Main UI is using a simple `img` element, which has SVG support, so this should not even require a change.

I do think there is an issue. There is an imageUrl for the addon image in various places in the code (the same method is duplicated), that assume a .png file.
See here:
https://github.com/openhab/openhab-webui/blob/57ed2da295007c939d0b219f9fcf034e658d39a9/bundles/org.openhab.ui/web/src/pages/addons/addon-details.vue#L219
https://github.com/openhab/openhab-webui/blob/57ed2da295007c939d0b219f9fcf034e658d39a9/bundles/org.openhab.ui/web/src/components/addons/addon-card.vue#L144
https://github.com/openhab/openhab-webui/blob/57ed2da295007c939d0b219f9fcf034e658d39a9/bundles/org.openhab.ui/web/src/components/addons/addon-list-item.vue#L76

florian-h05:
Hmm true.
I will take care.

PR #2247, opened 2024-02-24
Make basicui icon a bit bigger and more visible

jimtng:
I don't know how to "preview" the icon in the docs / mainui, so I hope this would look better. Is there a way to do that easily?

The details in the current one are too tiny
image

Using an image editing software and copy pasting / scaling, I came up with this "preview"


stefan-hoehn:
> Is there a way to do that easily?

Currently, unfortunately no, not that I a am aware of
@pgfeller and I are currently working on allowing to run the local build for the docs and the website much easier. So stay tuned.

stefan-hoehn (approved):
Thx in general that you work on providing all of these new images.  🥳

PR #2248, opened 2024-02-24
Add icon for pidcontroller

jimtng:
As always, if someone could come up with a better icon, feel free to replace this one!


stefan-hoehn (approved):
I like that one

PR #2253, opened 2024-02-25
Convert pwm, pid, map icons to svg and add jinja logo

jimtng:


stefan-hoehn (approved):
Thanks, looking good to me

PR #2254, opened 2024-02-25
Set params when using `@text/`

magx2:
Hey, I've spend some time to figure out how to pass params in `@text\`, because it was not put in the doc.

The example with `error` is kind of silly, but I did not come anything better.


stefan-hoehn (approved):
Thanks for providing this PR. LGTM.

PR #2256, opened 2024-02-26
Convert some add-on logos to SVG

florian-h05:
And add DynamoDB logo (unfortunately I could not find a SVG) and JPA logo (created by myself from the Java logo).


stefan-hoehn (approved):
very nice, thanks for adding them

stefan-hoehn:
As discussed offline we just need to make sure that not having pngs doesn't have have an impact on 3.x, so as you proposed we wouldhave 3.x patched for that. We will not patch 4.0: users of 4.0 are supposed to go to 4.1 then.

PR #2257, opened 2024-02-26
Profiles: include in sidebar

dilyanpalauzov:


jimtng:
+1 otherwise it seems that nothing links to this page (concepts/profile.html)

stefan-hoehn:
One of the reasons I am actually hesitant to link to that page that it is outdated and also contains duplicate content. So the better way would be to actually update the page or put the missing content to where it belongs.

PR #2266, opened 2024-03-01
Replace some PNG add-on logos with SVG version

florian-h05:


stefan-hoehn (approved):
Thanks, Florian, for providing them!

PR #2272, opened 2024-03-08
Document addon installation in Karaf from Marketplace

dilyanpalauzov:
as implemented in https://github.com/openhab/openhab-core/pull/4057 .

I do not understand how exactly the `openhab:addon list` command is supposed to work.

With OH 4.2.0M1 I can do in karaf `openhab:addons list karaf` or `openhab:addons list marketplace`.

The first lists as:

```sh
…
  binding-mecmeter                              4.2.0.M1             mecMeter Binding
  binding-meteoblue                             4.2.0.M1             meteoblue Binding
  binding-mycroft                               4.2.0.M1             mycroft Binding
  binding-mystrom                               4.2.0.M1             mystrom Binding
  misc-openhabcloud                             4.2.0.M1             openHAB Cloud Connector
  binding-zigbee                                4.2.0.M1             openHAB ZigBee Binding
  binding-pixometer                             4.2.0.M1             pixometer Binding
  binding-solax                                 4.2.0.M1             solax Binding
…
```
the second:
```sh
…
  marketplace:123573                            not set              Fire Alarm
  marketplace:128255                            not set              Scene Control Suite - Scene Modification
  marketplace:129987                            not set              Twitter
  marketplace:129618                            not set              BT devices in range widget
  marketplace:126440                            not set              ConnectedCar Details View Widget
  marketplace:121655                            not set              Caddx Binding Widget
  marketplace:106820                            not set              Keypad
…
```
What is the difference between these, as in
```
  automation-jsscriptingnashorn                 4.2.0.M1             JavaScript Scripting (Nashorn)
  marketplace:134770                            not set              JavaScript Scripting (Nashorn)
```
?  Aren’t all of them originating from the market place and why are then the names (ids) completely different?

### DIFF
--- a/configuration/addons.md
+++ b/configuration/addons.md
@@ -94,4 +94,17 @@ Place the .jar file in the folder Additional add-on files as described in File L
 ## Through the openHAB console
 
 Add-ons (also known as bundles) can be also installed and updated via the openHAB console.
-A detailed description can be found at [Installing or Updating Bundles](/docs/administration/bundles.html).
+
+To see the available online and offline add-ons:
+
+```sh
+openhab:addons list
+```
+
+To install an add-on in the background:
+
+```sh
+openhab:addons install <id-as-from-list-command>
+```
+
+Another approach can be found at [Installing or Updating Bundles](/docs/administration/bundles.html).


rkoshak:
>  Aren’t all of them originating from the market place and why are then the names (ids) completely different?

Not necessarily. Before OH 4, there was no Nashorn add-on. The only place to install it was from the Marketplace. After OH 4 was released, the Nashorn add-on became an official add-on. 

But official add-ons and Marketplace add-ons are different add-ons. So if you had a Marketplace add-on installed and then later installed the official add-on, it doesn't replace the Marketplace add-on. 

Based on the output you show above, I would conclude you have *both* the Marketplace and the official Nashorn add-on installed. You should remove the Marketplace add-on. You should see two entries for JS (Nashorn) under Add-on store -> Automation (note that in OH 4.2 M1 the organization of the add-on store has changed and you'll see them in the Automation section on the same page, not more separate tab). Click on each of them and it should be apparent which one comes from the Marketplace and you can remove it from there.

Or you can remove the marketplace one through the console. You should not have both installed and I don't think the Marketplace version is being maintained any longer.

Those add-ons with an ID "binding-name" are official add-ons. Those add-ons with an ID "marketplace:xxxxxxx" come from the marketplace.

dilyanpalauzov:
I tried to reproduce the above output, but now the only service I get is:

```
openhab> openhab:addons services
jar                  JAR-File add-on service
openhab> openhab:addons list
openhab> exit
```
while the web interface :8080/addons/bindings does show things like Adorne binding, air-Q Binding, AllPlay Binding.  I do not have Nashorn installed, it is returned twice by `openhab:addons list` (available online).

> But official add-ons and Marketplace add-ons are different add-ons.

Is this described somewhere?  Does the web interface (:8080/addons/) offer only official-addons, or also Marketplace add-ons?

rkoshak:
> Is this described somewhere?

I don't know but they are posted in a different place, they are listed separately in the Add-ons store, have a different ID, and they are provided by a separate add-on service. Given all of that it seems obvious to me they are different. You can find the "karaf" Nashorn add-on at https://github.com/openhab/openhab-addons/tree/main/bundles/org.openhab.automation.jsscriptingnashorn and the marketplace one at https://community.openhab.org/t/javascript-scripting-nashorn-3-3-0-0-4-0-0-0/134770. 

Note that that posting on the forum is *the* source for this add-on. The forum *is* the marketplace. OH pulls these add-ons from that category in the forum. 

>Does the web interface (:8080/addons/) offer only official-addons, or also Marketplace add-ons?

By default it only provides the list of official add-ons (i.e. karaf add-ons). However, if you set the service ID to "marketplace" it will give you the marketplace add-ons. Use /addons/services to see the list of all the addons services enabled on your system (I have jar, karaf (official add-ons), marketplace (official OH marketplace), and json (smarthomej marketplace). For example: http://openhab:8080/rest/addons?serviceId=marketplace. I'm sure there's an openhab:addons equivalent to that on the console as well.

I can't get into my karaf console at the moment (I've not tried since updating to OH 4.2 M1, may be a regression?) but I can messed around with the addons REST API endpoints. I'm assuming the endpoints match.

My mistake above. I misunderstood what `openhab:addons` does. Without further arguments all it does is list all the available add-ons. So my discussion about having both Nashorn add-ons installed is incorrect. 

Back to the original question, looking at https://next.openhab.org/docs/administration/runtime.html it appears that the new `openhab:addons` command is not yet documented. You should be able to get inline help though using `openhab:addons help`.

stefan-hoehn:
@rkoshak Would you then agree to the PR that @dilyanpalauzov has provided?

dilyanpalauzov:
This change is about briefly mentioning the opportunity in Karaf to use `addons list` and `addons install` commands.  All this market place and karaf/official addons is out of scope.  So I think as brief as the change here is, it does hint the user for this possibility.

rkoshak:
I don't have any problem with what's in this PR. But if documenting what the difference is between the four add-on services is out of scope for here I can't think of anywhere else it would go.

stefan-hoehn (approved):

PR #2284, opened 2024-04-14
[sitemap] Extended period attribute for chart to cover past and future

lolodomo:
Related to openhab/openhab-webui#2518


stefan-hoehn (approved):
Nice addition, thanks
(and I learned a lot about ISO8601 periods as well which was new to me 👏🏻)

PR #2285, opened 2024-04-17
[openwebnet] new logo using MyHOME

mvalla:
This PR changes the logo for the `openwebnet` binding to a new more recognizable one.

Reason is that, although the communication protocol used is OpenWebNet, the products/systems being integrated by this binding are all referred to as "MyHOME" or "MyHOME Radio" (for the radio Zigbee case).

Many users will find more natural to see the MyHOME logo when looking for this integration.

This also could be a first step to a possible further change where the name of the binding itself could be changed from `openwebnet` to `bticino-myhome` or `bticinomyhome`, but this is of course a much more impacting change so it needs further discussion with maintainers.


stefan-hoehn (approved):
LGTM, thanks.

PR #2287, opened 2024-04-21
[sitemap] New Slider parameter releaseOnly

mueller-ma:


lolodomo:
We could also explain in introduction the two different modes and what is the default behaviour (with the particular case of element generated for group member).

mueller-ma:
What exactly is the behavior when `releaseOnly` isn't set?

stefan-hoehn (approved):
LGTM, thx

lolodomo:
I will probably create an additional PR later. But the minimum is already here now.

PR #2138, opened 2023-09-30
Add logo for Danfoss Air Unit binding

jlaur:
This logo is for [Danfoss Air Unit binding](https://www.openhab.org/addons/bindings/danfossairunit/).

Source: https://www.danfoss.com/static/images/logo.svg
Converted to transparent PNG, width 512.

PR #2288, opened 2024-04-24
Replace logo for Danfoss Air Unit binding with SVG version

jlaur:
Now finally able to use the original logo in SVG format - see #2138.


stefan-hoehn (approved):
Thanks for providing the svg

PR #2289, opened 2024-04-30
Update persistence actions docs

mherwege:
Persistence actions have been extended to work with future persisted states in https://github.com/openhab/openhab-core/pull/3736.

This PR extends the documentation to provide a full list of all actions.


J-N-K:
@openhab/documentation-maintainers can you ping me when this is ready for merge? I'll merge the core PR then.

florian-h05 (approved):
LGTM, thanks!
Can you please add some docs about HistoricItem? It is return value but not documented …

mherwege:
> It is return value but not documented …

There was some documentation, but in the Examples section. I moved it up and extended it a bit.

configuration/persistence.md
```diff
@@ -232,44 +232,77 @@ Example:
...
+| `<item>.removeAllStatesBetween(ZonedDateTime, ZonedDateTime)`   | Removes all persisted States for an Item between certain points in time                                                                                                         |
 
 These extensions use the [default persistence service](#default-persistence-service).
 You may specify a different persistence service by appending a String as an optional additional parameter at the end of the extension.
 
+Some persistence services may not support persisting to other points in time than the current time and removing persisted States, therefore not all actions may be supported for all persistence services.
+
+Some extensions return a HistoricItem object.
+It represents the state of a persisted item at a certain point in time.
+The most useful methods of the HistoricItem object returned by some queries, are `.state` and `.getTimestamp`
```
florian-h05, line 304:
```suggestion
The most useful methods of the HistoricItem object returned by some queries, are `.getState()` and `.getTimestamp()`.
```

florian-h05 (changes requested):
Thanks, one minor comment left:

configuration/persistence.md
mherwege: That was the original text, and would work in DSL I believe. I changed it anyway.

florian-h05: .state and .timestamp should theoretically also work in JS, because in Nashorn compatibility mode of GraalJS access to the properties is possible that way. I guess it’s the same for DSL — but I personally think that using the getters is better style 👍

florian-h05 (approved):
LGTM now, thank you very much especially for the core PR!!

@stefan-hoehn Can you please merge?

stefan-hoehn (approved):

configuration/persistence.md
```diff
@@ -232,43 +232,71 @@ Example:
 
 The statement
 
-`Temperature.historicState(now.minusDays(1))`
+`Temperature.persistedState(now.minusDays(1))`
```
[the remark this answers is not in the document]
J-N-K: Currently both arte supported, but `historicState` is deprecated. The reason is that it can also return future states (if the persistence service supports it).

PR #2299, opened 2024-05-30
Clean-Up leftovers from Jekyll-based deployment to docs.openhab.org

florian-h05:
This removes the v2.1 and v2.2 docs folders.
If these docs would be used by any active deployment, it would be the https://v2.openhab.org/ deployment, which is deployed from the final-2.5.x branch.
Therefore we can savely remove these folders here.

Remove all Jekyll-related stuff, e.g. configs, resources.
Jekyll is not used anymore, as the docs are now integrated into the website build.
I have double checked that the website build does not pick up the removed stuff and even did a local production build of the website - everything looks good.


florian-h05:
@stefan-hoehn Can you please have a look?

stefan-hoehn (approved):
I trust you

PR #2302, opened 2024-06-02
solarforecast logo

weymann:
Logo for https://github.com/openhab/openhab-addons/pull/13308


stefan-hoehn (approved):
Thanks for the logo. I am merging this but would you be able to provide an svg for that which would be even nicer?

cc @florian-h05

PR #2309, opened 2024-06-13
[iotawatt] add IoTaWatt logo

PRosenb:


stefan-hoehn:
@PRosenb would you mind checking if there is an svg available because it is preferred over pngs?

PRosenb:
@stefan-hoehn yes, I was in contact with the creator and there is no SVG available unfortunately.

stefan-hoehn (approved):

PR #2323, opened 2024-06-27
Fix wrong negation

felixschndr:
I am pretty sure this `not` is/was wrong. If not _what is an item_ instead?


jimtng:
Thanks for the contribution, but I believe the original text is not incorrect. It explains itself in the next sentence clause. I agree though that the whole sentence is a bit nebulous but I guess it's kind of an introduction to what an Item is.

stefan-hoehn:
I kind of understand where Felix is coming from. 

> An Item does not simply store information that is set by software (e.g., `OFF`, 3.141 or "No Error"); the information stored by an Item may also be set by actions that take place in your home.

How about phrasing it like that?

> An Item does not simply store information that is set by software (e.g., `OFF`, 3.141 or "No Error") but rather information that is set by actions that take place in your home.

felixschndr:
How about this? (Just remove the _confusing_ part)
> An Item does store information that is set by actions that take place in your home (e.g., OFF, 3.141 or `No Error`).

jimtng:
My suggestion would be to simply delete lines 15-18 completely. They don't give you information that isn't already explained further down. All they seem to do is add "noise" and confusion in your mind as you read through, especially if you are coming from a "blank canvas" not knowing anything about Items.

The actual knowledge about items will be built up by reading through the subsequent paragraphs which do provide actual and practical information about what Items are and what they do.

felixschndr:
> My suggestion would be to simply delete lines 15-18 completely. They don't give you information that isn't already explained further down. All they seem to do is add "noise" and confusion in your mind as you read through, especially if you are coming from a "blank canvas" not knowing anything about Items.
> 
> The actual knowledge about items will be built up by reading through the subsequent paragraphs which do provide actual and practical information about what Items are and what they do.

I agree 👍🏻

stefan-hoehn:
Me too, do you change the PR accordingly?

felixschndr:
Done

stefan-hoehn (approved):

PR #2324, opened 2024-07-01
Add logo for siemens hvac binding

lo92fr:


stefan-hoehn (approved):
thx, I'll approve but it would be great if you could provide an svg instead.

PR #2330, opened 2024-07-08
Make recently added logos visible on the website

florian-h05:
This change is required as Jenkins is not migrating external content anymore, and therefore the new logos are not picked up.


florian-h05:
@kaikreuzer Can you please merge?

PR #2333, opened 2024-07-08
Add logo for Pentair binding

jsjames:


stefan-hoehn:
I would prefer the pentair _svg_ logo. There is actually one available.

stefan-hoehn (approved):
Great, thanks

PR #2341, opened 2024-07-29
[ism8] Fix broken image in docs

holgerfriedrich:
try to backport openhab/openhab-addons#17175 to stable docs, not sure if this is the right way to do it

### DIFF
--- a/_addons_bindings/ism8/readme.md
+++ b/_addons_bindings/ism8/readme.md
@@ -210,4 +210,4 @@ NULL=Undefiniert
 ```
 
 _Result_
-<img src="doc/Sitemap-Example.png" width="800" height="600">
+<img src="./doc/Sitemap-Example.png" width="800" height="600">


holgerfriedrich:
Docs (stable) are still broken, image "Result" is empty.

@stefan-hoehn can you assist how to back-port the fix from 4.3? Is this the way to do it?

stefan-hoehn:
@holgerfriedrich 
@florian-h05 
@pgfeller 

AFAIK, the addons content is not managed via the docs repo.  Hence the path "_addons_bindings/ism8/readme.md" doesn't even exist in the repo. So at least this PR IMHO doesn't make sense.
Only when the website is built, the whole content comes together.

Patrik, Florian, do you know? How do we backport this?

florian-h05 (approved):
LGTM, thanks!

This is the correct way of backporting the doc fix to the stable docs.

Some explanation how the ism8 readme ends up inside the docs repo:
Our Jenkins CI is collecting documentation from several repos and commits to the final branch of the openhab-docs repo. 
In normal cases, you fix a doc issue in the source repo and Jenkinks will pick up the fix there and bring it to openhab-docs final branch.
The final-stable branch is based of the current final branch on release and won't be updated anymore by Jenkins, so any doc fix/change needs to be done manually (like in this PR).

stefan-hoehn:
First of all, my bad. The path "_addons_bindings/ism8/readme.md" does exist but only in final-stable (and not in main) as these are created via the jenkins build when processing final-stable.

So, all good as mentioned by Florian.

PR #2354, opened 2024-08-29
jsr223.md: do not state that SimpleRule is used only in Jython

dilyanpalauzov:
Modify example for Groovy closure.  I believe its form is not `() -> {…}`, but `{ [params ->] … }` where params is optional, and -> is also optional when there are no params, cf. https://groovy-lang.org/closures.html

For RuleSimple preset insert hyperlinks to the documentation of the included classes.

Remove statement that SimpleRule is a base class for Rules only in Jython.


dilyanpalauzov:
Why is the CI failing in configuration/transformations.md, when I have not modified this file?

rkoshak:
There appear to be some markdown linting errors that made it into the merged version. I'm not sure how that happened nor the best way to handle that. You can fix it obviously as part of this PR but it would be good to know why this is showing up.

Maybe the linting rules changed?

dilyanpalauzov:
I do not understand it.  https://github.com/openhab/openhab-docs/pull/2358/ does pass the linter checks and there configuration/transforms.md is the same.  The problem with transforms.md is that there is a numbered list, in which since 20 August not all items start with `1. `, but with `2. `, `3. ` etc.

florian-h05 (approved):
LGTM, thanks!

Pinging @stefan-hoehn as a doc maintainer.

stefan-hoehn (approved):

PR #2367, opened 2024-09-16
[profiles] Update index.md

udo1toni:
For consistency, existing system profiles need keyword `system:`


jimtng:
I like not having to type so much and if a plain "follow" works without "system:follow" why not? It was implemented in core to allow it and "forcing" / documenting otherwise would negate that benefit.

udo1toni:
The point is, it's different in the items documentation: https://www.openhab.org/docs/configuration/items.html#profiles

jimtng:
Just to update my findings here as well (I did write about it on the forums too): core will tolerate it without "system:" but it will log an info message. Maybe we should ask @openhab/core-maintainers about which one is preferred here?

stefan-hoehn:
@jimtng So the docu change here is fine for you?

jimtng:
Yes

stefan-hoehn (approved):

PR #2380, opened 2024-10-01
Add back add-on logo to add-on READMEs

florian-h05:
This seems to got broken during all the work that happened on SVG add-on logos and the website in the last months.


stefan-hoehn (approved):
Thanks for finding an fixing this.

PR #2381, opened 2024-10-01
Adjust add-on README logo

florian-h05:
Follow-up for #2380.

Currently, the `<AddonLogo />` component is inserted as the first child into the readme, which causes styling issues.
It instead needs to be put after the heading.


stefan-hoehn (approved):
LGTM

PR #2382, opened 2024-10-06
Fix format for "datetime" context (config description parameter)

lolodomo:
The new expected format is "YYYY-MM-DD hh:mm:ss" instead of "YYYY-MM-DD hh:mm" as previously documented.

The "datetime" context is currently not used by official bindings and this context value is also not yet supported by Main UI.


lolodomo:
@florian-h05 for information

florian-h05 (approved):
Thanks!

stefan-hoehn (approved):

PR #2383, opened 2024-10-06
Update websocket.md

doits:
Clarify use of `accessToken` URL param in websocket url.

I was stuck connecting to the websocket and where to put the authentication token until I found some old forum thread by chance. Hopefully this helps other people.


stefan-hoehn (approved):
Thx, very much appreciated that you provided the more specific documentation. 👍

PR #2388, opened 2024-10-22
Thing actions: Clarify use of ActionOutputs annotation

florian-h05:
The current docs are wrong,
see https://github.com/openhab/openhab-core/blob/a22349abf4b3106ec8a3eb9d36799e334cdfbf25/bundles/org.openhab.core.automation/src/main/java/org/openhab/core/automation/module/provider/AnnotationActionModuleTypeHelper.java#L146-L159 for core implementation.

### DIFF
--- a/developers/bindings/index.md
+++ b/developers/bindings/index.md
@@ -631,7 +631,11 @@ This is to support the old DSL rules engine and make the action available there.
 
 Each parameter of an action member method must be annotated with `@ActionInput`.
 
-If you return values, you do so by returning a `Map<String,Object>` and annotate the method itself with as many `@ActionOutput`s as you will return map entries.
+If you return values, you do so by returning a `Map<String,Object>` and annotate the method itself with `@ActionOutputs` containing as many `@ActionOutput`s as you will return map entries, e.g.
+
+```java
+@ActionOutputs({ @ActionOutput(name = "foo", label = "Foo", type = "java.lang.String"), @ActionOutput(name = "bar", label = "Bar", type = "java.lang.String") })
+```
 
 ## Firmware information / Firmware update
 


florian-h05:
If the current docs are correct except from the above mistake, one only needs to annotate `@ActionOutput` inside the `@ActionOutputs` annotation if the action returns a `Map<String, Object>`. 
If there is only a single return value, no annotation is required and it won't be processed by core, as core is looking for `@ActionInputs`.
If a method returns such a Map, the `@ActionOutputs` annotation is needed.

For add-ons, this means:

1. we can remove `@ActionOutput` from all action returning only a single value
2. we need to add `@ActionOutputs` around the `@ActionOutput`s of all actions returning `Map<String, Object>`

At the moment, 2. affects not many actions:

```
cat **.java | grep @ActionOutput | grep "Map<String, Object>"
    public @ActionOutput(name = NEW_SCENE_ID_OUTPUT, type = "java.lang.Integer") Map<String, Object> createScene(
    public @ActionOutput(name = "result", type = "java.util.Map<String, Object>") Map<String, Object> calculateCheapestPeriod(
    public @ActionOutput(name = "result", type = "java.util.Map<String, Object>") Map<String, Object> calculateCheapestPeriod(
    public @ActionOutput(name = "result", type = "java.util.Map<String, Object>") Map<String, Object> calculateCheapestPeriod(
    public @ActionOutput(name = "result", type = "java.util.Map<String, Object>") Map<String, Object> calculateCheapestPeriod(
    public @ActionOutput(name = "errorMessages", type = "java.util.List<String>") @ActionOutput(name = "warningMessages", type = "java.util.List<String>") @ActionOutput(name = "infoMessages", type = "java.util.List<String>") @ActionOutput(name = "statusMessages", type = "java.util.List<String>") Map<String, Object> getMessages() {
    public @ActionOutput(name = "success", type = "java.lang.Boolean") @ActionOutput(name = "responseMessages", type = "java.util.List<String>") Map<String, Object> sendMessageWithResponse(
    public @ActionOutput(name = "index", type = "java.lang.Integer", label = "@text/actionOutputIndexLabel", description = "@text/actionOutputIndexDesc") @ActionOutput(name = "prev_index", type = "java.lang.Integer", label = "@text/actionOutputPrevIndexLabel", description = "@text/actionOutputPrevIndexDesc") @ActionOutput(name = "timestamp", type = "java.time.ZonedDateTime", label = "@text/actionOutputTimestampLabel", description = "@text/actionOutputTimestampDesc") @ActionOutput(name = "description", type = "java.lang.String", label = "@text/actionOutputDescriptionLabel", description = "@text/actionOutputDescriptionDesc") @ActionOutput(name = "details", type = "java.lang.String", label = "@text/actionOutputDetailsLabel", description = "@text/actionOutputDetailsDesc") Map<String, Object> readEvent(
```

developers/bindings/index.md
```diff
@@ -631,7 +631,11 @@ This is to support the old DSL rules engine and make the action available there.
 
 Each parameter of an action member method must be annotated with `@ActionInput`.
 
-If you return values, you do so by returning a `Map<String,Object>` and annotate the method itself with as many `@ActionOutput`s as you will return map entries.
+If you return values, you do so by returning a `Map<String,Object>` and annotate the method itself with `@ActionOutputs` containing as many `@ActionOutput`s as you will return map entries, e.g.
```
jlaur, line 634: Perhaps it should be clarified how to handle when this is dynamic? For example:

https://github.com/openhab/openhab-addons/blob/3cd5879209f40f0af8991d11b21ab748d31978f2/bundles/org.openhab.binding.energidataservice/src/main/java/org/openhab/binding/energidataservice/internal/action/EnergiDataServiceActions.java#L118

https://github.com/openhab/openhab-addons/blob/3cd5879209f40f0af8991d11b21ab748d31978f2/bundles/org.openhab.binding.energidataservice/src/main/java/org/openhab/binding/energidataservice/internal/action/EnergiDataServiceActions.java#L129-L138

I would assume that all possible return values should be annotated, in this simple case `CheapestStart` and `MostExpensiveStart`?

florian-h05:
To mention:

- It is not possible to have several actions with same method name.

PR #2406, opened 2024-11-16
Remove wrappers from inline transformation documentation examples

masipila:
Based on my observations on 4.3.0.M2, the inline transformations should not be like `JS(|"String has " + input.length + " characters")` like shown in the current documentation. Instead, it should be like `|"String has " + input.length + " characters"`.

I tested the syntax with DSL and ECMAScript but assume the same thing applies to Ruby and Groovy (but did not test those).

Cheers,
Markus

### DIFF
--- a/configuration/transformations.md
+++ b/configuration/transformations.md
@@ -145,7 +145,7 @@ Examples:
 ::: tab DSL
 
 ```java
-DSL(|"String has " + input.length + " characters")
+|"String has " + input.length + " characters"
 ```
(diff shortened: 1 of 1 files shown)


florian-h05 (commented):
Do you refer to Item state inline transformations or channel profile inline transformations?

florian-h05:
I checked the linked community post and you are indeed referring to profiles.
#2633 improves the overall docs.

PR #2425, opened 2024-12-13
exchange bintry for github link

drohhyn:
Bintray is long gone but still linked and recommended for new users.


stefan-hoehn (approved):
good catch, thx

PR #2426, opened 2024-12-15
Java 21 for OH5 development

holgerfriedrich:
Not sure if I managed to change it in all locations, but a first try...

### DIFF
--- a/configuration/migration/index.md
+++ b/configuration/migration/index.md
@@ -46,6 +46,8 @@ You should probably do your usual OS backup too to make sure you can get back to
 
 openHAB (since version 4) requires Java 17. Most modern Linux OS have an `openjdk-17-jdk` package available for you to install but alternative solutions such as [Zulu's OpenJDK](https://www.azul.com/downloads/zulu-community/?version=java-17-lts&package=jdk) (which openHABian can install for you) are available and these may run faster on 32-bit armhf platforms such as the Raspberry Pi.
 
+Upcoming openHAB 5 development requires Java 21.
+
 You must also be on the relevant repository for openHAB, this hasn't changed for openHAB 4.x so the [same instructions for moving to it](https://www.openhab.org/docs/installation/linux.html#package-repository-installation) are still valid.
 
 For DEB based installers (apt), your `.list` file should contain the line:
--- a/configuration/migration/index.md
+++ b/configuration/migration/index.md
@@ -46,7 +46,7 @@ You should probably do your usual OS backup too to make sure you can get back to
 
 openHAB (since version 4) requires Java 17. Most modern Linux OS have an `openjdk-17-jdk` package available for you to install but alternative solutions such as [Zulu's OpenJDK](https://www.azul.com/downloads/zulu-community/?version=java-17-lts&package=jdk) (which openHABian can install for you) are available and these may run faster on 32-bit armhf platforms such as the Raspberry Pi.
 
-Upcoming openHAB 5 development requires Java 21.
+openHAB 5 development requires Java 21.
 
 You must also be on the relevant repository for openHAB, this hasn't changed for openHAB 4.x so the [same instructions for moving to it](https://www.openhab.org/docs/installation/linux.html#package-repository-installation) are still valid.
 
--- a/installation/index.md
+++ b/installation/index.md
@@ -42,10 +42,10 @@ Oracle Java is also suitable for most configurations but it's not recommended. L
 
 ::: warning
 Java 17 is recommended for openHAB 4.x series.
-OpenHABian can install Java 21, but only on the native 64-bit image.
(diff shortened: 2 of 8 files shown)


holgerfriedrich:
@florian-h05 @kaikreuzer 
We probably need something like this in the official docs as well....
Sorry, I missed that until now.

configuration/migration/index.md
```diff
@@ -46,6 +46,8 @@ You should probably do your usual OS backup too to make sure you can get back to
 
 openHAB (since version 4) requires Java 17. Most modern Linux OS have an `openjdk-17-jdk` package available for you to install but alternative solutions such as [Zulu's OpenJDK](https://www.azul.com/downloads/zulu-community/?version=java-17-lts&package=jdk) (which openHABian can install for you) are available and these may run faster on 32-bit armhf platforms such as the Raspberry Pi.
 
+Upcoming openHAB 5 development requires Java 21.
```
jlaur, line 49: Maybe it would be better to not use the word "upcoming" so we don't have to come back to this later when it's no longer upcoming? And I guess also "development" can be omitted, since this page is about migration? Looking at the sentence above, maybe:

```suggestion
openHAB 5 requires Java 21.
```

?

(the sentence above regarding Java 17 for openHAB 4 might also need an update, but this is unrelated to your changes)

installation/index.md
```diff
@@ -41,10 +41,11 @@ You could also [download Azul Zulu](https://www.azul.com/downloads/zulu-communit
 Oracle Java is also suitable for most configurations but it's not recommended. Licensing restrictions may apply.
 
 ::: warning
-Java 17 is recommended for openHAB.
-Using Java 21 is possible, but still considered experimental.
+Java 17 is recommended for openHAB 4.x series.
 OpenHABian can install Java 21, but only on the native 64-bit image.
 Please note that versions of Java other than 17 and 21 are not supported at the moment.
+
+Upcoming openHAB 5 will require Java 21.
```
jlaur, line 48: Could be slightly rephrased to follow the same phasing as in line 44 (and dropping "upcoming"):
```suggestion
Java 21 is required for openHAB 5.
```

installation/index.md
```diff
@@ -41,10 +41,11 @@ You could also [download Azul Zulu](https://www.azul.com/downloads/zulu-communit
 Oracle Java is also suitable for most configurations but it's not recommended. Licensing restrictions may apply.
 
 ::: warning
-Java 17 is recommended for openHAB.
-Using Java 21 is possible, but still considered experimental.
+Java 17 is recommended for openHAB 4.x series.
```
jlaur, line 44:
```suggestion
Java 17 is required for openHAB 4.x series.
```

florian-h05 (commented):
Some minor comments, I would prefer to not speak about openHAB 5 in the future as these docs are the development docs which now target openHAB 5 snapshot.

configuration/migration/index.md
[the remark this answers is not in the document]
jlaur: See also https://github.com/openhab/openhab-docs/pull/2426#discussion_r1894504728

stefan-hoehn:
let me know when you want to have this merged.

@mstormi @ecdye 
I actually fell into the trap not knowing that 21 is required for openHAB 5 and when I upgraded to the latest snapshot (5, that is) with openhabian-config it wasn't working (very likely because of jdk 17). Could you detect OH 5 and then automatically upgrade to java 21 or would you prefer a different approach?

holgerfriedrich:
I will rework this once I find some time. It should have been added before we released 4.3....

@stefan-hoehn Maybe we should have pinned this forum post....
https://community.openhab.org/t/openhab-5-x-snapshot-requires-java-21/160890

stefan-hoehn:
> Maybe we should have pinned this forum post....

You mean, pin it in the forum? Yep, would be good idea. I think only a discourse admin can do it, AFAIK.

holgerfriedrich:
I have updated this to match the situation after release of 4.3.
Still not perfect, still struggling where to put remarks about 4.3 and where about 5.0.

Not sure where to put notes about dev process for binding developers - maybe better to be put in the README of the add-ons repo. https://community.openhab.org/t/openhab-5-x-snapshot-requires-java-21/160890/5

configuration/migration/index.md
florian-h05, line 49:
```suggestion
openHAB 5 development requires Java 21.
```

florian-h05 (changes requested):
LGTM, apart from one minor comment:

configuration/migration/index.md
holgerfriedrich: done!

florian-h05 (approved):

stefan-hoehn (approved):
LGTM, Thx

PR #2427, opened 2024-12-16
[blockly] persistence documentation update

mherwege:
Related to https://github.com/openhab/openhab-webui/pull/2732

This PR updates the Blockly persistence documentation with enhancements available in OH4.3:
- Add documentation for additional persistence blocks
- Add extra methods and parameters to existing blocks
- Update images

@stefan-hoehn FYI


configuration/blockly/rules-blockly-persistence.md
```diff
@@ -34,40 +34,49 @@ More about that topic can be viewed at ![youtube](../images/blockly/youtube-logo
 
 ![statistical-value](../images/blockly/blockly-persistence-get-statistical-value.png)
 
-_Function:_ computes any of the below functions for the given item since the time provided by _ZonedDateTime_-Block
+_Function:_ computes any of the below functions for the given item since the time provided by _ZonedDateTime_-Block, until the time provided by _ZonedDateTime_-Block, or between 2 timepoints provided by _ZonedDateTime_-Blocks.
```
stefan-hoehn, line 37: Can you add an image like this that demonstrated "between"?

image

configuration/blockly/rules-blockly-persistence.md
```diff
@@ -34,40 +34,49 @@ More about that topic can be viewed at ![youtube](../images/blockly/youtube-logo
...
 openHAB supports history and future values.
 A typical example for future values is a weather forecast.
-Due to adding future values in openHAB the amount of attributes has been vastly increased and the names had to be renamed to become more specific in terms of historic and future states.
+
+For each of the functions, you can choose the output type:
+
+- String: output as a string value
+- Quantity: output as a quantity type with unit
+- Number: output as a number
+- Timestamp: time of the persisted value (when applicable)
```
stefan-hoehn, line 47: Also here (I am bit behind, so it misses one entry=

image

configuration/blockly/rules-blockly-persistence.md
```diff
@@ -84,20 +93,44 @@ Notes: in case no or 0 values are retrieved, make sure that the item in question
...
+Persist State or States for an Item.
+
+Multiple options are available:
+
+- state (at current time): persist given State at the current time
+
+Following options are not supported by all persistence services (for example the standard rrd4j does _not_ support this).
+
+- state (at specific time): persist given State at given time
+- list of states (adding): persist list of States, not replacing the already persisted States
```
stefan-hoehn, line 122: Do you mind giving a example where this needed or helpful of if explained somewhere else link to it? I don't even know what that is for (sorry 😬)

stefan-hoehn (changes requested):
Thanks, Mark, I really appreciate that in particular because Blockly docs require a lot of images which is a lot of work !

configuration/blockly/rules-blockly-persistence.md
mherwege: True, although not all functions have it, but I will replace it.

mherwege: Ah OK, thought I had an image in for that, but didn't. The point is: Timestamp is only relevant for some of the analytic actions. E.g., it is not for sum, average, medium..., but it is for getting a persisted state, getting the last update, next update, minimum, maximum... as these results are at a specific point in time. Only when Timestamp is relevant, it is shown in the list. I didn't want to explicitely document this, is the UI should be very much self explaining. And Timestamp will be there when it is relevant. I can put in an image, but not sure it adds much.

configuration/blockly/rules-blockly-persistence.md
mherwege: It looks like documentation to link to is meager at best. This has been introduced in the context of the forecast persistence strategy, where it became useful to persist/retrieve a series of persisted values, not just one. I will try to formulate something to make it clearer.

mherwege:
@stefan-hoehn I believe I addressed your comments.

stefan-hoehn (approved):
Awesome, thanks a lot!

mherwege:
@stefan-hoehn Yes, but I don´t understanding the error in the log. Do you have any idea?

stefan-hoehn:
Yep, I checked your PR and could see that at least the one image name misses the _s_

image

which is why the build fails to find the image.

mherwege:
> Yep, I checked your PR and could see that at least the one image name misses the _s_

@stefan-hoehn Thanks for the tip. It should be better now.

stefan-hoehn (approved):
Thx for all the effort! <3

PR #2429, opened 2024-12-17
Set nginx proxy HTTP version to 1.1 in reverse proxy config

florian-h05:
HTTP 1.1 is required for WebSockets to work.
As the new log viewer uses the log WebSocket, it is required to modify the reverse proxy config.

### DIFF
--- a/installation/reverse-proxy.md
+++ b/installation/reverse-proxy.md
@@ -46,6 +46,7 @@ server {
     add_header 'Access-Control-Allow-Methods' 'GET,POST,OPTIONS,PUT,DELETE,PATCH' always;
 
     location / {
+        proxy_http_version                    1.1;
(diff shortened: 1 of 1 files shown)


florian-h05:
@kaikreuzer Can you please modify the release notes and (properly to the breaking changes section even though this is not really breaking) that when using nginx as reverse proxy, it is required to set the proxy HTTP version to 1.1 like in this PR?

florian-h05:
@mstormi
IIRC openHABian provides an option to install and configure nginx, so this here might be interesting for you.

stefan-hoehn (approved):
LGTM

mstormi:
thx for the heads up. I've added that to openHABian's default nginx config

PR #2444, opened 2025-01-04
Update doc preview infrastructure

florian-h05:
Align the used NodeJS and Ruby versions and scripts with the website build.


stefan-hoehn (commented):
Tried it out, works locally on my machine.

stefan-hoehn (approved):

PR #2453, opened 2025-01-25
Update armbian.md

cdwijs:
It's not possible to run Linux on half a MB of RAM. Is 512MB the correct number?


stefan-hoehn:
@ecdye @mstormi I think he is right, WDYT?

mstormi:
of course he is. Make it MB not Mb

cdwijs:
> of course he is. Make it MB not Mb

Done

stefan-hoehn (approved):

PR #2466, opened 2025-02-22
Semantic Tags: developer information/rules

andrewfg:
This PR improves the documentation for Semantic Tags in two respects:

1. Rules governing how developers/reviewers shall use the set of OH Core semantic tags in addon development.
2. Rules governing how developers/reviewers shall extend (or prune) the set of OH Core semantic tags.

The purpose is to have a clear set of rules for Semantic Tags as the fundamental basis for judging a number of current open PRs that aim to refine/improve/extend the current Semantic Tags functionality.

Follows on from the the discussion in https://github.com/openhab/openhab-core/issues/4619


developers/bindings/thing-xml.md
```diff
@@ -220,8 +220,13 @@ The following snippet shows a 'Control' and 'Light' tags definition:
 </tags>
 ```
 
-Please note that only [Point](https://openhab.org/javadoc/latest/org/openhab/core/semantics/model/point/package-summary.html) and/or [Property](https://openhab.org/javadoc/latest/org/openhab/core/semantics/model/property/package-summary.html) tags from the pre-defined tag library should be used.
-See also the [Semantic Model]({{base}}/tutorial/model.html) for more information on how tags are used with items.
+There are two possible types of tag -- namely "Semantic" and "Non Semantic" tags.
+The former are used by the openHAB User Interface to create automatic groupings of items according to the [Semantic Model]({{base}}/tutorial/model.html).
+The latter may (optionally) be used for any other tagging purpose at the discretion of the developer or the user.
+The list of Semantic tags is [here](https://github.com/openhab/openhab-core/blob/main/bundles/org.openhab.core.semantics/model/SemanticTags.csv).
```
jimtng, line 226:
```suggestion
The list of pre-defined Semantic tags is [here](https://github.com/openhab/openhab-core/blob/main/bundles/org.openhab.core.semantics/model/SemanticTags.csv).
```

developers/bindings/thing-xml.md
```diff
@@ -220,8 +220,13 @@ The following snippet shows a 'Control' and 'Light' tags definition:
...
+Note that the contents of this list are dynamic, and it may be extended from time to time.
+Binding developers can provide Semantic tags in the xml file for `property`, `point` and `equipment`.
```
jimtng, line 228: We don't mention Equipment yet until such time we actually support it.

developers/bindings/thing-xml.md
```diff
@@ -220,8 +220,13 @@ The following snippet shows a 'Control' and 'Light' tags definition:
 </tags>
 ```
 
-Please note that only [Point](https://openhab.org/javadoc/latest/org/openhab/core/semantics/model/point/package-summary.html) and/or [Property](https://openhab.org/javadoc/latest/org/openhab/core/semantics/model/property/package-summary.html) tags from the pre-defined tag library should be used.
-See also the [Semantic Model]({{base}}/tutorial/model.html) for more information on how tags are used with items.
+There are two possible types of tag -- namely "Semantic" and "Non Semantic" tags.
```
jimtng, line 223: IMO, these should only be used to add "Semantic" tags. We don't want bindings to start injecting non semantic tags too, do we?

developers/bindings/thing-xml.md
```diff
@@ -817,7 +822,7 @@ The full Java API for bridge and _Thing_ descriptions can be found in the Java p
...
 | tags                       | A list of default tags to be assigned to bound items                                                                                                                                                                                | optional                                      |
-| tag                        | A tag semantically describes the feature (typical usage) of the channel e.g. AlarmSystem. There are no pre-default tags, they are custom-specific                                                                                   | mandatory                                     |
+| tag                        | A tag semantically describes the feature (typical usage) of the channel e.g. AlarmSystem. There are two types of tags -- namely "Semantic" and "Non Semantic". See [Default Tags](#default-tags) below.                             | mandatory                                     |
```
jimtng: Also here, we just want Semantic tags I think?

developers/bindings/thing-xml.md
```diff
@@ -220,8 +220,13 @@ The following snippet shows a 'Control' and 'Light' tags definition:
 </tags>
 ```
 
-Please note that only [Point](https://openhab.org/javadoc/latest/org/openhab/core/semantics/model/point/package-summary.html) and/or [Property](https://openhab.org/javadoc/latest/org/openhab/core/semantics/model/property/package-summary.html) tags from the pre-defined tag library should be used.
-See also the [Semantic Model]({{base}}/tutorial/model.html) for more information on how tags are used with items.
```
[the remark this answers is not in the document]
andrewfg: > these two lines should stay.

The first two links are broken. And the last one is indeed kept (albeit with different wording).

developers/bindings/thing-xml.md
andrewfg: I dunno. That seems to have been the opinion of some others on the community. But I would happily go for Semantic ones only. (Maybe say that if a developer finds a new semantic tag that they want added, then they should include a PR towards OH core for that.

developers/bindings/thing-xml.md
andrewfg: See my above comment about how developers should add non (yet) pre-defined semantic tags.

developers/bindings/thing-xml.md
andrewfg: https://github.com/openhab/openhab-docs/issues/2467

andrewfg: But removed from this PR for the time being..

developers/bindings/thing-xml.md
andrewfg: > IMO, these should only be used to add "Semantic" tags. We don't want bindings to start injecting non semantic tags too, do we?

THERFORE we can (and should) include only the allowed point and property tags in the XSD schema. This will indeed prevent developers from adding other stuff, and force a degree of quality control.

NB and we should make a note that whenever the .csv file is updated the XSD should be updated too.

andrewfg:
^
Yes. I will add them.

andrewfg:
> Also not expected to set "Status" tag on all read-only channel types.

@lolodomo can you please explain the above rule in more detail?

andrewfg:
I added the rules from https://github.com/openhab/openhab-addons/issues/12262 .. however I am not sure if I fully understand them all (in particular the last one) => @lolodomo can you please check them?

andrewfg:
Reviewers please IGNORE the lint error concerning numbered lists. Because in this case the numbering MUST flow on from one section to the next -- rather than re-starting from 1. in each section.

developers/bindings/semantic-tags.md
```diff
@@ -0,0 +1,73 @@
...
+1. New tags MUST fit within the hierarchy of semantic tags already defined in openHAB Core.
+1. New tags SHALL NOT be synonyms of existing tags e.g. kitchen/kitchenette, sitting/living room.
+1. New tags MUST prove that PROPERTY, POINT and EQUIPMENT tags could be applied to three or more bindings.
+1. Developers MUST provide supporting evidence in their PR that the rules above and below are met.
+
+## Equipment Tags
+
+An EQUIPMENT tag describes the equipment type of a piece of equipment. Examples are light bulb, thermostat, pump, car, wall switch, push button, alarm system, or alarm detector.
+
+16. New EQUIPMENT tags MUST describe the equipment type of a physical or virtual device.
```
rkoshak, line 51: Should this be 1 instead of 16?

developers/bindings/semantic-tags.md
```diff
@@ -0,0 +1,73 @@
...
+1. New EQUIPMENT tags MUST be a noun.
+
+## Point Tags
+
+A POINT is a tag that describes the functional type (or purpose) of a data point within an equipment. Examples are measurement, control, set point, or status.
```
rkoshak, line 56: separate sentences

developers/bindings/semantic-tags.md
```diff
@@ -0,0 +1,73 @@
...
+
+18. New POINT tags MUST describe the functional type (or purpose) of the data point.
```
rkoshak, line 58: Should this be 11 instead of 18?

developers/bindings/semantic-tags.md
```diff
@@ -0,0 +1,73 @@
...
+1. New POINT tags MUST be what you do to, or how you get, the state of the Item.
```
rkoshak, line 59: We might need to elaborate. I'm not sure this is clear enough. Then again, the examples from above may be sufficient.

developers/bindings/semantic-tags.md
```diff
@@ -0,0 +1,73 @@
...
+
+## Property Tags
+
+A PROPERTY is a tag that describes the information type (or content type) of the data produced or consumed by a point. Examples are temperature, brightness, on-off, time stamp
+
+20. New PROPERTY tags MUST describe the information type (or content type) of the point.
```
rkoshak, line 65: Shouldn't this be 11?

developers/bindings/semantic-tags.md
```diff
@@ -0,0 +1,73 @@
...
+1. New PROPERTY tags MUST be a noun and represent some attribute commonly controlled or sensed in a home automation context.
+1. New PROPERTY tags MUST be defined in conjunction with a functional type of a POINT tag.
+
+## Location Tags
+
+A LOCATION is a tag that describes the location type of a point. Examples are ground floor, dining (eating) room, kitchen (cooking room), living room, or garage.
```
rkoshak, line 71: Separate sentences.

"location type of a point"

The most common use case is to put Equipment into a location. We should list it here and put it first. I'm not sure if it's confusing to also list location here because a location can be put in a locations. For example you might have bedroom Location Items in a TopFloor location in a Home location. But only the bedroom location would have equipment or points in it. But that might be too complicated to go into for this document. 

> location type of an equipment or a point

rkoshak (commented):
It looks good, I think there's some formatting and word choice things to address and it's good to go.

We may need to get buy in from core maintainers since they will be responsible for applying the rules in practice. Do the add-on maintainers need to buy in too?

andrewfg:
@rkoshak many thanks for the feedback. I will upload a revised version soon. However for the time being let me address your question about where semantic tags can be applied:

Currently tags can be applied to the following objects with a chain of inheritance as follows:
- `Channel-Type`: Point/Property tags applied at design time via the channel type xml
- `Channel`: Point/Property tags inherited from `Channel-Type` above, and/or applied at run time in the binding
- `Item`: Point/Property tags inherited from `Channel` above, and/or all types of tags applied at run time by users

And yes, I am indeed working on a PR to extend the application of tags with a chain of inheritance as follows:
- `Thing-Type`: Equipment tags applied at design time in the thing type xml
- `Things`: Equipment tags inherited from `Thing-Type` above, and/or applied at run time in the binding
- `Semantic Model`: Equipment tree structure ("create equipment from thing") inherited from `Thing` above, and/or by the user

developers/bindings/semantic-tags.md
andrewfg: Nope. Markdown auto- increments line numbers in numbered lists. After new paragraphs the number sequence reverts to start at 1. again, unless you explicitly force a hard starting number.

developers/bindings/semantic-tags.md
```diff
@@ -0,0 +1,73 @@
...
+
+## General Rules for Semantic Tags
+
+The various equipment, points, properties, and locations are displayed in the openHAB User Interface in a hierarchical tree structure, which is based on a hierarchy of semantic tags defined in openHAB Core. The list of semantic tags defined in openHAB Core is extensible. This document describes the developer rules for applying or extending this list.
+
+Note: the core predefined semantic tags are presented to users as propositions, suggestions, or hints. It is possible for users to override or customise such tags at their own discretion. The remainder of this document does NOT relate to user overrides or customisation. It applies solely to the openHAB Core tags.
+
+### Rules for **APPLYING** Core Predefined Semantic Tags
+
+1. Developers MUST apply core predefined tags in their addons where possible.
```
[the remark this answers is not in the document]
andrewfg: I am following the Internet Standard defined in [RFC-2119](https://datatracker.ietf.org/doc/html/rfc2119)

developers/bindings/semantic-tags.md
```diff
@@ -0,0 +1,73 @@
...
+1. Developers MAY hard code the tags in the ‘thing-type.xml’ or apply them dynamically at run time.
+1. Developers SHALL NOT hard code LOCATION tags in the ‘thing-type.xml’.
+1. Developers MAY apply LOCATION tags dynamically at run time (if the addon can determine the location).
```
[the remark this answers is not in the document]
andrewfg: > This implies that this is possible. 

Yes it is entirely possible. Now. See this comment https://github.com/openhab/openhab-core/issues/4619#issuecomment-2701487677 -- but as mentioned in that comment, perhaps we don't want to encourage it too much. => WDYT?

developers/bindings/semantic-tags.md
```diff
@@ -0,0 +1,73 @@
...
+1. Developers SHALL NOT apply overridden or customised tags in their addons.
+1. Developers MAY apply a POINT tag without a PROPERTY tag (in cases where a property makes no sense). However, in such a case the developer MAY submit a new PROPERTY tag that fits.
```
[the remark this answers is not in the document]
andrewfg: Indeed.

developers/bindings/semantic-tags.md
```diff
@@ -0,0 +1,73 @@
...
+1. Developers SHALL NOT apply a PROPERTY tag without a POINT tag.
+1. Semantic EQUIPMENT tags MUST be applied at thing level.
+1. Semantic POINT and PROPERTY tags MUST be applied at channel level.
+1. Developers MAY open a new PR to propose extensions to the core predefined tags as follows.
+
+### Rules for **EXTENDING** Core Predefined Semantic Tags
+
+11. The purpose of the semantic model is to maintain a stable structure, so it is intended that only rarely shall new tags be submitted or adopted.
```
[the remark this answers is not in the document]
andrewfg: Let's not get started on American vs British semantics. I am British. So I would probably take issue with ending sentences with with. Except when it makes sense..

developers/bindings/semantic-tags.md
andrewfg: > We might need to elaborate.

Any proposals?

developers/bindings/semantic-tags.md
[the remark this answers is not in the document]
rkoshak: When I was saying whether it's possible, I was referring to end-to-end. If I click on "add equipment to model" does the location tag influence anything on that page? If not, I think we should wait until that's implemented (presumable in MainUI) before adding this as a rule here.

developers/bindings/semantic-tags.md
rkoshak: I don't have anything yet that isn't going to be overly long. 

Maybe we just list a couple of examples.

> , e.g. a sensor reports a Measurement, a thermostat's target temperature is a Setpoint, etc.

developers/bindings/semantic-tags.md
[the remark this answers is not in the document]
andrewfg: Honestly I don't know what the UI currently can do. Perhaps you can say? But perhaps we should anyway consider whether it should or should not do it?

rkoshak:
I'm happy with all the changes.

> Also not expected to set "Status" tag on all read-only channel types.

I can't speak for @lolodomo but my understanding is this rule is to avoid the case where everyone just defaults to "Status". There Point tag should be the one closes to what the Item does. A temperature sensor should be a Measurement. However, a high temperature alert might be an Alarm. In both cases these represent a Status so Status is not incorrect. But there are better choices. 

I think the rule is to use the more appropriate tags rather than defaulting to a generic tag.

andrewfg:
I just pushed another commit with some more improvements.
- @rkoshak for the avoidance of doubt I removed the point where developers MAY add location tags; we can always add it back later if there is demand..
- @lolodomo can you please give feedback a) to the question above, and b) to the document in general?

andrewfg:
@rkoshak whilst thinking about examples of possible POINT + PROPERTY combos, it occurred to me that POINT=`LowBattery` is probably a violation of our (just defined) rules ..

I think it should be POINT=`Alarm` + PROPERTY=`Energy` 

(or POINT=`Alarm` + PROPERTY=`Voltage`).

developers/bindings/semantic-tags.md
```diff
@@ -0,0 +1,115 @@
...
+1. Developers MAY open a new PR to propose extensions to the core predefined tags as follows.
+
+### B. Rules for **EXTENDING** Core Predefined Semantic Tags
+
+The purpose of the semantic model is to provide a stable base set of tags for binding authors to use.
+Submiting and adopting new tags should be a rare occurance.
+
+1. New tags MUST fit within the hierarchy of semantic tags already defined in openHAB Core.
+1. New tags SHALL NOT be synonyms of existing tags e.g. kitchen/kitchenette, sitting/living room.
+1. New tags MUST show that new PROPERTY, POINT and EQUIPMENT tags could be applied to three or more bindings.
```
jimtng, line 74: Is the purpose of having the default semantic tags solely for the use in bindings? I'd say not. It should also contain common tags that will be "commonly used" / needed in a "typical" installation. Imagine an installation with 100% Generic MQTT based Things.

andrewfg: > Is the purpose of having the default semantic tags solely for the use in bindings? I'd say not.

Good point. I suppose it depends how you define 'user' versus 'developer'; in the cases of MQTT the 'user' is actually playing a kind of 'developer' role; or something like that.. -- But anyway let me add some wording to make that clearer.

@jimtng to help me get the wording right, can you please tell me how in case of MQTT the 'user/developer' creates their system? Do they create Channels on the MQTT system and apply the tags to the Channels (which then become inherited by the Items)? Or do they create Items on the Channels they created on the MQTT system and apply the tags to the Items?

> This rule implies that the existing list of PROPERTY, POINT, and EQUIPMENT tags is 100% complete for the end users as it is.

I believe that this is our intent. At least as far as plain main ui users are concerned (if not 'user/developers'). Or??

jimtng: > how in case of MQTT the 'user/developer' creates their system?

For me, normally "Semantic tags" isn't a part of Thing / Channel at all. It is purely on Items. So I'd just apply the tags to my items, after creating my MQTT Things / channels. My channels are all created manually, not through any discovery process, hence I said Generic MQTT.



> > This rule implies that the existing list of PROPERTY, POINT, and EQUIPMENT tags is 100% complete for the end users as it is.
> 
> I believe that this is our intent. At least as far as plain main ui users are concerned (if not 'user/developers'). Or??

How can anything ever be "complete"? And suppose it is currently not complete right now, then this particular rule prevents it from being perfected / completed.

andrewfg: > For me, normally "Semantic tags" isn't a part of Thing / Channel at all. 

That is essentially our problem. Currently very few bindings offer "Semantic tags" via their channels or things. Which means that the user has to do all the heavy lifting of figuring out an appropriate tagging scheme for their Items. Which is why so few users do it. Our purpose is that bindings offer comprehensive tagging suggestions so that the user can just press "accept".

> this particular rule prevents it from being perfected / completed. 

No. It says clearly that the OH Core set of tags is extensible. And the rules define to two cases 1) for 'users' who are expected to use the existing tag set, and 2) for 'developers' who want to request an extension of the existing tag set .

jimtng: > That is essentially our problem. Currently very few bindings offer "Semantic tags" via their channels or things. Which means that the user has to do all the heavy lifting of figuring out an appropriate tagging scheme for their Items. Which is why so few users do it. Our purpose is that bindings offer comprehensive tagging suggestions so that the user can just press "accept".

That may be the case for most of the bindings out there. But it is not the case for Generic MQTT. It seems that you have never used Generic MQTT? You define the channels yourself, so it doesn't matter whether you define the tags on the channel or the item, _you_ still define it yourself. The binding does not know what you're using it for.



> No. It says clearly that the OH Core set of tags is extensible. And the rules define to two cases 1) for 'users' who are expected to use the existing tag set,

So where would it fit in the rules when there is a need for users to use the existing tag set but finds it lacking, and the tags that he/she needs would make sense to be included in the _default_ tag set?

I'm not referring to the ability to specify custom tag set of course, that's irrelevant to this document.

andrewfg: > you have never used Generic MQTT?

In one Thing. But you are the expert. So rather than p***ing on me, can you please help me modify the rules to cover your use case.

jimtng: Maybe remove this rule, and leave it to the reviewer's discretion?

Furthermore, from the point of view of a binding developer for one particular binding, it's a bit of a stretch to have to go figure out how it can be used by other bindings.

andrewfg: I purposely used the word "could" rather than "will" .. but perhaps it's better to say "has the potential to.." -- the point is that we (not just me) want to impose a certain  entry barrier against a wild proliferation of tags..

andrewfg: @jimtng I pushed a commit just now which attempts to soften the hard requirement on three or more bindings. Can you please check it? But I am not sure if such a change also addresses you MQTT case? Perhaps not yet.

andrewfg:
@stefan-hoehn this PR is ready to merge.

stefan-hoehn (approved):
LGTM, thanks Andrew

PR #2469, opened 2025-02-23
Add Awtrix 3 logo

DrRSatzteil:
This PR adds a logo for the new Awtrix 3 binding (see https://github.com/openhab/openhab-addons/pull/18242)


stefan-hoehn:
As stated at https://www.openhab.org/docs/developer/addons/ images should be smaller than 10kb. You can also try to find an svg instead. There is a similar image that the apple app store uses which is much smaller. Maybe you submit this one.

DrRSatzteil:
I read the documentation and it says that the icon should be less than 30kb when it's an png. I will check if I can find an svg instead.

stefan-hoehn:
And yes, you are right, I misread it. 30Kb is fine. Thanks for the updated image.

DrRSatzteil:
Yeah this also looks good. I could also revert my latest commit, the image I used first was 512x512 and somewhere below the 30kb limit (I think it was 24kb or something)

stefan-hoehn:
Ok, then go for that, as long as it around 30kb, I am fine.

DrRSatzteil:
Is roughly 24kb now and has 512x512 dimensions

DrRSatzteil:
BTW: How is that icon connected to the binding shown in the Add-On Store? Does that need another change somewhere?

stefan-hoehn:
The name of the image must matching the binding id. However, as yours is mqtt.awtrix I wonder how this is picked up.
Maybe @lsiepel knows?

DrRSatzteil:
I just had a look how this was done for other similar bindings and apparently the whole binding name was used for the logo (like mqtt.homie) or there are quite a few modbus examples. I will rename the file then to match the binding name.

lsiepel:
AFAICT the full binding name must be used.

stefan-hoehn:
@DrRSatzteil The only way to see if this is correct is not only to build the docs but also the website (this can be done locally but it is a bit of an effort).  So, let's move on with (and thanks for the contribution). Have an eye on the next website build and check if it works as expected. If not, just ping me and I will dig deeper into it.

stefan-hoehn (approved):

PR #2485, opened 2025-04-19
Update item `expire` documentation

jimtng:
See https://github.com/openhab/openhab-core/pull/4724

### DIFF
--- a/configuration/items.md
+++ b/configuration/items.md
@@ -667,22 +667,34 @@ This parameter allows to post an update or command to an item after a period of 
 
 The expiration timer is started or restarted every time an item receives an update or a command _other than_ the specified "expire" update/command.
 Any future expiring update or command is cancelled, if the item receives an update or command that matches the "expire" update/command.
+See `ignoreStateUpdate` and `ignoreCommands` [configurations](#configurations-for-expire) below.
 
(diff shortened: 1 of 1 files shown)


configuration/items.md
```diff
@@ -667,22 +667,34 @@ This parameter allows to post an update or command to an item after a period of
...
 expire="1h 30m 45s"
 expire="1h05s"
 expire="55h 59m 12s"
+expire="2d 7h 59m 12s"
 ```
 
-Every part is optional, but all parts present must be in the given order (hours, minutes, seconds).
-Whitespaces are allowed between the sections.
+A non-negative expiry value can be specified using the [ISO8601 Duration format](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/time/Duration.html#parse(java.lang.CharSequence)).
+Java's `Duration` class supports days, hours, minutes, and seconds.
```
rkoshak, line 684: ISO8601 also supports fractions of a second but Expire doesn't handle anything less than a second and only whole seconds, IIRC. Should this be mentioned? 

e.g. `PT0.5S` is 500 msec.

stefan-hoehn (approved):

PR #2489, opened 2025-04-29
Rules Overview: add Python, update Ruby, JS and Groovy versions.

jimtng:
- Groovy 4.0
- ECMAScript 2024+
- Python 3.x
- Ruby 3.4 (Jruby 10)

@florian-h05, @HolgerHees, @wborn, @dilyanpalauzov

### DIFF
--- a/configuration/index.md
+++ b/configuration/index.md
@@ -20,9 +20,10 @@ In order to represent all of these, openHAB defines the following base component
 - [Groups](items.html#groups) - collections or categories containing Items
 - [Transformations](transformations.html) - Helper functions to transform your data
 - [Persistence](persistence.html) - Services to store data over time
-- [Rules](rules-dsl.html) - Automation logic, the "smart" in your Smart Home!
-- [Blockly Rules](blockly/) - Visually define the automation logic of your Smart Home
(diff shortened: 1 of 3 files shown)


dilyanpalauzov:
This speaks once about JRuby and later instead uses Ruby.  At the same time it mentions on the same line Jython and Python.

It would be good to describe with some words how Jython and Python differ - advantages and disadvantages.

HolgerHees:
> This speaks once about JRuby and later instead uses Ruby. At the same time it mentions on the same line Jython and Python.
> 
> It would be good to describe with some words how Jython and Python differ - advantages and disadvantages.

The major difference is the python version. Maybe we can add a hint that python is the successor of jython.

jimtng:
How about we don't mention Jython in the first two pages, and just list it in the table?

jimtng:
I've made some adjustments

wborn (approved):

stefan-hoehn:
Can you please rebase and see if the error goes away after the original problem which is not related to this PR has been fixed?

florian-h05 (approved):

stefan-hoehn (approved):

PR #2496, opened 2025-05-14
Tagging conventions for commonly confused cases

andrewfg:
This adds a documentation chapter for commonly confused/discussed tagging use cases.

Background: during the current ongoing mass tagging process it was discovered that there are some common use cases where code owners and code maintainers have been confused by, or raised questions about, the proper convention for common tag use cases. These cases have often been decided on-the-fly in the review comments of the respective PRs (which makes it hard to keep track of afterwards). This document captures the conclusions for posterity in one central place.

### DIFF
--- a/developers/bindings/thing-xml.md
+++ b/developers/bindings/thing-xml.md
@@ -994,3 +994,31 @@ The contents of this list are dynamic, and it may be extended from time to time.
 If you are an addon developer and you think there is something missing from the list please open an [Issue](https://github.com/openhab/openhab-core/issues) or [Pull Request](https://github.com/openhab/openhab-core/pulls) on GitHub.
 
 1. For further reading please see the [Description of the Semantic Model]({{base}}/tutorial/model.html), the [Developer Guidelines on Semantic Tags](semantic-tags.md) and the [Thing-Type and Channel-Type validation schema](https://www.openhab.org/schemas/thing-description-1.0.0.xsd).
+
+## Tagging Conventions for Commonly Confused Use Cases
+
+The following are some use cases that commonly lead to confusion.
+The purpose of this list is to provide the convention for tagging such cases:
+
+1. In the case of `Point` and `Property` tags it may help to consider the `Point` as the VERB and the `Property` as the OBJECT in a sentence.
+So a channel may make a 'Measurement' of a 'Temperature', or show the 'Status' of an operating 'Mode' etc.
+
+1. Do not confuse `Property` tags with Units of Measure.
+A `Property` tag is a WORD (see above) that describes the nature an the action being taken by its respective `Point`.
+So for example `Speed` need not be taken precisely to mean `m/sec` .. but can in general cover operations that "do much stuff in little time".
+
+1. For turning a piece of equipment on or off:
+`Switch.Light` should be used if the equipment is a light, otherwise `Switch.Power` should be used.
+
+1. For switching the operating mode of a piece of equipment (e.g. Auto/Manual, Day/Night, Disable/Enable, etc.):
+`Switch.Mode` should be used.
+
+1. For equipment having a set-point, even if the set-point is read-only for openHAB, it is OK to use `Setpoint.whatever`.
+
+1. For weather channels:
+`Forecast.whatever` is advised.
+
+1. For astronomical channels: `Calculation.whatever` should be used instead of `Forecast.whatever`.
+
+1. For equipment (e.g. fans, pumps) that can run at several speed/power/volume/flow-rates (e.g. Off/Low/Medium/High):
+`Control.Speed` should be used.
--- a/developers/bindings/thing-xml.md
+++ b/developers/bindings/thing-xml.md
@@ -1022,8 +1022,8 @@ However for **_this specific case_** ("throughput of a fan/pump") `Control.Speed
 
 1. The `Calculation` point type is used when (past or future) data is derived via a precise mathematical formula.
 By contrast the `Forecast` point type is used when (future) data is derived via a human or algorithmic estimation.
-So `Calculation` should be used for astronomical data e.g. the time of sunrise tomorrow.
-Whereas `Forecast` should be used for weather or solar forecasts, future energy prices or currency rates, etc.
+So `Calculation` should be used e.g. for astronomical data such as the time of sunrise tomorrow.
+Whereas `Forecast` should be used e.g. for weather or solar forecasts, etc.
 
 1. For tagging channels that represent entertainment media (e.g. album, artist, composer, actor, director. etc.):
 `Status.Info` should be used.
--- a/developers/bindings/thing-xml.md
+++ b/developers/bindings/thing-xml.md
@@ -1011,7 +1011,7 @@ So for example `Speed` need not be taken precisely to mean `m/sec` .. but can in
 `Switch.Light` should be used if the equipment is a light, otherwise `Switch.Power` should be used.
 
 1. For reporting or controlling the operating mode of a piece of equipment (e.g. auto/manual, day/night, disable/enable, etc.):
-If the channel has a two states (e.g. via a `Switch` type channel) then `Switch.Mode` should be used.
+If the channel has two states (e.g. via a `Switch` type channel) then `Switch.Mode` should be used.
 Or if it has multiple states (e.g. heat/cool/dry/fan/auto via a `String` type channel) then `Control.Mode` should be used.
 
 1. For equipment having a set-point, even if the set-point is read-only for openHAB, it is OK to use `Setpoint.whatever`.
@@ -1028,4 +1028,5 @@ Or if it has multiple states (e.g. heat/cool/dry/fan/auto via a `String` type ch
 `Status.Info` should be used.
 
 1. For tagging channels that relate to the progress of playing entertainment media, or of an automatic program:
-`Status.Progress` should be used.
+`Status.Progress` should be used to indicate the current progress.
+Or `Control.Progress` to change the progress (e.g. via a 'fast forward' or 'next' command).
--- a/developers/bindings/thing-xml.md
+++ b/developers/bindings/thing-xml.md
@@ -1015,8 +1015,8 @@ If the channel has two states (e.g. via a `Switch` type channel) then `Switch.Mo
 Or if it has multiple states (e.g. heat/cool/dry/fan/auto via a `String` type channel) then `Control.Mode` should be used.
 
 1. For equipment (e.g. fans, pumps) that can run at several speed/power/volume/flow-rates (e.g. Off/Low/Medium/High):
-In theory the above-mentioned "operating mode of a piece of equipment" `Control.Mode` ***could*** also be applied in this case.
-However for ***this specific case*** ("throughput of a fan/pump") `Control.Speed` should be used instead.
+In theory the above-mentioned "operating mode of a piece of equipment" `Control.Mode` **_could_** also be applied in this case.
+However for **_this specific case_** ("throughput of a fan/pump") `Control.Speed` should be used instead.
 
 1. For equipment having a set-point, even if the set-point is read-only for openHAB, it is appropriate to use `Setpoint.whatever`.
 
--- a/developers/bindings/thing-xml.md
+++ b/developers/bindings/thing-xml.md
@@ -1005,7 +1005,6 @@ So a channel may make a 'Measurement' of a 'Temperature', or show the 'Status' o
 
 1. Do not confuse `Property` tags with Units of Measure.
 A `Property` tag is a WORD (see above) that describes the nature of the action being taken by its respective `Point`.
-
 So for example `Speed` need not be taken precisely to mean `m/sec` .. but can in general cover operations that "do much stuff in little time".
 
 1. For turning a piece of equipment on or off:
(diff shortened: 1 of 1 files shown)


andrewfg:
Pinging @maniac103 / @jimtng / @lolodomo / @lsiepel / @mherwege / @jlaur / @rkoshak

jimtng:
nice, but surprised that you've only added 2 new entries here. I haven't been following this closely.

andrewfg:
> surprised that you've only added 2 new entries here

I added 8 in the first commit; another 2 in the second; i.e. total is 10 (so far..)

jimtng:
oops my bad, I looked at a single commit

developers/bindings/thing-xml.md
```diff
@@ -994,3 +994,37 @@ The contents of this list are dynamic, and it may be extended from time to time.
...
+## Tagging Conventions for Commonly Confused Use Cases
+
+The following are some use cases that commonly lead to confusion.
+The purpose of this list is to provide the convention for tagging such cases:
+
+1. In the case of `Point` and `Property` tags it may help to consider the `Point` as the VERB and the `Property` as the OBJECT in a sentence.
+So a channel may make a 'Measurement' of a 'Temperature', or show the 'Status' of an operating 'Mode' etc.
+
+1. Do not confuse `Property` tags with Units of Measure.
+A `Property` tag is a WORD (see above) that describes the nature an the action being taken by its respective `Point`.
```
jimtng, line 1007:
```suggestion
A `Property` tag is a WORD (see above) that describes the nature of the action being taken by its respective `Point`.

```

I'm not sure if this is what it's supposed to be

developers/bindings/thing-xml.md
```diff
@@ -994,3 +994,37 @@ The contents of this list are dynamic, and it may be extended from time to time.
...
+So for example `Speed` need not be taken precisely to mean `m/sec` .. but can in general cover operations that "do much stuff in little time".
+
+1. For turning a piece of equipment on or off:
+`Switch.Light` should be used if the equipment is a light, otherwise `Switch.Power` should be used.
+
+1. For switching the operating mode of a piece of equipment (e.g. Auto/Manual, Day/Night, Disable/Enable, etc.):
+`Switch.Mode` should be used.
```
jimtng, line 1014: Reading this, I'm not sure whether I should use Switch.Mode when there could be more than two modes, e.g. heat, cool, dry, fan, auto. Add some further clarifications maybe?

andrewfg: > use Switch.Mode when there could be more than two modes, e.g. heat, cool, dry, fan, auto.

I would say `Control.Mode` in that case.

> Add some further clarifications maybe?

Yes I will do it.

jimtng: So can this apply even when the linked item type is not `Switch`?

andrewfg: Yes why not? I would say `Switch.Mode` (for binary `Switch` Items) and `Control.Mode` (for non-binary `String` Items).

andrewfg:
@jimtng do you confirm that this one is now Ok? and if so @stefan-hoehn can you please merge it?

jimtng:
LGTM, but I'd like to also ask others to review, please.

andrewfg:
> but I'd like to also ask others to review, please.

@lsiepel @jlaur @lolodomo => any inputs?

stefan-hoehn (approved):

developers/bindings/thing-xml.md
```diff
@@ -994,3 +994,38 @@ The contents of this list are dynamic, and it may be extended from time to time.
...
+
+1. Do not confuse `Property` tags with Units of Measure.
+A `Property` tag is a WORD (see above) that describes the nature of the action being taken by its respective `Point`.
+So for example `Speed` need not be taken precisely to mean `m/sec` .. but can in general cover operations that "do much stuff in little time".
+
+1. For turning a piece of equipment on or off:
+`Switch.Light` should be used if the equipment is a light, otherwise `Switch.Power` should be used.
+
+1. For reporting or controlling the operating mode of a piece of equipment (e.g. auto/manual, day/night, disable/enable, etc.):
+If the channel has a two states (e.g. via a `Switch` type channel) then `Switch.Mode` should be used.
```
lsiepel:
```suggestion
If the channel has two states (e.g. via a `Switch` type channel) then `Switch.Mode` should be used.

```

developers/bindings/thing-xml.md
```diff
@@ -994,3 +994,38 @@ The contents of this list are dynamic, and it may be extended from time to time.
...
+Or if it has multiple states (e.g. heat/cool/dry/fan/auto via a `String` type channel) then `Control.Mode` should be used.
+
+1. For equipment having a set-point, even if the set-point is read-only for openHAB, it is OK to use `Setpoint.whatever`.
+
+1. For weather channels:
```
lsiepel: Would this `Forecast` also apply to future currency and/or energie cost channels?

developers/bindings/thing-xml.md
```diff
@@ -994,3 +994,38 @@ The contents of this list are dynamic, and it may be extended from time to time.
...
+`Forecast.whatever` is advised.
+
+1. For astronomical channels: `Calculation.whatever` should be used instead of `Forecast.whatever`.
```
lsiepel: Only for astronomical channels? Or can we think of some more types of use where the value is precisly calculated?

developers/bindings/thing-xml.md
```diff
@@ -994,3 +994,38 @@ The contents of this list are dynamic, and it may be extended from time to time.
...
+
+1. For equipment (e.g. fans, pumps) that can run at several speed/power/volume/flow-rates (e.g. Off/Low/Medium/High):
+`Control.Speed` should be used.
```
lsiepel: IMO i have some doubts about this, it is not so intuitive and leaves room for discussion.

What about a fan that has the modes `off, auto, 1, 2, 3` will it be tagged `Control.Mode` ? And `off, 1, 2, 3` will be tagged `Control.Speed` ?

Same for the others, speed is most of the times the result of a mode. What about a HVAC or heatpump that has the modes Off, ECO, MAX each with their own speed (flow rate or power or fan speed)

developers/bindings/thing-xml.md
```diff
@@ -994,3 +994,38 @@ The contents of this list are dynamic, and it may be extended from time to time.
...
+
+1. For tagging channels that represent entertainment media (e.g. album, artist, composer, actor, director. etc.):
+`Status.Info` should be used.
+
+1. For tagging channels that relate to the progress of playing entertainment media, or of an automatic program:
```
lsiepel, line 1031: Maybe add that  fast forward  etc can be `Control.Progress` . All that is within the current track. Can this `Progress` also be used between tracks like `Next` If so, maybe add them too.

lsiepel (changes requested):
Thanks for adding this documentation, it gives good guidance to the tagging. Left some questions.

developers/bindings/thing-xml.md
andrewfg: > Would this Forecast also apply to future currency and/or energy cost channels?

Yes. This PR is to cover _**commonly confused cases**_ and in my opinion the use of Forecast on future currency or energy prices is quite obvious (i.e. not confusing) so not needing to be mentioned here. However the fact that you ask the question indicates otherwise. => So do you want to add another sentence here?

developers/bindings/thing-xml.md
andrewfg: See comment above.

This sentence was added to explain the specific confusing use case of astro channels where e.g. the time of sunrise tomorrow is a calculation and not a forecast. (And the time of sunrise yesterday is also calculation and not a measurement..)

=> But again, do you want to add (an)other sentence(s) here to cover other possibly confusing use cases?

developers/bindings/thing-xml.md
andrewfg: > some doubts about this, it is not so intuitive and leaves room for discussion

The purpose of this PR is to expressly eliminate such discussion and give a precise answer. => So can you please suggest a text with your counter proposal?

developers/bindings/thing-xml.md
andrewfg: Makes sense to me.

developers/bindings/thing-xml.md
andrewfg: However -- **_specifically for fans_** -- my proposal is that if the channel controls or monitors something related to the velocity / rpm / throughput of an air moving device then we should use Status.Speed / Control.Speed instead of Status.Mode / Control.Mode -- note: as I said before **_specifically for fans_** ..

.. or if you insist on the contrary then this confusing case should **_specifically for fans_** state the contrary.

developers/bindings/thing-xml.md
lsiepel: Yes, that would be appreciated. `Forecast` and `Calaculation` are usually about future values and are easily mixed. Thanks.

developers/bindings/thing-xml.md
lsiepel: Yes, please. Just so that it is clear that Calculation is not limited to the astronomical related channels.

developers/bindings/thing-xml.md
lsiepel: Let me think about it, I see the issue, but creating a unambigious fix is not that simple ;-)

andrewfg: > Let me think about it

Ok. There is only one thing worse than making a hard decision .. namely NOT making any decision.

andrewfg: See the revised text in my latest commit..

developers/bindings/thing-xml.md
```diff
@@ -994,3 +994,40 @@ The contents of this list are dynamic, and it may be extended from time to time.
...
+1. For equipment (e.g. fans, pumps) that can run at several speed/power/volume/flow-rates (e.g. Off/Low/Medium/High):
+In theory the above-mentioned "operating mode of a piece of equipment" `Control.Mode` **_could_** also be applied in this case.
+However for **_this specific case_** ("throughput of a fan/pump") `Control.Speed` should be used instead.
+
+1. For equipment having a set-point, even if the set-point is read-only for openHAB, it is appropriate to use `Setpoint.whatever`.
+
+1. The `Calculation` point type is used when (past or future) data is derived via a precise mathematical formula.
+By contrast the `Forecast` point type is used when (future) data is derived via a human or algorithmic estimation.
+So `Calculation` should be used for astronomical data e.g. the time of sunrise tomorrow.
+Whereas `Forecast` should be used for weather or solar forecasts, future energy prices or currency rates, etc.
```
jlaur: I guess future energy prices are not to be confused with day-ahead electricity prices? They are published in advance (typically the day before). Once published, they are final and binding (i.e., they don’t change afterward). Once published. they are scheduled or planned prices, and in practice they are deterministic and immutable from the consumer's point of view. Therefore, typically there is no reason to have different channels for historic, current and future prices. The future prices will eventually become historic - they still won't change.

So maybe we should explicitly mention not to use `Forecast` to avoid confusion? We can refer to "day-ahead prices" (electricity/spot prices).

My main suggestion is to exclude them from `Forecast`.

If you ask about how they should then be tagged - I'm not sure. `Status`? But we don't have any `Price` property tag, so I can't recommend a full-fledged tag combination.

developers/bindings/thing-xml.md
jlaur: Please see https://github.com/openhab/openhab-docs/pull/2496#discussion_r2119243369.

developers/bindings/thing-xml.md
jlaur: I would suggest to leave out currency rates here. I'm not sure future currency exchange rates has much to do with home automation. Spot rates are the most common type, but they represent a real-time rate at which you can exchange currency now. Then there are forward rates, but they are not forecasts. Predicted/forecasted rates are forecasts, but they are something purely speculative, based on macroeconomic indicators. Used in economic models, investment outlooks, or trading tools. So I'd say usually not home automation. I got a little help from ChatGPT, I'm no expert in economics. 😉

andrewfg: @lsiepel are you happy to delete currency? I added it at your request..

andrewfg: @jlaur thanks for the inputs. The purpose on this PR is to provide "black-on-white" clarity for tags in confusing use cases. However your comment above sounds more like transcendental meditation about the meaning of the universe ;) => Can you please suggest some hard "black-on-white" text?

jlaur: You can remove the mentioning of "future energy prices" in the sentence above, and then optionally add in another sentence "should NOT be used for future energy prices".

andrewfg: > remove the mentioning of "future energy prices"

Ok.

> optionally add

Let us avoid this option: a negative statement of what it shall not be doesn't help unless there is a positive alternative proposal of what it **shall** be.

andrewfg:
@jlaur @lsiepel I removed currency and added more explanation on fans. We should try to get this into 5.0

andrewfg:
> let me know when I should merge it.

@stefan-hoehn it is ok to merge this

Issue #2499, opened 2025-05-16
Add Python to script transformation page


jimtng:
Ref: https://github.com/openhab/openhab-addons/pull/18616

Page to update: https://next.openhab.org/docs/configuration/transformations.html

Tagging @HolgerHees

PR #2500, opened 2025-06-02
add python transformation examples

HolgerHees:
this will fix https://github.com/openhab/openhab-docs/issues/2499

### DIFF
--- a/configuration/transformations.md
+++ b/configuration/transformations.md
@@ -130,6 +130,16 @@ The script file name is `stringlength.groovy` and the transformation is `GROOVY(
 
 :::
 
+::: tab Python
+
+The script file name is `stringlength.py` and the transformation is `PY(stringlength.py)`.
+
+```python
+"String has " + str(len(input)) + " characters"
+```
+
+:::
+
 ::::
 
 ### Inline Script Transformation
@@ -173,6 +183,14 @@ RB(|"String has #{input.length} characters")
 
 ```groovy
 GROOVY(|"String has ${input.length()} characters")
+```
+
+:::
+
+::: tab Python
+
+```python
+PY(|"String has " + str(len(input)) + "characters")
 ```
 
 :::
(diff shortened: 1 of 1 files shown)


HolgerHees:
@jimtng could you please review?

configuration/transformations.md
```diff
@@ -130,6 +130,16 @@ The script file name is `stringlength.groovy` and the transformation is `GROOVY(
 
 :::
 
+::: tab Python
+
+The script file name is `stringlength.py` and the transformation is `PY(stringlength.py)`.
+
+```python
+"String has " + str(len(input)) + " characters"
```
jimtng: Have you tested this? I thought you need to explicitly call `return` in python?

configuration/transformations.md
```diff
@@ -177,6 +187,14 @@ GROOVY(|"String has ${input.length()} characters")
 
 :::
 
+::: tab Python
+
+```python
+PY(|"String has " + str(len(input)) + "characters")
```
jimtng: ditto this about `return`

configuration/transformations.md
HolgerHees: yes, it is tested. It behaves like jruby and groovy. If the object has a toString it works. 

only if you have a more complex function, you need a complete function body with a return value.

details can be found here

https://github.com/openhab/openhab-python?tab=readme-ov-file#py-transformation

configuration/transformations.md
HolgerHees: this is tested too and behaves like all other languages

jimtng (approved):
Thanks for confirming. The only other comment is whether we want to put python after jsscripting, before jruby considering its popularity. If so the order should be standardized in all other (future) similar multi language tabs.

HolgerHees:
> Thanks for confirming. The only other comment is whether we want to put python after jsscripting, before jruby considering its popularity. If so the order should be standardized in all other (future) similar multi language tabs.

As a newbie, I didn't want to be so bold as to put myself in second place. :-) But of course, I hope it will become popular, but I don't know if it will happen. ;-)

stefan-hoehn (approved):

PR #2508, opened 2025-06-15
YAML configuration

jimtng:


configuration/yaml.md
```diff
@@ -0,0 +1,644 @@
...
+
+#### Variable Sources and Precedence
+
+Variables can be defined in multiple places, resolved in the following order of precedence:
+
+1. Inline `vars` in `!include` directives
+1. Global `variables` defined in the main file
+1. Local `variables` defined in the [included file](#including-other-files)
+
+Variables defined inline with `!include` take precedence over all others and apply only to that include context.
```
Nadahar: Not sure if I understand "apply only to that include context". Maybe a rephrasing would be helpful.

configuration/yaml.md
```diff
@@ -0,0 +1,644 @@
...
+
+```yaml
+keyname: !include
+  file: filename.inc.yaml
+  vars:
+    var1: value1
+    var2: value2
+```
+
+Or compact form:
```
Nadahar: I think this looks a bit strange. Maybe `##### Compact form` could be used instead?

configuration/yaml.md
jimtng: How about:

> Variables specified in the `vars` block of an `!include` directive are visible only within the included file, but they have higher precedence over global and local variables.
This enables included files to act as parameterized templates that can be reused with different values.

configuration/yaml.md
```diff
@@ -0,0 +1,644 @@
...
+```
+
+#### File Naming & Reload Behavior
+
+- Use the `.inc.yaml` or `.inc.yml` extension for included files so they aren’t loaded as primary configuration files.
+- Changes to included files automatically trigger a reload of the primary file.
+
+### Packages
+
+openHAB’s YAML configuration system supports packages—modular bundles of related configuration elements (such as Things, Items, and Tags) that define a device or feature in a reusable way.
```
Nadahar: The hyphen between "packages" and "modular" looks strange IMO. Maybe just surround it with spaces?

Nadahar: Also, maybe Things, Items and Tags should be links to the respective documentation pages?

configuration/yaml.md
```diff
@@ -0,0 +1,644 @@
...
+```
+
+🔄 Merge Behavior
+
+- Package contents are merged into the main YAML structure.
+- Scalar definitions in the main file override matching definitions from the package (by key).
+  This can be used to refine a generic package on a case-by-case basis.
+- Map and List definitions in the main file will be merged.
+
+File structure can be nested—e.g., you may organize packages under includes/ or templates/ as long as they’re excluded from default parsing by giving them a `.inc.yaml` or `.inc.yml` extension.
```
Nadahar: Spaceless hyphen again.

Nadahar: I think "includes/" and "templates/" should be in code quotes.

Nadahar (commented):
I haven't studied the implementation enough to have an opinion, but I've made some comments on the formatting/phrasing itself.

configuration/yaml.md
jimtng: The compact form is essentially still a full syntax, so it shouldn't have its own section. It's just a YAML shorthand thing.

configuration/yaml.md
Nadahar: "they have higher precedence over global and local variables": Not a native English speaker, but again not sure that I understand. Does it mean either "they have higher priority than global and local variables" or "they have precedence over global and local variables"

configuration/yaml.md
jimtng: > The hyphen between "packages" and "modular" looks strange IMO. Maybe just surround it with spaces?

It's an [em-dash](https://www.thesaurus.com/e/grammar/em-dash/)

I'll think about replacing it with alternative syntax, but no promises unless there's a consensus against it.



> Also, maybe Things, Items and Tags should be links to the respective documentation pages?

There are links to the respective documentation pages [earlier on](https://deploy-preview-2508--openhab-docs-preview.netlify.app/docs/configuration/yaml.html#general-structure) (up) in a table. I don't want to link them here so as not to distract readers from understanding the core concept of "packages". I want them to understand that package will package "stuff", doesn't matter what "stuff" actually is.

By the time they've reached the "Packages" section they should've already known what Things/Items/Tags are.

configuration/yaml.md
Nadahar: I understand that, it's just that when the font ends up being much larger for the compact form, I find it "confusing": When you use 5 #'s the font gets quite small. It's a markdown quirk I guess, the font sizes used depends on the rendering and is somewhat out of your control. I just thought it looked strange, so I made a note. Maybe there's no "better" way to do it.

configuration/yaml.md
jimtng: > Spaceless hyphen again.

See above. They are rendered as emdash in the html - a slightly longer dash. I agree, here it's not strictly necessary. I'll change this to a comma.





> I think "includes/" and "templates/" should be in code quotes.

Done.

configuration/yaml.md
jimtng: Currently I've replaced the emdash with its HTML Entity for clarification.

Nadahar: > It's an [em-dash](https://www.thesaurus.com/e/grammar/em-dash/)

Ok. I often use hyphens (of any length) as pauses when I write, but I find skipping the spaces makes it look just wrong - and hard to read. I didn't know that is "was a thing", I've seen it before, but just assumed that it was an error. To me at least, it makes it "hard to read", in that I "stumble" when I reach that part, have to stop, try to figure out what it's supposed to mean, and then continue.

configuration/yaml.md
jimtng: > I understand that, it's just that when the font ends up being much larger for the compact form, I find it "confusing": When you use 5 #'s the font gets quite small. It's a markdown quirk I guess, the font sizes used depends on the rendering and is somewhat out of your control. I just thought it looked strange, so I made a note. Maybe there's no "better" way to do it.

I see your point. I've combined them into the same code fence.

configuration/yaml.md
jimtng: I have added spaces around the emdash.

Nadahar: That made all the difference to me, not I think if "flows" just fine.

configuration/yaml.md
jimtng: Changed to

> Variables specified in the `vars` block of an `!include` directive are visible only within the included file, but they take precedence over global and local variables.

spacemanspiff2007:
I have some general remarks on the structure

I think it would be better if we move the DSL config in a sub structure. E.g.
```
Configuration Overview
 - Editors, Installation of Addons
Configuration by openHAB syntax (since OH1)
- Things, Items, Persistence, Transformations
Configuration by YAML
- Configuration
- Extensions
Rules
- Textual Rules
- Blockly Rules
- JSR223 Scripting
```

Additionally I think that `REST API`, `Websocket API` and `Api Topk Generation` should be in a main section `openHAB API`.

----

In the quick example and structure I would always move the tags to the bottom. Users are mainly looking for Things / Items configuration.

---

YAML Extensions should be a separate page

configuration/yaml.md
```diff
@@ -0,0 +1,642 @@
...
+        itemDimension: <item dimension> # Only if itemType is "Number"
+        label: <channel label>
+        description: <description>
+        config:
+          <channel config key>: <config value>
+```
+
+| Configuration Key | Description                                                                                                                                   |
+|:------------------|:----------------------------------------------------------------------------------------------------------------------------------------------|
+| thing_uid         | The UID of the Thing or Bridge being defined. The UID format is `binding_id`:`thing_type_id`:[`bridge_id`:]`thing_id`, e.g. `astro:sun:home`. |
```
spacemanspiff2007: How about `<binding_id>:>thing_type_id>:[<bridge_id>:]<thing_id>`. The many changes in style look strange.
I think it would also be better if its `<thing_uid>` instead of "thing_uid"

configuration/yaml.md
```diff
@@ -0,0 +1,642 @@
...
+| `${var}`          | Inserts the value of var, or an empty string if undefined. |
+| `${var-default}`  | Uses default if var is undefined.                          |
+| `${var:-default}` | Uses default if var is undefined or empty.                 |
+| `${var${nested}}` | Supports nested interpolation up to 10 levels deep.        |
+
+Notes:
+
+- Substitutions only work in double-quoted or unquoted strings.
+- Single-quoted strings are treated as literals and will not be interpolated.
+- Substitutions are also performed in key names and in default values, e.g. `${var:-${variable_default}}`.
```
spacemanspiff2007: This block should be directly in Variable Substitions

configuration/yaml.md
```diff
@@ -0,0 +1,642 @@
...
+##### Full Syntax
+
+```yaml
+keyname: !include
+  file: filename.inc.yaml
+  vars:
+    var1: value1
+    var2: value2
+
+# Or compact form:
```
spacemanspiff2007: `# Or yaml inline form:`

configuration/yaml.md
```diff
@@ -0,0 +1,642 @@
...
+      name: Living_Room_Light
+      label: Living Room Light
+
+  # Reuse the same template
+  bedroom-light: !include
+    file: templates/mqtt-light.inc.yaml
+    vars:
+      thingid: bedroom-light
+      name: Bed_Room_Light
+      label: Bed Room Light
```
spacemanspiff2007:
```yaml
# here with inline form
vars: {thingid: bedroom-light, name: Bed_Room_Light, label: Bed Room Light}
```

configuration/yaml.md
jimtng: Can you please elaborate?

configuration/yaml.md
jimtng: Can you point me to where this term "inline form" is used? Or did you mean "one-liner" ?

IMO, "compact form" is more apt here.

configuration/yaml.md
jimtng: To what end? I'm trying to demonstrate the concept of packaging here, and I don't want to conflate the example with different concepts.

jimtng:
> I think it would be better if we move the DSL config in a sub structure. E.g.
> 
> ```
> Configuration Overview
>  - Editors, Installation of Addons
> Configuration by openHAB syntax (since OH1)
> - Things, Items, Persistence, Transformations
> Configuration by YAML
> - Configuration
> - Extensions
> Rules
> - Textual Rules
> - Blockly Rules
> - JSR223 Scripting
> ```
> 
> Additionally I think that `REST API`, `Websocket API` and `Api Topk Generation` should be in a main section `openHAB API`.

This should be a separate issue / PR. I don't want to mix this restructure and the new documentation in a single PR.

I do agree that the structure could be improved.

configuration/yaml.md
spacemanspiff2007: The first two entries of the list should be at line 329.
It generally describes where substitutions work

configuration/yaml.md
jimtng: Thanks, I've changed it as suggested.

configuration/yaml.md
jimtng: I see your point, but I'd like to keep them as is for several reasons:
- One could view them as part of "syntax"
- Alternatively you could see the "Notes" as additional notes after having read both the introduction about Variables and the syntax.
- I'd like to keep the introduction simple as it flows onto the syntax before introducing extra complexities so that it's an easier flow to read.
- Having one "Notes" section after intro then another "Notes" after syntax makes for "Notes overload". It may be a different case if it was a huge notes that could benefit from being split up.

rkoshak:
> This should be a separate issue / PR. I don't want to mix this restructure and the new documentation in a single PR.

Even without a restructure, I would have expected this new page to fall under Configuration Overview, not at the same level as it. Everything that the new YAML does has it's DSL versions documented there.

I agree that the whole structure needs to be reworked and that should go in a different PR. I was surprised on where the YAML docs fell in the structure though. But it's weird for Actions to fall outside Configuration Overview too so :man_shrugging: .

configuration/yaml.md
```diff
@@ -0,0 +1,640 @@
...
+| `readOnly`                           | An optional key which when set to `true` (the default) would declare that the file may not be modified by openHAB. |
+| [variables](#variable-substitutions) | Variables for Substitutions                                                                                        |
+| [packages](#packages)                | Packages                                                                                                           |
+| [tags](#tags)                        | Custom Semantic Tags                                                                                               |
+| [things](#things)                    | openHAB [Things]({{base}}/concepts/things.html)                                                                    |
+| [items](#items)                      | openHAB [Items]({{base}}/concepts/items.html)                                                                      |
+
+The YAML files in general must follow the standard YAML syntax, with a few openHAB-specific features:
```
rkoshak, line 25: Does order matter? Is it valid, for example, to put the version tag at the end of the file? We should make a note one way or the other.

jimtng:
> Even without a restructure, I would have expected this new page to fall under Configuration Overview

But it _is_ under / inside Configuration Overview. It's the last item. Its siblings are Addons, Things, Items, etc.

configuration/yaml.md
```diff
@@ -0,0 +1,640 @@
...
+| `label`           | Item label. Unlike the DSL syntax in an `.items` file, the label here must only contain the label without any formatting syntax.                                                                                                                                                 |
+| `icon`            | The item's icon.                                                                                                                                                                                                                                                                 |
+| `groups`          | The list of parent group names.                                                                                                                                                                                                                                                  |
+| `tags`            | The list of tags.                                                                                                                                                                                                                                                                |
```
rkoshak: Should we mention this can be both semantic tags and non-semantic tags?

configuration/yaml.md
```diff
@@ -0,0 +1,640 @@
...
+| `autoupdate`      | A short form for the `autoupdate` metadata. Valid values are: `true` or `false`.                                                                                                                                                                                                 |
+| `channel`         | A short form for the `channels` list to define a linked channel UID for when there is only one channel with a blank config.                                                                                                                                                      |
+| `channels`        | A key-value map to define the channels linked from this item.                                                                                                                                                                                                                    |
+| `metadata`        | A key-value map to define item metadata. The metadata defined in here overrides any short forms when both are specified.                                                                                                                                                         |
+
```
rkoshak, line 252: I think we need to discuss profiles before this final example. Can I use a profile with `channel` or only `channels`? Where do I go to figure out what the keys I need to use are defined?  These are the questions I'm left without answers.

configuration/yaml.md
jimtng: Good question. I've added "in no particular order" when introducing the list of keys (before the table)

configuration/yaml.md
jimtng: Thanks for the feedback. I've added some explanations + links after the example. Let me know whether it provides sufficient information to answer potential questions.

configuration/yaml.md
jimtng: I've added this information

configuration/yaml.md
Nadahar: Looking at the code, the document is first parsed into a hierarchy of `JsonNode`s. `version` is then retrieved from the "root level" of the hierarchy:
```java
                JsonNode versionNode = fileContent.get(VERSION);
                if (versionNode == null || !versionNode.canConvertToInt()) {
                    logger.warn("Version is missing or not a number in model {}. Ignoring it.", modelName);
                    removeModel(modelName);
                    return;
                }
```
It therefore should not matter where in the file it is placed, as long as it's at the root level.

configuration/yaml.md
jimtng: See my comment in the main reply too.

jimtng:
@spacemanspiff2007:

> ```yaml
> # here with inline form
> vars: {thingid: bedroom-light, name: Bed_Room_Light, label: Bed Room Light}
> ```


The esphome documentation includes a brief introduction on YAML in general. Such section could also be used to introduce some YAML "cheatsheet" which is what this would fall into. It's something that I have avoided in order to keep the page from being too long. If we want to split anything, that's the page that should be split up instead of splitting up the "YAML Extensions" section. I initially wanted to just add a link to some external page about this, but I haven't found a suitable one to link to. Candidates include the official YAML documentation (too detailed and too technical), wikipedia yaml page (not quite in the style I want), and various blogs (too unofficial).

> YAML Extensions should be a separate page

I feel that the YAML Extensions part is a fundamental part of the documentation that should be kept as one page. This is of course a personal opinion only.

jimtng:
> In the quick example and structure I would always move the tags to the bottom. Users are mainly looking for Things / Items configuration.

I tried this just now, but it feels weird. Logically we should create the tags first before referring to them in Items, even though it should also work when the tags are created later.

Although entities can be defined in any order, they are processed in order, so when you have the `tags:` section last, the tags won't actually be created until after the things and items were created. The end result is still the same though, and it doesn't seem to matter if tags are defined last in the file.

@rkoshak wdyt about moving the tags section last (after things and items) in the doc?

While testing this, I discovered a bug in the Main UI -> Item Details page. When you refresh the page, the "Semantic class" would turn into "None". This is unrelated to the yaml config. It happens on UI items + built in semantic tag too.

rkoshak:
I didn't have a strong opinion either way. I can see some good reasons for putting tags last and some good reasons to put them up front. 

If I think about it from the order users works learn about the concepts, tags should probably go last as defining custom tags would be learned will after defining things and items.

jimtng:
I've moved the tags to the end now

jimtng:
> If I think about it from the order users works learn about the concepts, tags should probably go last as defining custom tags would be learned will after defining things and items.

IMO, this isn't about "learning concepts". This is the reference documentation. What we wrote here will be the model for most users, beginners and seasoned users included. So I still personally prefer tags to be first.

spacemanspiff2007:
> I feel that the YAML Extensions part is a fundamental part of the documentation that should be kept as one page. This is of course a personal opinion only.

It's an awesome feature. But I think it deserves its own page because its not bound to (doesn't know) how the configuration is done and is a topic on it's own. It's basically a stand alone feature and not related to the configuration structure.

For the user this would be:
learn yaml (config) -> learn to make it smarter (extension) | helper (cheatsheet) which can be referenced from the fist two

But I'm also a fan of shorter pages where I don't have to scroll so much :shrug:

configuration/yaml.md
```diff
@@ -0,0 +1,660 @@
...
+          stateTopic: bedroom1-switch/button1
+          postCommand: true
+      statusled:
+        type: switch
+        config:
+          stateTopic: bedroom1-switch/light/status_led/state
+          commandTopic: bedroom1-switch/light/status_led/command
+          transformationPattern:
+            - JSONPATH:$.state
+          formatBeforePublish: '{"state": "%s"}' # This needs to be quoted to be parsed as a string
```
spacemanspiff2007, line 182: quoted with either ' or "

configuration/yaml.md
```diff
@@ -0,0 +1,660 @@
...
+      - group1
+      - group2
+    tags:
+      - tag1
+      - tag2
+    format: <format pattern> # short form for state description's pattern
+    unit: <unit> # short form for the unit metadata
+    autoupdate: true|false # short form for the autoupdate metadata
+    channel: <channel_uid> # short form to define an item-channel link if there is only one channel
+    channels:
```
spacemanspiff2007:
```yaml
channel: <channel_uid> # short form to define an item-channel link if there is only one channel or no channel config
channels:              # long form if there are more than one channels or a channel profile is used
``

File: configuration/yaml.md [hunk 14]
```diff
@@ -0,0 +1,660 @@
...
+
+  BedRoom1_Switch_StatusLED:
+    type: Switch
+    groups:
+      - BedRoom1_Switch_Group
+    channels:
+      mqtt:topic:bedroom1-switch:statusled:
+        # The configuration of this item-channel link is defined below
+        profile: transform:MAP
+        function: "|open=ON;closed=OFF" # | is a YAML special character, so enclose it in quotes.
```
spacemanspiff2007, line 300: what does "|" do?

configuration/yaml.md
jimtng: It's a bit more complicated, and I don't want to be too verbose here. If it's quoted using double quotes, the double quotes inside needs to be escaped. So just saying `either ' or "` may lead to more problems. 

I expect people to just intuitively know what to use, or read up on a more detailed explanation on yaml. Therefore I think I'll leave this comment as is.

configuration/yaml.md
jimtng: Again, this will be a part of a more detailed explanation about YAML syntax. Officially it's https://yaml.org/spec/1.1/.

configuration/yaml.md
jimtng: I've changed the comments to clarify

mherwege:
> While testing this, I discovered a bug in the Main UI -> Item Details page. When you refresh the page, the "Semantic class" would turn into "None". This is unrelated to the yaml config. It happens on UI items + built in semantic tag too.

@jimtng Are you sure it had a valid semantic class in the first place? I can't reproduce this on my side with a snapshot I installed this morning. But there may be something I have missed.

jimtng:
If https://github.com/openhab/openhab-core/pull/4818 doesn't make it before feature freeze, we'll have to remove the yaml extension from the docs.

jimtng:
> > While testing this, I discovered a bug in the Main UI -> Item Details page. When you refresh the page, the "Semantic class" would turn into "None". This is unrelated to the yaml config. It happens on UI items + built in semantic tag too.
> 
> @jimtng Are you sure it had a valid semantic class in the first place? I can't reproduce this on my side with a snapshot I installed this morning. But there may be something I have missed.

Yes, it had a valid semantic class. It showed it on the item details / edit page, but when I hit Cmd+R to refresh the page, the semantic class turned to None. I have to get out of the page (click Cancel on the item edit page), then click Edit again (or Details if the item is file-based) in order to show the proper semantic tag again.

configuration/yaml.md
spacemanspiff2007: I know the block text with the |  operator in yaml but what does the "|" in the transform:MAP function do?

jimtng: https://www.openhab.org/addons/transformations/map/#inline-map

mherwege:
> > > While testing this, I discovered a bug in the Main UI -> Item Details page. When you refresh the page, the "Semantic class" would turn into "None". This is unrelated to the yaml config. It happens on UI items + built in semantic tag too.
> > 
> > 
> > @jimtng Are you sure it had a valid semantic class in the first place? I can't reproduce this on my side with a snapshot I installed this morning. But there may be something I have missed.
> 
> Yes, it had a valid semantic class. It showed it on the item details / edit page, but when I hit Cmd+R to refresh the page, the semantic class turned to None. I have to get out of the page (click Cancel on the item edit page), then click Edit again (or Details if the item is file-based) in order to show the proper semantic tag again.

I can reproduce it now. It doesn't happen when you have the development tools open in the browser, that's why I didn't notice it. I still have open PR's (https://github.com/openhab/openhab-webui/pull/3178) that replace this with a semantics picker. I will see if I can reproduce it there.

mherwege:
> > > > While testing this, I discovered a bug in the Main UI -> Item Details page. When you refresh the page, the "Semantic class" would turn into "None". This is unrelated to the yaml config. It happens on UI items + built in semantic tag too.
> > > 
> > > 
> > > @jimtng Are you sure it had a valid semantic class in the first place? I can't reproduce this on my side with a snapshot I installed this morning. But there may be something I have missed.
> > 
> > 
> > Yes, it had a valid semantic class. It showed it on the item details / edit page, but when I hit Cmd+R to refresh the page, the semantic class turned to None. I have to get out of the page (click Cancel on the item edit page), then click Edit again (or Details if the item is file-based) in order to show the proper semantic tag again.
> 
> I can reproduce it now. It doesn't happen when you have the development tools open in the browser, that's why I didn't notice it. I still have open PR's ([openhab/openhab-webui#3178](https://github.com/openhab/openhab-webui/pull/3178)) that replace this with a semantics picker. I will see if I can reproduce it there.

I can't reproduce with the PR applied, so if the PR makes it to OH 5.0, I believe this would be solved.

jimtng:
@lolodomo would you mind taking a look at this please?

lolodomo:
> @lolodomo would you mind taking a look at this please?

Yes, I will. This PR must be merged before next Sunday.

stefan-hoehn:
LGTM so far. Ping me when this needs to be merged.

configuration/yaml.md
```diff
@@ -0,0 +1,338 @@
...
+  TV_Power:
+    type: Switch
+    label: TV Power
+    icon: television
+    autoupdate: false
+    channel: lgwebos:WebOSTV:3aab9eea-953b-4272-bdbd-f0cd0ecf4a46:power
+
+  TV_Volume:
+    type: Dimmer
+    label: Volume [%d]
```
lolodomo: Please fix that, the pattern must be set in a field named "format".
Pattern set in label like in DSL syntax is not supported.

configuration/yaml.md
```diff
@@ -0,0 +1,338 @@
...
+|:--------------------------|:-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `<channel_id>`            | The channel ID. It is a single word that when combined with the Thing UID forms the full channel UID.                                                                                |
+| `type`                    | Channel type as defined by the Binding. When the channel type is defined, `kind`, `itemType`, and `itemDimension` should not be defined because the channel type will override them. |
+| `kind`                    | Whether it's a State channel or a Trigger channel. Default: `state`                                                                                                                  |
+| `itemType`                | The item type that this channel can be linked to. Usually this is defined by the binding when channel `type` is defined.                                                             |
```
lolodomo: I would rather describe it like that: The item type that this channel can be linked to. **It is required when channel `type` is not defined**.

configuration/yaml.md
```diff
@@ -0,0 +1,338 @@
...
+
+Channel Map:
+
+| Channel Configuration Key | Description                                                                                                                                                                          |
+|:--------------------------|:-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `<channel_id>`            | The channel ID. It is a single word that when combined with the Thing UID forms the full channel UID.                                                                                |
+| `type`                    | Channel type as defined by the Binding. When the channel type is defined, `kind`, `itemType`, and `itemDimension` should not be defined because the channel type will override them. |
+| `kind`                    | Whether it's a State channel or a Trigger channel. Default: `state`                                                                                                                  |
```
lolodomo: "state" and "trigger" whatever their case are the unique valid values.

configuration/yaml.md
```diff
@@ -0,0 +1,338 @@
...
+```yaml
+things:
+  <thing_uid>:
+    isBridge: true|false
+    bridge: <bridge_uid>
+    label: <Thing label>
+    location: <Location>
+    config:
+      <config key>: <config value>
+    channels: # To define custom channels if necessary
```
lolodomo: Please enhance the comment: **To set a channel configuration parameter or** to define custom channels if necessary

configuration/yaml.md
lolodomo: I see it is already mentioned above but could be repeated in this table.

configuration/yaml.md
```diff
@@ -0,0 +1,338 @@
...
+        <config key>: <config value>
+    metadata:
+      <metadata_key>:
+        value: <metadata value>
+        config:
+          <config key>: <config value>
+```
+
+| Configuration Key | Description                                                                                                                                                                                                                                                                      |
+|:------------------|:---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
```
lolodomo: Please add a first entry for <ItemName> explaining the constraint on it. As a reminder, here is the considered pattern: [a-zA-Z0-9_][a-zA-Z0-9_-]*

configuration/yaml.md
```diff
@@ -0,0 +1,338 @@
...
+| `type`            | Valid [item type]({{base}}/configuration/items.html#type) (mandatory).                                                                                                                                                                                                           |
```
lolodomo: You can precise that the case is not important

configuration/yaml.md
jimtng: Did you mean the constraint on the item name?

configuration/yaml.md
```diff
@@ -0,0 +1,338 @@
...
+| `dimension`       | For dimensioned `Number` items, the `item type` must be set to just `Number`, and the dimension is set separately using this key. Valid dimensions are listed under the **Type** column in the [List of Units]({{base}}/concepts/units-of-measurement.html#list-of-units) table. |
+| `group`           | This key allows defining the group type, dimension and function. For more details, see [Derive Group State from Member Items]({{base}}/configuration/items.html#derive-group-state-from-member-items).                                                                           |
+| `label`           | Item label. Unlike the DSL syntax in an `.items` file, the label here must only contain the label without any formatting syntax.                                                                                                                                                 |
+| `icon`            | The item's icon.                                                                                                                                                                                                                                                                 |
```
lolodomo: You could add a link to where is defined a valid icon syntax.

configuration/yaml.md
```diff
@@ -0,0 +1,338 @@
...
+      parameters: <List of group function parameters> # Example: [ON, OFF]
+    label: labeltext
+    icon: iconname
+    groups:
+      - group1
+      - group2
+    tags:
+      - tag1
+      - tag2
+    format: <format pattern> # short form for state description's pattern
```
lolodomo: add "metadata" at the end

configuration/yaml.md
```diff
@@ -0,0 +1,338 @@
...
+    unit: <unit> # short form for the unit metadata
+    autoupdate: true|false # short form for the autoupdate metadata
+    channel: <channel_uid> # short form to link to a single channel without a profile
```
lolodomo: without any configuration parameter and without a profile

configuration/yaml.md
```diff
@@ -0,0 +1,338 @@
...
+    channels: # long form to link to multiple channels or to assign a profile to a link
```
lolodomo: or to setup channel link configuration

configuration/yaml.md
jimtng: I'll just remove "without a profile" here otherwise it gets too long. It's implied that all they can enter is the channel_uid, and the next line provides more explanation.

configuration/yaml.md
jimtng: I'll reword it differently but still mention config + profile

configuration/yaml.md
```diff
@@ -0,0 +1,338 @@
...
+    tags:
+      - Lightbulb
+
+  BedRoom1_Light_Switch:
+    type: Switch
+    label: Bed Room1 Light Switch
+    icon: switch
+    format: '%s'
+    groups:
+      - BedRoom1_Light_Bulb
```
lolodomo: For the full consistency of your example, note that BedRoom1_Light_Bulb is not defined in that example.

configuration/yaml.md
```diff
@@ -0,0 +1,338 @@
...
+      - Switch
+    channel: mqtt:topic:bedroom1-switch:switch1
+
+  BedRoom1_Switch_StatusLED:
+    type: Switch
+    groups:
+      - BedRoom1_Switch_Group
+    channels:
+      mqtt:topic:bedroom1-switch:statusled:
+        # The configuration of this item-channel link is defined below
```
lolodomo, line 298: It could have been interesting to provide a more common example than profile, the case when the channel has a configuration parameter.

configuration/yaml.md
```diff
@@ -0,0 +1,338 @@
...
+
+  Location_Outdoor_VegetableGarden:
+    label: Vegetable Garden
+
+  Property_Height: {} # no label/description/synonyms
+```
+
+The tag's full UID defines the location of the semantic tag in the tree.
+In the example, `Location_Indoor_Room_HomeCinemaRoom` defines semantic tag `HomeCinemaRoom` as a child of `Room` which is a child of `Indoor` which is a child of `Location`.
+There are constraints on the name of a semantic tag (like HomeCinemaRoom): it must start with a capital letter and contain only letters and numbers.
```
lolodomo: Note that I don't see a check in the tag provider code. Something to be enhanced.

configuration/yaml.md
jimtng: indeed it's a bit all over the place. I'll clean it up

configuration/yaml.md
lolodomo: Yes please, like you did for thing UID

configuration/yaml.md
jimtng: Are you able to provide a better / more useful set of examples, or point me somewhere? I'm generally drawing from my own installation and therefore may be limited.

lolodomo:
Review finished

configuration/yaml.md
jimtng: btw, item names must not contain dashes

jimtng: nor start with a number

lolodomo: Where did you find that ? I believe I applied the same constraints as the ones we used for DSL.

configuration/yaml.md
jimtng: What other kinds of item-channel-link configuration are there other than a profile?

configuration/yaml.md
jimtng: There are many places scattered around core and mainui

https://github.com/openhab/openhab-core/blob/ff1156566baa86074068832e77a22375b6fbcc22/bundles/org.openhab.core/src/main/java/org/openhab/core/items/ItemUtil.java#L56-L58

https://github.com/search?q=repo%3Aopenhab%2Fopenhab-core%20isValidItemName&type=code

configuration/yaml.md
lolodomo: Any channel that defines a configuration parameter.
I have in mind as an example one of both channels of the NTP binding where you can adjust through a parameter how will be formatted the day/hour.

https://github.com/openhab/openhab-addons/blob/main/bundles/org.openhab.binding.ntp/src/main/resources/OH-INF/thing/thing-types.xml#L68

jimtng: > https://github.com/openhab/openhab-addons/blob/main/bundles/org.openhab.binding.ntp/src/main/resources/OH-INF/thing/thing-types.xml#L68

Isn't that for the `Thing` channel configuration, whereas this is the item-channel-link which is part of the Item definition?

lolodomo: Oups, I think you are right, sorry!
I have a doubt now.

configuration/yaml.md
lolodomo: Here is what is supported by DSL syntax:
https://github.com/openhab/openhab-core/blob/main/bundles/org.openhab.core.model.item/src/org/openhab/core/model/Items.xtext#L71
Not checked if there is an additional test later when handling the syntax

jimtng: > Here is what is supported by DSL syntax:
> https://github.com/openhab/openhab-core/blob/main/bundles/org.openhab.core.model.item/src/org/openhab/core/model/Items.xtext#L71

That agrees with the checks in other places, so yes. I'll mention the regex in the doc. I'll push an update shortly.

configuration/yaml.md
jimtng: AFAIK, item-channel-link config is currently only used by profiles, but it should be possible to use it for other things, such as the tagging we're discussing in the other issue, but for now only profiles use it.

jimtng:
Changes pushed. I'll mark the relevant comments as resolved to keep track of things. Feel free to add new code comments if needed.

configuration/yaml.md
jimtng: fixed.

configuration/yaml.md
jimtng: extra sentence added

configuration/yaml.md
jimtng: I've changed the to `state` and `trigger` for better clarity.

configuration/yaml.md
jimtng: added "channel config" in the comment

configuration/yaml.md
jimtng: I've added this suggested line + description of the syntax.

configuration/yaml.md
jimtng: added `case-insensitive`

configuration/yaml.md
jimtng: links added

configuration/yaml.md
jimtng: added

configuration/yaml.md
jimtng: I've changed the example a bit.

configuration/yaml.md
jimtng: I'm leaving the config as is for now, but feel free to comment again if you have further suggestions

configuration/yaml.md
jimtng: I assume this requires no changes in this PR, so marking this as resolved.

configuration/yaml.md
jimtng: Actually, the DSL syntax should not allow `-`. The DSL checker allowed it, but AbstractRegistry doesn't

```
21:42:44.984 [WARN ] [core.common.registry.AbstractRegistry] - Cannot add "NumberItem" with key "Teest-1": The specified name of the item 'Teest-1' is not valid!
```

DSL syntax properly caught `^` as invalid before passing it over to registry:
```
21:42:56.199 [INFO ] [del.core.internal.ModelRepositoryImpl] - Loading model 'test.items'
21:42:56.203 [WARN ] [del.core.internal.ModelRepositoryImpl] - Configuration model 'test.items' has errors, therefore ignoring it: [1,13]: extraneous input '^' expecting EOF
```

jimtng: Oh... it also allows digits as the first char!

configuration/yaml.md
rkoshak: Sorry, I missed you were waiting for me on this one.

I think it's more clear now. I think adding an example for when no configuration is required would help clarify that last sentence though. It took me a couple of reads to fully understand. But it's not a major deal either way.

lolodomo:
It is fine for me.

@jimtng : ideally, you could also include a table for tags listing each field, as you did for things and items.

jimtng:
> ideally, you could also include a table for tags listing each field, as you did for things and items.

Working on it

configuration/yaml.md
jimtng: I'll add an example with long form `channels` with no config

jimtng: @rkoshak here are the changes related to your request:
https://github.com/openhab/openhab-docs/pull/2508/commits/7b8bcdccd64131f961d0c9659d84e77a6b55c153

jimtng:
rebased to main. 

@lolodomo, I've also added a table for tags and refined the text.

configuration/yaml.md
```diff
@@ -0,0 +1,385 @@
...
+[[toc]]
+
+## General Structure
+
+Top-level entries in the YAML file must be unique key-value maps with the following valid keys in no particular order:
+
+| Key                                  | Object Type                                                                                                        |
+|:-------------------------------------|:-------------------------------------------------------------------------------------------------------------------|
+| `version`                            | A mandatory key that contains the file-version. The currently supported version is `1`.                            |
+| `readOnly`                           | An optional key which when set to `true` (the default) would declare that the file may not be modified by openHAB. |
```
spacemanspiff2007: Could you add an example when openHAB would modify a file? Is it only during migrations or does it happen regularly?

configuration/yaml.md
```diff
@@ -0,0 +1,385 @@
...
+      - Switch
+    channel: mqtt:topic:bedroom1-switch:switch1 # short form
+
+  BedRoom1_Light_Button:
+    type: String
+    label: Bedroom1 Light Button Action
+    groups:
+      - BedRoom1_WallSwitch
+    channels: # long form
+      mqtt:topic:bedroom1-switch:button1: {} # No config
```
spacemanspiff2007, line 290: Does it have to be ``{}`` or can it also be empty / null?

jimtng: yes it has to be `{}`. It cannot be empty, that would be a syntax error.

configuration/yaml.md
jimtng: This part is still not really implemented, so there's really nothing to explain at the moment. There's code for core to modify the yaml files, but at the moment this code just sits there, nothing calls it, afaik.

spacemanspiff2007: Could we then add a little snippet "Currently this has no effect"?

configuration/yaml.md
spacemanspiff2007: This is a valid yaml
````yaml
a:
b: asfd
````
and the short form for
````yaml
a: null
b: asfd
````
Syntactically and logically it would be correct.
But does the channel logic work with null values or does it have to be a mapping?

File: configuration/yaml.md (hunk 28 above)
Comment (jimtng) [reply configuration/yaml.md]: I'd be more inclined to remove it (not document it) instead - @lolodomo wdyt?

File: configuration/yaml.md (hunk 29 above)
Comment (jimtng) [reply configuration/yaml.md:290]:
```
config:
```
The above would cause parsing error, because it seems to be interpreted as "" by yaml, and when read by the core code which expects an object, it would complain:

```
Cannot coerce empty String ("") to element of `java.util.LinkedHashMap` (but could if coercion was enabled using `CoercionConfig`)
```


```
config: null 
```

The above seems to work, i.e. no errors, but I have not really looked into what happens.

Comment (jimtng) [reply configuration/yaml.md:291]: I'm here just to document things. 

The following is not supported, so just a hypothetical discussion, and this is really not the place to do that, anyhow...

We could possibly allow this to specify multiple channels without any config:
```
channels:
  - mqtt:topic:topic1
  - mqtt:topic:topic2
```
and if one needs a config, all must use the map syntax:
```
channels:
  mqtt:topic:topic1:
    profile: xxxx
  mqtt:topic:topic2: {} # no config for this one
```
But once again that's a discussion for a different place, and this has probably already been discussed.

There's also a third possibility
```
channels:
  - mqtt:topic:topic1:
       profile: xxx
  - mqtt:topic:topic2
```

We could support all of these, or just keep it simple and stick to how we do it right now.

spacemanspiff2007: > We could possibly allow this to specify multiple channels without any config:
> 
> ```
> channels:
>   - mqtt:topic:topic1
>   - mqtt:topic:topic2
> ```

I'm strongly against introducing a list here - we should stick to the map syntax.
We settled on it for good reasons.

So the correct way should be
````yaml
channels:
  mqtt:topic:topic1:
  mqtt:topic:topic2:
````
or with explicit empty mapping
````yaml
channels:
  mqtt:topic:topic1: {}
  mqtt:topic:topic2: {}
````
Both should result in the same object and work interchangeably.
I'll try to see if it works as expected.

However - this is not a documentation issue so if there is an issue this should be discussed somewhere else.

I was just wondering if an empty entry is allowed and if it should be added to the docs additionally to the empty map.

jimtng: I think we should leave the doc as is in regards to this, since empty causes an error.

jimtng: but, it was a good point to bring up!

configuration/yaml.md
lolodomo: Yes, I agree that we could currently avoid to document it.

lolodomo:
> @lolodomo, I've also added a table for tags and refined the text.

Thank you.

Minor comment: you named the column "Configuration Key" while it is just "Key" in all other tables.

configuration/yaml.md
jimtng: Removed.

jimtng:
> Minor comment: you named the column "Configuration Key" while it is just "Key" in all other tables.

Good catch! That was a recent change. Fixed now.

lolodomo:
I believe we are ready for a merge ;)

lolodomo:
> LGTM so far. Ping me when this needs to be merged.

@stefan-hoehn : you can merge.

PR #2552, opened 2025-08-25
Improve modelling rules for light channels

andrewfg:
See discussion in https://github.com/openhab/openhab-webui/issues/3325
See also https://github.com/openhab/openhab-webui/pull/3344

### DIFF
--- a/developers/addons/faq.md
+++ b/developers/addons/faq.md
@@ -102,7 +102,7 @@ This channel is orthogonal to the brightness and on/off state, but not orthogona
 So changing the color temperature may (or may not) have a consequent impact on the color channel.
 
     - _Absolute Color Temperature Channel:_
-If the light also supports absolute color temperature (in Kelvin).
+If the light also supports absolute color temperature (in Kelvin or Mired/Mirek).
 The case is similar to color temperature in percent above.
 You can implement this by exposing another separate `Dimmer` channel.
 This channel should have the channel-type `system.color-temperature-abs` thus inheriting the `Advanced` attribute.


andrewfg:
@mherwege for info..

andrewfg:
This seems to have failed the CI check due to reasons not related to the new/modified markdown content.

andrewfg:
@jlaur / @lolodomo you probably remember when I wrote the Hue API v2 binding we had some (heated) discussion about the OH "rules" governing what channels a light thing should and should not implement. Well in a recent discussion with @mherwege concerning the Web UI Location card I needed to refer to those rules. But it seems that they were never properly documented as such. There is a FAQ which partly and not very explicitly describes some suggestions. So I have reformulated this FAQ to make it more like rules and more explicit concerning implementation. Therefore please feel free to 1) comment on this PR, and 2) refer to it when you next review a light binding.

mherwege:
LGTM

andrewfg:
@florian-h05 the CI build failed due to an npm error which I think is not related to my markdown text, but could you please check and let me know if I need to change something?

florian-h05:
I have retriggered CI, seems there was a temporary issue with the npm registry.
The build is now properly working and reporting markdownlint errors.

andrewfg:
@florian-h05 many thanks. I have now fixed the lint errors.

developers/addons/faq.md
```diff
@@ -51,34 +51,68 @@ If you do not find an answer to your question, do not hesitate to ask it on the
...
 
-1. _How do I model a dimmable light?_
-
-    A dimmable light should only have a single channel of type `system.brightness` for its control.
-    Note that although this channel type is declared for item type 'Dimmer', it is perfectly possible to link a `Switch`
-    item to it, additionally to a `Dimmer` item.
-    The possibility to switch the light ON/OFF is therefore implicitly available and there is no need for an additional channel with item type `Switch`, even if the hardware might have separated switching and dimming in different commands.
-    In the openHAB abstraction layer, 0% brightness is identical to OFF.
-    Sending an ON command to such a channel does not have to switch to 100% brightness, though.
+1. _How do I model a light with color, brightness, on/off, and/or color temperature chanels?_
```
lsiepel:
```suggestion
1. _How do I model a light with color, brightness, on/off, and/or color temperature channels?_
```

developers/addons/faq.md
```diff
@@ -51,34 +51,68 @@ If you do not find an answer to your question, do not hesitate to ask it on the
...
+
+    Lights have different capabilities -- some are just on/off, some are dimmable, some have cold/warm color temperature control, and some have full color control.
+So when you model a light in your binding, you must follow the _**"Highest Capability Channel Rule"**_.
+This rule means supporting **only one** of the following _Single Color Channel for Top Capability Light, Single Dimmer Channel for Mid Capability Light_, or _Single Switch Channel for Low Capability Light_ scenarios:
+
+    - _Single Color Channel for Top Capability Light:_
+    If the light supports full color, then it must expose one single `Color` channel only.
```
lsiepel:
```suggestion
    If the light supports full color, it must expose one single `Color` channel only.
```

developers/addons/faq.md
```diff
@@ -51,34 +51,68 @@ If you do not find an answer to your question, do not hesitate to ask it on the
...
+    This should have the channel-type `system.color` therefore inheriting default tags `[Control, Color]`.
+    Although the channel is of type `Color`, items of type `Dimmer` or `Switch` can be linked to it too.
```
lsiepel:
```suggestion
    Although the channel is of type `Color`, items of type `Dimmer` or `Switch` can be linked to it.
```

developers/addons/faq.md
```diff
@@ -51,34 +51,68 @@ If you do not find an answer to your question, do not hesitate to ask it on the
...
+    The possibility to dim or switch the light is therefore implicitly available so there is no need for an additional channel with item type `Dimmer` or `Switch`, even if the hardware might have separated switching and dimming in different commands.
+    Therefore the channel must be able to process all of the following types of command:
+      - HSBType commands control the color hue, the brightness, and the on/off state, _plus_
+      - PercentType commands control  the brightness, and the on/off state, _plus_
+      - OnOffType commands control the on/off state.
```
lsiepel:
```suggestion
      - HSBType commands control the color hue, brightness, and on/off state, _plus_
      - PercentType commands control the brightness, and on/off state, _plus_
      - OnOffType commands control the on/off state
```

developers/addons/faq.md
```diff
@@ -51,34 +51,68 @@ If you do not find an answer to your question, do not hesitate to ask it on the
...
+
+    Lights have different capabilities -- some are just on/off, some are dimmable, some have cold/warm color temperature control, and some have full color control.
+So when you model a light in your binding, you must follow the _**"Highest Capability Channel Rule"**_.
+This rule means supporting **only one** of the following _Single Color Channel for Top Capability Light, Single Dimmer Channel for Mid Capability Light_, or _Single Switch Channel for Low Capability Light_ scenarios:
+
+    - _Single Color Channel for Top Capability Light:_
+    If the light supports full color, then it must expose one single `Color` channel only.
+    This should have the channel-type `system.color` therefore inheriting default tags `[Control, Color]`.
+    Although the channel is of type `Color`, items of type `Dimmer` or `Switch` can be linked to it too.
+    The possibility to dim or switch the light is therefore implicitly available so there is no need for an additional channel with item type `Dimmer` or `Switch`, even if the hardware might have separated switching and dimming in different commands.
```
lsiepel:
```suggestion
    The ability to dim or switch the light is therefore implicitly available so there is no need for an additional channel with item type `Dimmer` or `Switch`, even if the hardware might have separated switching and dimming in different commands.
```

developers/addons/faq.md
```diff
@@ -51,34 +51,68 @@ If you do not find an answer to your question, do not hesitate to ask it on the
...
+    Therefore the channel must be able to process all of the following types of command:
+      - HSBType commands control the color hue, the brightness, and the on/off state, _plus_
+      - PercentType commands control  the brightness, and the on/off state, _plus_
+      - OnOffType commands control the on/off state.
+
+      When the brightness is changed through a linked `Dimmer` item, the hue and saturation values must be kept.
+    Likewise, when an `OFF` command is sent through a linked `Switch` item, it must adjust the brightness to 0%, and/or set the on/off state to OFF, so that with the next ON (or brightness change), the hue and saturation values are restored untouched.
+    In the openHAB abstraction layer, 0% brightness is identical to `OFF`.
+    Sending an `ON` command to such a channel does not have to switch to 100% brightness, though.
```
lsiepel:
```suggestion
    Sending an `ON` command to such a channel does _not_ have to switch to 100% brightness; depending on the device, it can also dim to the last brightness value (other than 0%) it had before.
```

developers/addons/faq.md
```diff
@@ -51,34 +51,68 @@ If you do not find an answer to your question, do not hesitate to ask it on the
...
+    If the light supports full color, then it must expose one single `Color` channel only.
+    This should have the channel-type `system.color` therefore inheriting default tags `[Control, Color]`.
+    Although the channel is of type `Color`, items of type `Dimmer` or `Switch` can be linked to it too.
+    The possibility to dim or switch the light is therefore implicitly available so there is no need for an additional channel with item type `Dimmer` or `Switch`, even if the hardware might have separated switching and dimming in different commands.
+    Therefore the channel must be able to process all of the following types of command:
+      - HSBType commands control the color hue, the brightness, and the on/off state, _plus_
+      - PercentType commands control  the brightness, and the on/off state, _plus_
+      - OnOffType commands control the on/off state.
+
+      When the brightness is changed through a linked `Dimmer` item, the hue and saturation values must be kept.
```
lsiepel:
```suggestion
      When the brightness is changed through a linked `Dimmer` item, the hue and saturation values must be preserved.
```

developers/addons/faq.md
```diff
@@ -51,34 +51,68 @@ If you do not find an answer to your question, do not hesitate to ask it on the
...
+    Likewise, when an `OFF` command is sent through a linked `Switch` item, it must adjust the brightness to 0%, and/or set the on/off state to OFF, so that with the next ON (or brightness change), the hue and saturation values are restored untouched.
```
lsiepel:
```suggestion
    Likewise, when an `OFF` command is sent through a linked `Switch` item, it must adjust the brightness to 0% and/or set the on/off state to OFF, so that with the next ON (or brightness change), the hue and saturation values are restored untouched.
```

andrewfg:
> Would it be possible to provide some core utility classes to make this easy for developers to adapt and/or contribute?
> It might be useful to mention or provide a link to the helpers.

What do you have in mind? Maybe conversion between color HSBType B part <=> Percentype brightness <=> OnOffType brightness != 0 ?? Or do you have something else in mind? Alternatively maybe a link to [sample implementation](https://github.com/openhab/openhab-addons/blob/6ab5d848e87a5feea940c6ff7582885bcef62540/bundles/org.openhab.binding.hue/src/main/java/org/openhab/binding/hue/internal/handler/Clip2ThingHandler.java#L384) ??

> Maybe also add a note that brightness channels like a display should not be tagged as they are not considered a light ?

No. The tag combination `Control+Brightness` is 100% correct for a point that controls the brightness of a display. There is admittedly a question about whether the UI Location Card should interpret `Control+Brightness` as being (only) a light. However that is a separate discussion relating to the UI logic, and NOT the tagging logic.

lsiepel:
> > Would it be possible to provide some core utility classes to make this easy for developers to adapt and/or contribute?
> > It might be useful to mention or provide a link to the helpers.
> 
> What do you have in mind? Maybe conversion between color HSBType B part <=> Percentype brightness <=> OnOffType brightness != 0 ?? Or do you have something else in mind? Alternatively maybe a link to [sample implementation](https://github.com/openhab/openhab-addons/blob/6ab5d848e87a5feea940c6ff7582885bcef62540/bundles/org.openhab.binding.hue/src/main/java/org/openhab/binding/hue/internal/handler/Clip2ThingHandler.java#L384) ??

Yes, an implementation example is really helpfull. I'm not that familiair to core helpers regarding color, but i guess developers face the same problems. Conversion helpers between individual channels and HSBType, with or without warm/cold white. I have to look at what you did for the zwavejs binding, it did envolve some logic that many developers are not immediate familiar with.

mherwege:
> No. The tag combination `Control+Brightness` is 100% correct for a point that controls the brightness of a display. There is admittedly a question about whether the UI Location Card should interpret `Control+Brightness` as being (only) a light. However that is a separate discussion relating to the UI logic, and NOT the tagging logic.

I understand that, but without that assumption, we can never restore the UI functionality as it was before. We do not have a way to distinguish a light from something else if we do not force it to be part of a LightSource Equipment. And we did not have that requirements before to simplify modelling. After all, forcing creation of an extra group item containing a single item looks ridiculous. And that's what was avoided before.
We have a similar challenge with the exception we made to allow points on a location, e.g. for a light connected to a hub equipment in one location, but physically controlling the light in another location. Also there, we count these points as lights.

One way around this could be to allow the LightSource Equipment tag to be added directly to the Point item if the Equipment is represented by a single point. But that is a much bigger change, not only in core, but in addons and very much so the UI.
Another way is to introduced extra properties distinguishing between properties for lights and for other uses.

andrewfg:
@mherwege I suppose we could say that `Control+Brightness` shall be used uniquely for real lights, and (say) `Control+Illuminance` or `Control+Level` shall be (mis-)used for displays .. but it seems rather clunky.

mherwege:
> @mherwege I suppose we could say that `Control+Brightness` shall be used uniquely for real lights, and (say) `Control+Illuminance` or `Control+Level` shall be (mis-)used for displays .. but it seems rather clunky.

Yes, agree it feels clunky. But I don't see another easy solution. Allowing a combination of Equipment and Point tags on the same item feels clunky as well. It could solve the modelling issue, but is not easy to generate automatically from channel and thing definitions and has a whole lot of consequences throughout the code.

lsiepel:
Let's invite @openhab/add-ons-maintainers to share their thoughts.

andrewfg:
Apropos new [helper methods in core](https://github.com/openhab/openhab-core/pull/4995) .. I would add a link to this class and the `ColorUtil` class once that PR will have been reviewed and merged.

andrewfg:
I have been thinking about this some more. And I think we are over thinking the issue.

Let us go back to basics: 

1. If an item implements Switch+Light or Control+Brightness or Control+Color then it DOES represent a light. Period.
2. And if it does represent a light it shall be included in the location card. Period.
3. And if the user does not want this light to appear in the card (because he considers it not to be a light but rather a display or whatever) then he can simply manually de-tag or re-tag that item. This is up to the user. Period.
4. As a general rule bindings shall not offer more than one channel with Switch+Light or Control+Brightness or Control+Color default tags unless each such channel does represent a different light.

lsiepel:
What are your thoughts on the display (control+brightness) of example? How would that fit in those statements (1).

I know it is easy to loose yourself with exemptions.

andrewfg:
> What are your thoughts on the display (control+brightness) of example? How would that fit in those statements (1)

It is a light.

mherwege:
> It is a light.

Can it be switched on and off as well? Is the binding code allowing the channel to be connected to a switch item? I would not consider this as a light, it is the dim level of a display.

andrewfg:
> Can it be switched on and off as well?

No.

@mherwege my suggestion was to not lose sleep over such things; so what is your counter suggestion?

mherwege:
@andrewfg My preference would be to remove the semantic tags on such a channel. Otherwise, the light badges will always show at least on light on in a room where this device is, and it will cost most people a lot of time to figure out where it comes from. Rather have them add it manually. I would reserve the light, brightness and color tags for use on real lights. This is just my opinion.

developers/addons/faq.md
```diff
@@ -51,34 +51,69 @@ If you do not find an answer to your question, do not hesitate to ask it on the
...
+Color temperature is nothing else than a color palette with a limited (linear) choice of colors.
+However there is no dedicated item type for color temperature and it is tricky (though not impossible) to map color temperatures to HSB values.
+So you cannot "overload" the `Color` channel in the same way as is done for brightness above.
+Instead you must expose a separate `Dimmer` channel.
+This should have the channel-type `system.color-temperature` thus inheriting default tags `[Control, ColorTemperature]`.
+This channel is orthogonal to the brightness and on/off state, but not orthogonal to the HS part of the HSB.
+So changing the color temperature may (or may not) have a consequent impact on the color channel.
+
+    - _Absolute Color Temperature Channel:_
+If the light also supports absolute color temperature (in Kelvin).
```
ccutrer:
```suggestion
If the light also supports absolute color temperature (in Kelvin or mireks).
```

stefan-hoehn:
All: let me know when you are done - I trust your reviews - I will then merge

andrewfg:
@stefan-hoehn I think this is ready.

PR #2563, opened 2025-09-25
Updated persistence settings screenshot to show RR4Dj selection

james-h-dt:
Signed-off-by: james-h-dt <67227920+james-h-dt@users.noreply.github.com>


stefan-hoehn:
please explain why you want to update this. The settings dialog is different depending on the persistence services you installed. Mine would also look different...

james-h-dt:
The updated image reflects the current default persistence settings for a
vanilla installation. It also shows the same selection as the updated text.

rkoshak:
The default version of that dialog would show no persistence engines installed at all. The first run wizard strongly suggests installing these, but that's not the default. It's a choice made by the user.

But the screen should match the text. See my comments on that PR.

This probably should have been included as part of that PR since they go together.

PR #2564, opened 2025-09-27
Remove configuration/rules-ng.md

dilyanpalauzov:
It is not readable since commit 8d8ae4d8e0a2f1fefbfa and year 2020.


stefan-hoehn:
@openhab/maintainers Jerome has removed this page "temporarily"  in #1347 "until we have a "new UI" replacement finished.". I wonder if this isn't the case by now. I wouldn't want to lightheartedly delete a page, hence can someone double check if we can bring it back instead and what is missing to do so?

florian-h05:
I chime in here on Stefan's opinion, quickly scanning over that page, I think this might be something to go into the developer documentation on how to create modules for the rule engine.

stefan-hoehn:
We jointly double checked the page and agree that this can be removed.

stefan-hoehn (approved):

PR #2567, opened 2025-10-02
JSR223: Add lifecycle, media, provider presets & scriptLoaded, scriptUnloaded functions

dilyanpalauzov:
I add example only for Groovy, as I do know how to use it in the other automation add-ons.  Besides, the other rule engines provide other ways for the same functionality: `script_unloaded` in JRuby, and for others I do not know if the functionality is from openhab-core, or from the add-on.

### DIFF
--- a/configuration/jsr223.md
+++ b/configuration/jsr223.md
@@ -706,7 +706,7 @@ Read the JSR223 language specific documentation for examples of using these `Tri
 
 :::
 
-### The `scriptLoaded` and `scriptUnloaded` functions
+### `scriptLoaded` and `scriptUnloaded` functions
 
 When a script file is loaded, the `scriptLoaded(String)` function is invoked if it is defined.
 The path of the script file will be passed as parameter.


dilyanpalauzov:
Now includes Nashorn example.

dilyanpalauzov:
The current proposed change is correct, but only part of the truth.  My understanding is that if a class `implements ScriptExtensionProvider` and returns in `getDefaultPresets()` non-empty set, then it is loaded by default by org.openhab.core.automation.module.script.internal.ScriptExtensionManager:
```java
    public List<String> getDefaultPresets() {
        List<String> defaultPresets = new ArrayList<>();
        for (ScriptExtensionProvider provider : scriptExtensionProviders) {
            defaultPresets.addAll(provider.getDefaultPresets());
        }
        return defaultPresets;
    }
```

That is, the `MediaScriptScopeProvider`, `LifecycleScriptExtensionProvider` and `DefaultScriptScopeProvider` script extensions, or in other words the `default`, `media` and `lifecycle` presets, do not require `importPresent`.

dilyanpalauzov:
Besides, one of the presets, which is loaded without `importPreset` is called `default`, not “Defaut”: In https://github.com/openhab/openhab-core/blob/main/bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/defaultscope/DefaultScriptScopeProvider.java: `private static final String PRESET_DEFAULT = "default";`

dilyanpalauzov:
I added also the `media` preset - variables `org.openhab.core.audio.AudioManager audio` and `org.openhab.core.voice.VoiceManager video`.

configuration/jsr223.md
```diff
@@ -335,6 +337,47 @@ With `scriptExtension.get("automationManager")` the `automationManager` can be m
 - `events.storeStates(Item...)`
 - `events.restoreStates(Map<Item, State>)`
 
+#### `lifecycle` Preset (`importPreset` not required)
+
+It provides a mechanism to execute code, when the input file is deleted.  Modifying a file deletes it and creates it again.
```
stefan-hoehn: 2nd sentence has be in the next line. (make sure the first sentence doesn't have a trailing space then.

stefan-hoehn (changes requested):
@florian-h05 please check the content.

dilyanpalauzov:
I have described also the `scriptLoaded(String)` and `scriptUnloaded()` functions, and included examples in Groovy, JavaScript and Python.

florian-h05 (approved):
Thanks!
I've reworded a few things during review.

configuration/jsr223.md
```diff
@@ -568,10 +566,10 @@ The preset is available since openHAB 5.0.
...
 
-All instances from this preset have an `addPermanent(…)` method, in addition to `add(…)`.
-With the `add(…)` method, elements added over the above instances are removed, when the script, which inserted the elements, is deleted or changed.
-When `addPermanent(…)` is used, the inserted elements are … not removed.
-`itemRegistry.addPermanent(…)`/`thingRegistry.addPermanent(…)` from this preset offer the same functionality as `itemRegistry.add(…)`/`things.add(…)` from the `default` preset.
+All instances from this preset have an `addPermanent(...)` method, in addition to `add(...)`.
+With the `add(...)` method, elements added over the above instances are removed, when the script, which inserted the elements, is deleted or changed.
+When `addPermanent(...)` is used, the inserted elements `...` are not removed.
+`itemRegistry.addPermanent(...)`/`metadataRegistry.addPermanent(...)`/`thingRegistry.addPermanent(...)` from this preset offer the same functionality as `itemRegistry.add(...)`/`metadataRegistry.add(...)`/`things.add(...)` from the `default` preset.
```
dilyanpalauzov, line 572: The `default` preset is implemented in https://github.com/openhab/openhab-core/blob/main/bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/defaultscope/DefaultScriptScopeProvider.java.  It does not have `metadataRegistry`.

configuration/jsr223.md
```diff
@@ -710,12 +708,13 @@ Read the JSR223 language specific documentation for examples of using these `Tri
...
-Similarly, when a script is changed or deleted, and the script contained the `scriptUnloaded()` function, it is invoked.
-`scriptLoaded(String)` is only executed when a script from the automation add-on directory, like `${OPENHAB_CONF}/automation/jsr223/`, is loaded.
-It is not executed in transformations, UI rules with Inline Script action, UI scripts.
-The [`lifecycleTracker`](#lifecycle-preset-importpreset-not-required) object and the `scriptUnloaded()` function serve similar purposes.
-When a script is executed several times, on each execution it adds a Disposable, and then the script is changed, its `scriptUnloaded()` is executed once and each added Disposable is also `dispose()`d.
+When a script file is loaded, the `scriptLoaded(String)` function is invoked if it is defined.
+The path of the script file will be passed as parameter.
+Similarly, when a script file is unloaded, the `scriptUnloaded()` function is invoked if it is defined.
+Note that `scriptLoaded(String)` and `scriptUnloaded()` are only executed for script files - not Script Actions and Script conditions of rules.
```
dilyanpalauzov, line 714: The original text suggests that for transformations (file in etc/openhab/transform) `scriptUnloaded()` is executed, but `scriptLoaded(String)` is not executed.  After the changes this information is lost.

configuration/jsr223.md
```diff
@@ -710,12 +708,13 @@ Read the JSR223 language specific documentation for examples of using these `Tri
...
+
+[`lifecycleTracker`](#lifecycle-preset-importpreset-not-required)'s `addDisposeHook(Disposable)` and the `scriptUnloaded()` function serve similar purposes, but it is recommended to use `addDisposeHook(Disposable)` as it's functionality is not limited to script files only.
```
dilyanpalauzov, line 716: ”it`s"  → “its”.

Why was this removed:

> When a script is executed several times, on each execution it adds a Disposable, and then the script is changed, its `scriptUnloaded()` is executed once and each added Disposable is also `dispose()`d.

This is not what I meant:
> The `lifecyleTracker` also supports adding multiple dispose hooks, contrary to `scriptUnloaded()` that can only be defined once.

I meant that defining once `scriptUnloaded()` and one Disposable in a script, then executing the script several times, then modifying the script, will execute scriptUnloaded once, and the Disposable many times. (as many times as the script was executed, provided that each execution created unconditinonally a new Disposable - this can be tested in transformations)

configuration/jsr223.md
```diff
@@ -568,10 +566,10 @@ The preset is available since openHAB 5.0.
...
 The `itemChannelLinkRegistry` does not provide the same methods as the [`org.openhab.core.thing.link.ItemChannelLinkRegistry`](https://www.openhab.org/javadoc/latest/org/openhab/core/thing/link/itemchannellinkregistry) interface.
 
-All instances from this preset have an `addPermanent(…)` method, in addition to `add(…)`.
-With the `add(…)` method, elements added over the above instances are removed, when the script, which inserted the elements, is deleted or changed.
-When `addPermanent(…)` is used, the inserted elements are … not removed.
-`itemRegistry.addPermanent(…)`/`thingRegistry.addPermanent(…)` from this preset offer the same functionality as `itemRegistry.add(…)`/`things.add(…)` from the `default` preset.
+All instances from this preset have an `addPermanent(...)` method, in addition to `add(...)`.
+With the `add(...)` method, elements added over the above instances are removed, when the script, which inserted the elements, is deleted or changed.
+When `addPermanent(...)` is used, the inserted elements `...` are not removed.
```
dilyanpalauzov, line 571: The second ellipsis I used on that line was supposed to “mark a pause in speech” - as allowed by https://en.wikipedia.org/wiki/Ellipsis , not to cite the parameters introduced by the first ellipsis.

configuration/jsr223.md
```diff
@@ -799,7 +798,7 @@ logger.error("I am logged first.")
 
 ::: tab Python
 
-When the setting “Use scope and import wrapper” is on, the `scope` module is enabled:
+When the setting "Use scope and import wrapper" is enabled, the `scope` module is enabled:
```
dilyanpalauzov, line 801: This repeats “enabled” twice in the same sentence.

configuration/jsr223.md
florian-h05: How should a script be executed several times? A transformation can be, but a script not. Since `scriptUnloaded` is not available in transformations that information provides no benefit.

dilyanpalauzov: When I wrote that tests I believe I created transformations with  `scriptUnloaded()`, `scriptLoaded(String)` and lifecycleTracker.  `scriptLoaded(String)` was not executed, `scriptUnloaded()` and the Disposable were executed.  I mean transformations with a file under `etc/openhab/transform/`.  So an openHAB-Groovy-filebased-Transformation is a Groovy-Script and it can be executed several times.

PR #2570, opened 2025-10-03
Script transformations: State that execution is synchronised

dilyanpalauzov:
This is stated at https://github.com/dalgwen/openhab-addons/blob/c37d9faa599907b7c865f81eba66998503233849/bundles/org.openhab.automation.java223/README.md?plain=1#L267 and at https://community.openhab.org/t/rules-and-concurrency/166577/14.

### DIFF
--- a/configuration/transformations.md
+++ b/configuration/transformations.md
@@ -66,6 +66,7 @@ To keep these examples simple, the contents of the referenced files `window_esp.
 
 The script transformation is available from the framework and needs no additional installation.
 It allows transforming values using any of the available scripting languages in openHAB (JSR-223 or DSL).
+openHAB ensures that one and the same script transformation is not executed in parallel, there is no need to program protections against race conditions.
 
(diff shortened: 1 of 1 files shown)


configuration/transformations.md
```diff
@@ -66,6 +66,7 @@ To keep these examples simple, the contents of the referenced files `window_esp.
 
 The script transformation is available from the framework and needs no additional installation.
 It allows transforming values using any of the available scripting languages in openHAB (JSR-223 or DSL).
+openHAB ensures that one and the same transformation is not executed in parallel, there is no need to program protections against race conditions.
```
florian-h05:
```suggestion
openHAB ensures that one and the same script transformation is not executed in parallel, there is no need to program protections against race conditions.
```

Let‘s make it super clear that this only applies to SCRIPT transformations.

florian-h05:
No force pushing after a review has started!!

florian-h05 (approved):

PR #2487, opened 2025-04-24
stateDescription: change to read-write

dilyanpalauzov:
This takes the text from https://github.com/openhab/openhab-addons/issues/18362#issuecomment-2797687192 and extends the documentation, telling the user how an item, bound to read-only channel, can be made read-write.

However I have experience only with MQTT and there `CommandTopic` applies.  For other things `CommandTopic` might not be adequate.

PR #2573, opened 2025-10-04
concept/rules: state that a rule cannot be executed in parallel

dilyanpalauzov:
I create a file rules/a.rules:
```
rule "Slow Rule"
when
  Item r changed
then
  logError("A", "SLOW rule start at " + ZonedDateTime.now().toString().substring(11, 25))
  Thread.sleep(10000)
  logError("A", "SLOW rule end at " + ZonedDateTime.now().toString().substring(11, 25))
end

rule "Fast Rule"
when
  Item r changed
then
  logError("A", "FAST rule run at " + ZonedDateTime.now().toString().substring(11, 25))
end
```
and then change `r` six times fast. openhab.log contains:
```
2025-10-04 14:49:03.083 [INFO ] [el.core.internal.ModelRepositoryImpl] - Loading DSL model 'a.rules'
2025-10-04 14:50:04.628 [ERROR] [org.openhab.core.model.script.A     ] - SLOW rule start at 14:50:04.62676
2025-10-04 14:50:04.630 [ERROR] [org.openhab.core.model.script.A     ] - FAST rule run at 14:50:04.62817
2025-10-04 14:50:04.797 [ERROR] [org.openhab.core.model.script.A     ] - FAST rule run at 14:50:04.79549
2025-10-04 14:50:05.518 [ERROR] [org.openhab.core.model.script.A     ] - FAST rule run at 14:50:05.51655
2025-10-04 14:50:06.242 [ERROR] [org.openhab.core.model.script.A     ] - FAST rule run at 14:50:06.24042
2025-10-04 14:50:06.948 [ERROR] [org.openhab.core.model.script.A     ] - FAST rule run at 14:50:06.94709
2025-10-04 14:50:07.683 [ERROR] [org.openhab.core.model.script.A     ] - FAST rule run at 14:50:07.68087
2025-10-04 14:50:14.634 [ERROR] [org.openhab.core.model.script.A     ] - SLOW rule end at 14:50:14.63184
2025-10-04 14:50:14.643 [ERROR] [org.openhab.core.model.script.A     ] - SLOW rule start at 14:50:14.64059
2025-10-04 14:50:24.649 [ERROR] [org.openhab.core.model.script.A     ] - SLOW rule end at 14:50:24.64700
2025-10-04 14:50:24.657 [ERROR] [org.openhab.core.model.script.A     ] - SLOW rule start at 14:50:24.65547
2025-10-04 14:50:34.663 [ERROR] [org.openhab.core.model.script.A     ] - SLOW rule end at 14:50:34.66138
2025-10-04 14:50:34.679 [ERROR] [org.openhab.core.model.script.A     ] - SLOW rule start at 14:50:34.67715
2025-10-04 14:50:44.685 [ERROR] [org.openhab.core.model.script.A     ] - SLOW rule end at 14:50:44.68312
2025-10-04 14:50:44.700 [ERROR] [org.openhab.core.model.script.A     ] - SLOW rule start at 14:50:44.69756
2025-10-04 14:50:54.706 [ERROR] [org.openhab.core.model.script.A     ] - SLOW rule end at 14:50:54.70419
2025-10-04 14:50:54.718 [ERROR] [org.openhab.core.model.script.A     ] - SLOW rule start at 14:50:54.71562
2025-10-04 14:51:04.725 [ERROR] [org.openhab.core.model.script.A     ] - SLOW rule end at 14:51:04.72241
```
So both rules are executed six times, but SLOW Rule waits its own execution to end, before starting from the start.


dilyanpalauzov:
If a rule cannot be executed in parallel, why does https://www.openhab.org/docs/configuration/rules-dsl.html#concurrency-guard say “If a rule triggers on UI events it may be necessary to guard against concurrency.”?

rkoshak:
>"If a rule triggers on UI events it may be necessary to guard against concurrency.”

That sentence is probably poorly worded. 

MainUI widgets can directly call a rule as its action. This usage is similar to running a rule manually (e.g. pressing the play button), running the rule directly through the REST API (`POST /rules/{ruleUID]/runnow`), or running a rule from another rule. In these cases, the locks you exercised above all get bypassed. 

Note, I think that MainUI actually has some logic in place that will prevent you from manually running a rule that's already running on that rule's page. But that's logic built into the web page and it does not extend to UI widgets.

dilyanpalauzov:
I do not understand.  Either OpenHAB-core, irrespective of web pages, ensures that a single Rule (or a script) can have a single running instance, or it does not make such promises.  In this case, OpenHAB just waits for a rule to finish and if the rule is triggered in the meantime again, after it finishes, it is fired again.  This is what the example above shows, this is how it works with transformations.

If OpenHAB lets a rule under circumstances start running, before another instance of the same rule has finished, then this is completely different story and the conditions, when this can happen, should be described, so that locks guard against this.

Do UI-widgets or REST API calls allow a rule (e.g. a rule containing `Thread.sleep()`) to be restarted, while the same rule is executing (was started in the past and has not yet finished)?

rkoshak:
> I do not understand. Either OpenHAB-core, irrespective of web pages, ensures that a single Rule (or a script) can have a single running instance, or it does not make such promises. 

There are many ways to run a rule. Some ways there are locks. Other ways there are no locks. 

Rules that are run based on a trigger are locked. Rules directly run (i.e. not run based on a trigger) there is no lock.

> In this case, OpenHAB just waits for a rule to finish and if the rule is triggered in the meantime again, after it finishes, it is fired again. 

Right, the key word is "triggered" here. MainUI doesn't trigger the rule. It directly invokes it. That comment in the docs you point out is referring to this direct invocation. 

> Do UI-widgets or REST API calls allow a rule (e.g. a rule containing Thread.sleep()) to be restarted, while the same rule is executing (was started in the past and has not yet finished)?

Yes. You will have two of this rule running at the same time. 


Rule Action:
image

Widget's Action:
image

From a Script (JS in this case, note that Rules DSL lacks this ability)

```js
rules.runRule('RuleUID', { }, true);
```

REST API:

image


None of these ways to run a rule have locks. Only when a rule is triggered is it locked.

dilyanpalauzov:
I have changed the wording to state, that rules fired by their triggers cannot be executed in parallel, while rules executed by [`RuleManager.runNow()`](https://www.openhab.org/javadoc/latest/org/openhab/core/automation/rulemanager) - from a script or another rule; or started from a MainUI widget can be executed concurrently.

@Nadahar said at https://github.com/openhab/openhab-addons/pull/19410#issuecomment-3352743357, that each rule has a dedicated thread -  .

Does it mean, that when a rule is executed by [`RuleManager.runNow()`](https://www.openhab.org/javadoc/latest/org/openhab/core/automation/rulemanager) from another rule, MainUI widget, the rule will run in that very same thread, on `runNow()` runs the rule in a different thread?

> Note, I think that MainUI actually has some logic in place that will prevent you from manually running a rule that's already running on that rule's page. But that's logic built into the web page and it does not extend to UI widgets.

Is this logic in MainUI irrelevant (bypassed), if the same page is opened in two browsers?

Nadahar:
> Does it mean, that when a rule is executed by [`RuleManager.runNow()`](https://www.openhab.org/javadoc/latest/org/openhab/core/automation/rulemanager) from another rule, MainUI widget, the rule will run in that very same thread, on `runNow()` runs the rule in a different thread?

Actually, I believe that there's a "problem" when using `runNow()` (and all the derived ways to launch a rule), and also when run from timers. I don't think that the "rule thread" is respected/used in this situation, which leads to problems. The very reason one thread was dedicated to each rule was to avoid all the problems/errors experienced when this wasn't the case. Everything within a rule is usually done without any thread-safety, and it would potentially be very complicated to change that, which is why the "one rule one thread" solution was chosen as I understand it.

dilyanpalauzov:
> Actually, I believe that there's a "problem" when … run from timers.

I checked for timer triggers, in particular for `timer.GenericCronTrigger` with Groovy:
```groovy
logger = org.slf4j.LoggerFactory.getLogger("C")
global_i = 0
automationManager.addRule(
  new org.openhab.core.automation.module.script.rulesupport.shared.simple.SimpleRule() {
    @Override
    Object execute(org.openhab.core.automation.Action module, Map<String, ?> inputs) {
        Integer local_i = global_i++
        logger.error("START " + local_i)
        Thread.sleep(15000)
        logger.error("END " + local_i)
    }
    {
      triggers = [org.openhab.core.automation.util.TriggerBuilder.create().withId("a")
          .withTypeUID("timer.GenericCronTrigger").withConfiguration(
             new org.openhab.core.config.core.Configuration([cronExpression: "0/5 * * * * *"])).build()
      ]
    }
})
```
prints
```
2025-10-11 09:57:31.091 [INFO ] [ort.loader.AbstractScriptFileWatcher] - (Re-)Loading script '/etc/openhab/automation/jsr223/ff.groovy'                                                                                         
2025-10-11 09:57:35.229 [ERROR] [C                                   ] - START 0 
2025-10-11 09:57:50.234 [ERROR] [C                                   ] - END 0 
2025-10-11 09:57:50.237 [ERROR] [C                                   ] - START 1 
2025-10-11 09:58:05.240 [ERROR] [C                                   ] - END 1 
2025-10-11 09:58:05.244 [ERROR] [C                                   ] - START 2 
2025-10-11 09:58:20.246 [ERROR] [C                                   ] - END 2 
2025-10-11 09:58:20.251 [ERROR] [C                                   ] - START 3 
2025-10-11 09:58:35.254 [ERROR] [C                                   ] - END 3 
2025-10-11 09:58:35.257 [ERROR] [C                                   ] - START 4 
2025-10-11 09:58:50.260 [ERROR] [C                                   ] - END 4 
2025-10-11 09:58:50.264 [ERROR] [C                                   ] - START 5 
2025-10-11 09:59:05.266 [ERROR] [C                                   ] - END 5
```
This trigger fires every five seconds, but because of `Thread.sleep(15000)` the system executes the trigger every fifteen seconds.

By their nature the other timer triggers `timer.TimeOfDayTrigger` and `timer.DateTimeTrigger` cannot run in parallel. The first runs once a day, the second also. Unless the trigger configuration for TimeOfDayTrigger is changed in a loop, or the item bound to DateTimeTrigger is changed in a loop, which loop repeat every second and the action handler is slow. I am not going to test if `timer.TimeOfDayTrigger` and `timer.DateTimeTrigger` can be triggered in parallel.

Nadahar:
@dilyanpalauzov I didn't mean timer triggers, but timers used in scripts. I'm not entirely sure how they are used to execute rules, I've just read it referenced.

However, you should keep an eye on this, it might change how this works: https://github.com/openhab/openhab-core/pull/5069

dilyanpalauzov:
# Are profiles also restricted to run sequentially, in the same way as transformations are?

Sequential executions for profiles is stated at https://github.com/dalgwen/openhab-addons/blob/java223/main/bundles/org.openhab.automation.java223/README.md#concurrency with these words:
```
openHAB also prevents several executions of the same rule at the same time. This means, for example, that a script defined as the 'Action' part of a GUI rule cannot run concurrently, even if triggered by another rule at the same moment (The second one will fail).

openHAB prevents several executions of the same script at the same time. This is also the case for transformations/profile (the second transformation will wait for the first to finish). This can be an issue, especially if your script takes some non-negligible time to execute.
```

# Timers in Rules and Scripts

My understanding is that a timer in a script is in essence a new thread, and then locks can be justified, if the timer-body references variables, which exist outside the timer, and and the rule is retriggered - the rule accesses the same variables at the same time - once after being retriggered, and once from the timer.

That said, I think timers and locks in rules and scripts do not need extra text.

I have such a rule (DSL), which changes a value of a String item (e3_colour) to contain a colour (red, orange, yellow).  The longer the device is on, the hotter the colour.
```
var e3Lock = new java.util.concurrent.locks.ReentrantLock
var Timer timer
val ir = org.openhab.core.model.script.ScriptServiceUtil.getItemRegistry

rule "E3 Power Change"
when
  Item e3_power changed
then
  if (previousState != NULL Math.abs(e3_power.getStateAs(Number).intValue - (previousState as Number).intValue) > 10) {
     val prev = (previousState as Number).floatValue < 60
     val cur = e3_power.getStateAs(Number).floatValue < 60
     if ((!prev && cur) || (prev && !cur))
       e3_posl_promjana.postUpdate(new DateTimeType)
   }

  e3Lock.lock
  try {
    if ((e3_power.state as Number).intValue < 5) {
      if (timer !== null) { timer.cancel; timer = null }
      if (e3_color.state != NULL) e3_color.postUpdate(NULL)
      return
    }

    timer = createTimer("Set E3 color", now.plusMinutes(4)) [|
      e3Lock.lock
      try {
         val last_mod = (e3_posl_promjana.state as DateTimeType).getZonedDateTime(ZoneId.systemDefault)
         var b = ZonedDateTime.now
         for (StringType c: newArrayList(new StringType(''), new StringType('yellow'), new StringType('gold'), new StringType('orange'), new StringType('maroon'))) {
           b = b.minusMinutes(4)
           if (last_mod.isAfter(b)) {
             if (timer !== null) timer.reschedule(now.plusMinutes(1))
             e3_color.postUpdate(if (c == '') NULL else c)
             return
           }
        }
      } finally {
        e3Lock.unlock
      }
      }
      if (e3_color.state != "red") {
        e3_color.postUpdate("red")
        timer = null
      }
    ]
  } finally {
    e3Lock.unlock
  }
end
```

dilyanpalauzov:
If a profile cannot run in parallel to itself, does this apply only for profiles written in automation language?  (There is such statement for transformations written in automation language). Or for all profiles?

concepts/rules.md
```diff
@@ -21,6 +21,14 @@ openHAB can send a notification to your phone.
 Other systems may have a concept of _Automations_, _Tasks_, and other terms.
 In openHAB, rules are used to implement all of these concepts.
 
+Rules can be executed, either when fired by their triggers, or when called explicitly - from a script, another rule, MainUI widget.
+When a rule is fired by its trigger, the execution of one and the same rule does not happen in parallel.
+If a rule is triggered for execution, while the rule is currently run, it will be queued and run later.
```
rkoshak: > If a rule is triggered for execution, while the rule is currently run, it will be queued and run later.

... while a rule is currently *running**...

configuration/rules-dsl.md
```diff
@@ -801,7 +801,7 @@ Caveat: Please note the semicolon after the return statement which terminates th
 
 ### Concurrency Guard
 
-If a rule triggers on UI events it may be necessary to guard against concurrency.
+If a rule is executed not fired by its triggers, but explicitly - from a script, from another rule, from a MainUI widget - the rule can be started, before its previous execution has ended.  It may be necessary to guard against concurrency.
```
rkoshak: Each sentence should be on a separate line.

The wording of the first sentence is awkward. I recommend something more like:

"If a rule is explicitly run from another script, rule, or MainUI widget instead of a trigger the rule can be started before the current execution has ended."

Nadahar:
> My understanding is that a timer in a script is in essence a new thread

Yes, unless https://github.com/openhab/openhab-core/pull/5069 is merged, in which case it isn't anymore.

That said, I agree that using a lock like you do above isn't hard, but there are some complicating factors. First, according to @rkoshak there are bugs with Rules DSL which means that `finally` won't always run if an exception occurs. That would leave the lock in question in a permanently locked state, and OH would have to be restarted to release it.

In addition, there are other scripting languages than Rules DSL where doing something like this is much more cumbersome. And, many people don't really understand the concept at all, why there is a need for locks or how you use them. For reasons not clear to me, concurrency seem to be one of those things that only some people "get". There are so many concurrency mistakes made, also by developers that are otherwise very capable.

dilyanpalauzov:
### On the text rewording here

I have taken the suggested text over.  In:

> If a rule is explicitly run from another script, rule, or MainUI widget instead of a trigger the rule can be started before the current execution has ended.

I have removed “or” before MainUI, as the list is not exclusive: HTTP POST `/rest/rules/{rule-uid}/runnow/` also executes the rule extraordinary. And I put commata around “instead of a trigger“.

### Bugs in Rules DSL …

> according to @rkoshak there are bugs with Rules DSL which means that finally won't always run if an exception occurs. That would leave the lock in question in a permanently locked state, and OH would have to be restarted to release it again.

Where are these bugs described?  Can I have an example reproducing this?  Can a lock be handled as a try-with-resources, so after `try` is left, the lock is released, without having something in `finally`?

(Moving Rules DSL out of openhab-core is another solution to the problem, or it lets isolate the problem either in DSL or in all rule executions systems - right now I do not know where it is.)

### Can non-script transformations run in parallel?

Sometimes - yes, sometimes - not!

Not only transformations from automations scripts are executed sequentially (one transformation per thread), but all transformations.  I modified the MAP transformation to sleep for 5s with 
[this transform.map.delay5.patch](https://github.com/user-attachments/files/22905971/transform.map.delay5.patch).

<details>
<summary>the patch to delay transformation.map with 5s</summary>

```diff
--- a/bundles/org.openhab.transform.map/src/main/java/org/openhab/transform/map/internal/MapTransformationService.java
+++ b/bundles/org.openhab.transform.map/src/main/java/org/openhab/transform/map/internal/MapTransformationService.java
@@ -84,8 +84,13 @@ public class MapTransformationService
 
     @Override
     public @Nullable String transform(String function, String source) throws TransformationException {
+        logger.error("A/" + Long.toString(Thread.currentThread().getId()) + "/ "
+                + java.time.ZonedDateTime.now().toString().substring(11, 22) + " input=[" + source + "]");
+        try {
+            Thread.sleep(5000);
+        } catch (Exception e) {
+        }
         Properties properties = null;
-
         Matcher matcher = INLINE_MAP_CONFIG_PATTERN.matcher(function);
         if (matcher.matches()) {
             properties = cachedInlineMap.computeIfAbsent(function, f -> {
@@ -120,15 +125,22 @@ public class MapTransformationService
             if (target == null) {
                 target = properties.getProperty("");
                 if (target == null) {
+                    logger.error("B/" + Long.toString(Thread.currentThread().getId()) + "/ "
+                            + java.time.ZonedDateTime.now().toString().substring(11, 22) + " input=[" + source
+                            + "] new Exception");
                     throw new TransformationException("Target value not found in map for '" + source + "'");
                 } else if (SOURCE_VALUE.equals(target)) {
                     target = source;
                 }
             }
-
+            logger.error("C/" + Long.toString(Thread.currentThread().getId()) + "/ "
+                    + java.time.ZonedDateTime.now().toString().substring(11, 22) + " input=[" + source + "] output="
+                    + target);
             logger.debug("Transformation resulted in '{}'", target);
             return target;
         }
+        logger.error("D/" + Long.toString(Thread.currentThread().getId()) + "/ "
+                + java.time.ZonedDateTime.now().toString().substring(11, 22) + " input=[" + source + "] new Exception");
         throw new TransformationException("Could not find configuration '" + function + "' or failed to parse it.");
     }
 
```

</details> 

<details open>
<summary>I have the MAP transformations called from</summary>

```
# grep -ri  [^e]map items/ things/ sitemaps/
items/r9.items:Switch r9 "9 Реле [MAP(switch.map):%s]"{channel="mqtt:topic:r9:r9", autoupdate="false"}
items/r9.items:Switch r9a "9 Реле [MAP(switch.map):%s]" {channel="mqtt:topic:r9:r9", autoupdate="false"}
items/r9.items:Switch r9b "9 Реле [MAP(switch.map):%s]" {channel="mqtt:topic:r9:r9", autoupdate="false"}
items/r9.items:Switch r9c "9 Реле [MAP(switch.map):%s]" {channel="mqtt:topic:r9:r9", autoupdate="false"}
items/r9.items:Switch r9d "9 Реле [MAP(switch.map):%s]" {channel="mqtt:topic:r9:r9", autoupdate="false"}
items/r9.items:Switch r9e "9 Реле [MAP(switch.map):%s]" {channel="mqtt:topic:r9:r9", autoupdate="false"}
items/r9.items:Switch r9f "9 Реле [MAP(switch.map):%s]" {channel="mqtt:topic:r9:r9", autoupdate="false"}
items/r3.items:Switch r3 "3 Реле [MAP(switch.map):%s]" {channel="mqtt:topic:r3:r3", autoupdate="false"}
sitemaps/r3.sitemap:        Switch item=r3_manual mappings=[ON="Ръчно", OFF="Автоматично"] labelcolor=[OFF="green", ON="red"] visibility=[!=NULL]
```
and the seven r9, r9a...r9f items appear in r9.sitemaps:
```
sitemap r9 label="R9" {
  Text item=r9
  Text item=r9a
  Text item=r9b
  Text item=r9c
  Text item=r9d
  Text item=r9e
  Text item=r9f
}
```

</details>

For the results below I visit `/settings/items` and the sitemap `/basicui/app?sitemap=x`.

However I do not understand the results.  
<details open>
<summary>These are the results from the MAP transformation logging in /var/log/openhab/openhab.log (for brevity I removed the date at the start of the lines and ` [ERROR] `)</summary>

```
15:12:23.613[ap.internal.MapTransformationService] - A/316/ 15:12:23.61 input=[ON]
15:12:28.618[ap.internal.MapTransformationService] - C/316/ 15:12:28.61 input=[ON] output=TurnedOn
15:12:28.654[ap.internal.MapTransformationService] - A/316/ 15:12:28.65 input=[ON]
15:12:33.656[ap.internal.MapTransformationService] - C/316/ 15:12:33.65 input=[ON] output=TurnedOn
15:12:33.660[ap.internal.MapTransformationService] - A/316/ 15:12:33.65 input=[ON]
15:12:38.661[ap.internal.MapTransformationService] - C/316/ 15:12:38.66 input=[ON] output=TurnedOn
15:12:38.674[ap.internal.MapTransformationService] - A/316/ 15:12:38.66 input=[ON]
15:12:38.686[ap.internal.MapTransformationService] - A/485/ 15:12:38.68 input=[ON]
15:12:40.251[ap.internal.MapTransformationService] - A/785/ 15:12:40.25 input=[ON]
15:12:41.655[ap.internal.MapTransformationService] - A/331/ 15:12:41.65 input=[ON]
15:12:43.676[ap.internal.MapTransformationService] - C/316/ 15:12:43.67 input=[ON] output=TurnedOn
15:12:43.680[ap.internal.MapTransformationService] - A/316/ 15:12:43.68 input=[ON]
15:12:43.688[ap.internal.MapTransformationService] - C/485/ 15:12:43.68 input=[ON] output=TurnedOn
15:12:43.697[ap.internal.MapTransformationService] - A/485/ 15:12:43.69 input=[ON]
15:12:45.256[ap.internal.MapTransformationService] - C/785/ 15:12:45.25 input=[ON] output=TurnedOn
15:12:45.263[ap.internal.MapTransformationService] - A/785/ 15:12:45.26 input=[ON]
15:12:46.656[ap.internal.MapTransformationService] - C/331/ 15:12:46.65 input=[ON] output=TurnedOn
15:12:46.664[ap.internal.MapTransformationService] - A/331/ 15:12:46.66 input=[ON]
15:12:48.682[ap.internal.MapTransformationService] - C/316/ 15:12:48.68 input=[ON] output=TurnedOn
15:12:48.691[ap.internal.MapTransformationService] - A/316/ 15:12:48.69 input=[ON]
15:12:48.699[ap.internal.MapTransformationService] - C/485/ 15:12:48.69 input=[ON] output=TurnedOn
15:12:48.705[ap.internal.MapTransformationService] - A/485/ 15:12:48.70 input=[ON]
15:12:50.265[ap.internal.MapTransformationService] - C/785/ 15:12:50.26 input=[ON] output=TurnedOn
15:12:50.272[ap.internal.MapTransformationService] - A/785/ 15:12:50.27 input=[ON]
15:12:51.666[ap.internal.MapTransformationService] - C/331/ 15:12:51.66 input=[ON] output=TurnedOn
15:12:51.673[ap.internal.MapTransformationService] - A/331/ 15:12:51.67 input=[ON]
15:12:53.694[ap.internal.MapTransformationService] - C/316/ 15:12:53.69 input=[ON] output=TurnedOn
15:12:53.697[ap.internal.MapTransformationService] - A/316/ 15:12:53.69 input=[ON]
15:12:53.707[ap.internal.MapTransformationService] - C/485/ 15:12:53.70 input=[ON] output=TurnedOn
15:12:53.711[ap.internal.MapTransformationService] - A/485/ 15:12:53.71 input=[ON]
15:12:55.274[ap.internal.MapTransformationService] - C/785/ 15:12:55.27 input=[ON] output=TurnedOn
15:12:55.278[ap.internal.MapTransformationService] - A/785/ 15:12:55.27 input=[ON]
15:12:56.675[ap.internal.MapTransformationService] - C/331/ 15:12:56.67 input=[ON] output=TurnedOn
15:12:56.679[ap.internal.MapTransformationService] - A/331/ 15:12:56.67 input=[ON]
15:12:58.699[ap.internal.MapTransformationService] - C/316/ 15:12:58.69 input=[ON] output=TurnedOn
15:12:58.713[ap.internal.MapTransformationService] - C/485/ 15:12:58.71 input=[ON] output=TurnedOn
15:12:58.717[ap.internal.MapTransformationService] - A/485/ 15:12:58.71 input=[ON]
15:13:00.279[ap.internal.MapTransformationService] - C/785/ 15:13:00.27 input=[ON] output=TurnedOn
15:13:00.283[ap.internal.MapTransformationService] - A/785/ 15:13:00.28 input=[ON]
15:13:01.681[ap.internal.MapTransformationService] - C/331/ 15:13:01.68 input=[ON] output=TurnedOn
15:13:01.685[ap.internal.MapTransformationService] - A/331/ 15:13:01.68 input=[ON]
15:13:03.718[ap.internal.MapTransformationService] - C/485/ 15:13:03.71 input=[ON] output=TurnedOn
15:13:03.726[ap.internal.MapTransformationService] - A/485/ 15:13:03.72 input=[ON]
15:13:05.284[ap.internal.MapTransformationService] - C/785/ 15:13:05.28 input=[ON] output=TurnedOn
15:13:05.291[ap.internal.MapTransformationService] - A/785/ 15:13:05.29 input=[ON]
15:13:06.686[ap.internal.MapTransformationService] - C/331/ 15:13:06.68 input=[ON] output=TurnedOn
15:13:06.692[ap.internal.MapTransformationService] - A/331/ 15:13:06.69 input=[ON]
15:13:08.728[ap.internal.MapTransformationService] - C/485/ 15:13:08.72 input=[ON] output=TurnedOn
15:13:08.735[ap.internal.MapTransformationService] - A/485/ 15:13:08.73 input=[ON]
15:13:10.294[ap.internal.MapTransformationService] - C/785/ 15:13:10.29 input=[ON] output=TurnedOn
15:13:10.301[ap.internal.MapTransformationService] - A/785/ 15:13:10.30 input=[ON]
15:13:11.695[ap.internal.MapTransformationService] - C/331/ 15:13:11.69 input=[ON] output=TurnedOn
15:13:11.701[ap.internal.MapTransformationService] - A/331/ 15:13:11.70 input=[ON]
15:13:13.739[ap.internal.MapTransformationService] - C/485/ 15:13:13.73 input=[ON] output=TurnedOn
15:13:13.759[ap.internal.MapTransformationService] - A/485/ 15:13:13.75 input=[ON]
15:13:15.304[ap.internal.MapTransformationService] - C/785/ 15:13:15.30 input=[ON] output=TurnedOn
15:13:15.322[ap.internal.MapTransformationService] - A/785/ 15:13:15.32 input=[ON]
15:13:16.703[ap.internal.MapTransformationService] - C/331/ 15:13:16.70 input=[ON] output=TurnedOn
15:13:16.721[ap.internal.MapTransformationService] - A/331/ 15:13:16.72 input=[ON]
15:13:18.761[ap.internal.MapTransformationService] - C/485/ 15:13:18.76 input=[ON] output=TurnedOn
15:13:20.324[ap.internal.MapTransformationService] - C/785/ 15:13:20.32 input=[ON] output=TurnedOn
15:13:21.724[ap.internal.MapTransformationService] - C/331/ 15:13:21.72 input=[ON] output=TurnedOn
```

</details>

That transformation is executed in threads 316, 331, 485, 785.  Thread 316 executes the transformation 7 times, the other execute the transformation 8 times.  Indeed `settings/items/` does not show the items with the transformation in their label.  My explanation is that 8 items apply the transformation to their labels, while the sitemap contains 7 items.  (Verified, this is the case).

So, one and the same transformation - script or not - is executed several times in the same thread, when
* several items have the same transformation in their label and all the items are included in a sitemap (seven executions above)
* several items have the same transformation in their label, are linked to channels (same or different channels) and the channel receives data (eight executions above). (I do not understand how this 8 happens when mqtt:topic:r3:r3 is not updated and mqtt:topic:r9:r9 is updated and ).

How can it be generalized when one and the same transformation runs several times (sequentially) in the same thread, and when in parallel using different threads?

# Parallelism in Profiles from transformations

I see no point in looking at how this works, before it is clarified when is one and the same transformation executed several times in the same thread and when in a different thread. (As I cannot know if the findings for profiles will be because of how profiles work, or because of how transformations work.)

Nadahar:
> Where are these bugs described? Can I have an example reproducing this? Can a lock be handled as a try-with-resources, so after `try` is left, the lock is released, without having something in `finally`?

I'm not sure that they are described except for in forum posts, probably. @rkoshak might have an idea how to trigger it. A `Lock` isn't `AutoClosable`, to no, you can't use try-with-resources, but that wouldn't help either from what I understand. As I understand the problem, there are some exceptions, like those from invalid casts, the makes the "Rules DSL parser/processor" crash, so that script execution is aborted there and then. There's nothing wrong with try-finally in itself, but if the execution is aborted during the block, `finally` will never be executed, and the same would be true for an `AutoClosable` resource.

From what I understand, this is a pretty fundamental flaw with xtext/xtend (I'm not sure exactly where) that nobody has been able to solve, and it's one of the reasons why focus was shifted away from Rules DSL. If you can't trust a `finally` block to be executed, you're basically left without options to control what will happen.

> (Moving Rules DSL out of openhab-core is another solution to the problem, or it lets isolate the problem either in DSL or in all rule executions systems - right now I do not know where it is.)

As I'm sure you know, I'm very much against moving it out of core, since it's the one BIG benefit of Rules DSL as I see it - that it's NOT an add-on. But, understand one thing: The different script languages have different challenges, some more than others. 

No one person have the overview of all this, and what comes up as examples are somewhat random. If you want to follow this to the end, and really find all the problems, I guess you will have to actually try to do it in all the different languages, and see what barriers present themselves. So, the focus on "if the bug is in Rules DSL or not" isn't really that important here. Rules were never meant to or designed to run in parallel with themselves as I understand it, so there are probably lots of small decisions that would have to be different, if they were meant to be.

What I don't understand is why you'd want a rule to run in parallel with itself. It sounds like a mess even without concurrency issues. A rule normally does some evaluation of something and then makes a change. If this evaluation → change process goes on in parallel, how can this be anything but a mess? One instance of the rule is evaluating while the other is applying the action, the next iteration has decided something different partly based on what the first one did, and so it will immediately change the decision made. It sounds like a recipe for "random behavior" to me.

When it comes to parallel execution of transformations, I'm not familiar enough with the code to know how it's done. But, what you're seeing could mean that they, in theory, run in parallel, but that for other reasons (the code that launches the transformations), they often aren't in effect run in parallel. There are lots of moving parts here, and one lock (in the "wrong place") can often be enough to prevent things from happening in parallel.

rkoshak:
> Where are these bugs described? Can I have an example reproducing this? Can a lock be handled as a try-with-resources, so after try is left, the lock is released, without having something in finally?Where are these bugs described? Can I have an example reproducing this? Can a lock be handled as a try-with-resources, so after try is left, the lock is released, without having something in finally?

I haven't used Rules DSL in many years and do not care to. But it used to be the case that if you had a type error in the try block (e.g. tried to do an operation with two Objects of incompatible types), Rules DSL would throw *and catch* and log a NullException. Because it caught its own exception, the rule just exits and is never given the chance to run the finally. 

There were other errors that got generated deep in the Xtend call stack which cause the rule to just stop. The exception never bubbles back up to the rule so the catch and finally never execute. 

> (Moving Rules DSL out of openhab-core is another solution to the problem, or it lets isolate the problem either in DSL or in all rule executions systems - right now I do not know where it is.)

I pushed really hard for this. But it's been decided. It's not going to happen. So we work with what we have. Ultimately, .items files, .things files, .sitemap, et. al. legacy file formats are defined by and parsed using Xtext/Xtend too. Moving Rules DSL to become an add-on doesn't actually remove much of anything from core because just about everything Rules DSL needs is also needed by these other file formats.

But I'm also not 100% certain something like this isn't also possible in other rules languages. There's a whole stack of Java stuff that a rule's scripts are running on top of. If there is an exception anywhere in that stack and the exception doesn't flow up to the script, you can *never* guarantee that the finally block will be run in any rule script. In practice, it probably doesn't happen that way, certainly not often, but there is also no guarantee.

> (As I cannot know if the findings for profiles will be because of how profiles work, or because of how transformations work.)

I suspect it's how the profile is called that makes the difference, same as with rules. The sitemap needs to make some REST API call to apply the transformation in the label so, as when a rule is run directly like that, it probably bypasses any locks. That's my guess anyway.

> What I don't understand is why you'd want a rule to run in parallel with itself. It sounds like a mess even without concurrency issues. A rule normally does some evaluation of something and then makes a change. If this evaluation → change process goes on in parallel, how can this be anything but a mess? One instance of the rule is evaluating while the other is applying the action, the next iteration has decided something different partly based on what the first one did, and so it will immediately change the decision made. It sounds like a recipe for "random behavior" to me.

This is exactly how it worked in OH 1.x and 2.x (for those not using the Experimental rule engine). Imagine the fun we had to deal with when using an "Item state updated" trigger on a Group with an aggregation function. The rule got triggered N-1 times where N is the number of members of the Group in rapid succession. 

I spent roughly 75% of my time on the forum helping users find, diagnose, and solve errors caused by the same rule running more than once at a time. It was indeed a mess. 

As I see it, there are only two ways you can make it so the same rule can run more than once at a time, and this applies to all the languages.

1. Rules can only be stateless. No more variables, no cache (unless all that locking stuff gets straightened out), simply a few lines of code which neither depend upon nor save any data from one run to the next. 

2. A separate instance (ScriptEngine) is created on the fly for every *execution* of a rule so if you have the same rule running twice, each has it's own thread, own ScriptEngine, and in all other ways are completely seprate from each other without anything shared. They are essentially two separate rules. 

1 isn't feasible because it eliminates pretty much every use case people have for rules. 

2 isn't feasible because, particularly on SBCs like RPis, the amount of time it takes to parse and instantiate the script is long. For a poorly created rule in Rules DSL (i.e. forcing types where you don't absolutely need to) it can take dozens of seconds. We can't afford that kind of latency every time the rule runs. I'm certain we could find ways to amortize and reuse stuff so it's not quite *that* bad, but it's never going to be down to acceptable levels of latency.  (Note, it used to be the case not that long ago that rules were not loaded and parsed/compiled until the first time the rule was triggered, and the user experience was horrible).

dilyanpalauzov:
> From what I understand, this is a pretty fundamental flaw with xtext/xtend (I'm not sure exactly where) 
> Ultimately, .items files, .things files, .sitemap, et. al. legacy file formats are defined by and parsed using Xtext/Xtend too.

Let me explain you my understanding of the difference between Xtext and Xbase/Xtend.

XText defines the grammar for sitemaps, items, things, rules files.

In grammar for rules I mean "start possibly with imports, then `var`s", then accept zero or more times “`rule "NAME" when`, then triggers, then `then`, then MAGIC BLOCK, then `end`”.

Xbase and Xtend are only in charge for the magic block above and DSL transformations.  Xbase and Xtend parse the code (only the magic block) and then can either interpret it, or generate .java files.  I think in openHAB it is interpreted, as it seems to run even on systems without javac, without implementation of `javax.toolx.JavaCompiler` (Zulu/Azul Java JRE).  If there is a fundamental flaw with Xtend/Xbase, then this fundamental flaw will be preserved in the generated (by Xtend) `.java` files.

In addition Xtext provides some services for Language Server support, and Xbase/Xtend might also provide such services.

In another addition Xtend/Xbase is used to compile parts of openHAB, e.g. bundles/org.openhab.core.model.sitemap/src/org/openhab/core/model/sitemap/SitemapRuntimeModule.xtend, but this is only compile time dependency.  So sitemaps depend on Xtend for their compilation, but at run time (after openHAB is shipped) model.sitemap, model.persistance, model.thing do not need Xtend, they need only Xtext (for parsing files).

<details>
<summary>All Xtend files in openHAB core</summary>

```
$ find -name '*.xtend'
./org.openhab.core.model.sitemap/src/org/openhab/core/model/sitemap/SitemapRuntimeModule.xtend
./org.openhab.core.model.sitemap/src/org/openhab/core/model/sitemap/SitemapStandaloneSetup.xtend
./org.openhab.core.model.sitemap/src/org/openhab/core/model/sitemap/formatting/SitemapFormatter.xtend
./org.openhab.core.model.sitemap/src/org/openhab/core/model/sitemap/scoping/SitemapScopeProvider.xtend
./org.openhab.core.model.sitemap/src/org/openhab/core/model/sitemap/generator/SitemapGenerator.xtend
./org.openhab.core.model.sitemap/src/org/openhab/core/model/sitemap/validation/SitemapValidator.xtend
./org.openhab.core.model.sitemap/src/org/openhab/core/model/sitemap/serializer/SitemapSyntacticSequencer.xtend
./org.openhab.core.model.sitemap/src/org/openhab/core/model/sitemap/serializer/SitemapSemanticSequencer.xtend
./org.openhab.core.model.script.ide/src/org/openhab/core/model/script/ide/ScriptIdeSetup.xtend
./org.openhab.core.model.script.ide/src/org/openhab/core/model/script/ide/ScriptIdeModule.xtend
./org.openhab.core.model.thing.ide/src/org/openhab/core/model/thing/ide/ThingIdeModule.xtend
./org.openhab.core.model.thing.ide/src/org/openhab/core/model/thing/ide/ThingIdeSetup.xtend
./org.openhab.core.model.item.ide/src/org/openhab/core/model/ide/ItemsIdeSetup.xtend
./org.openhab.core.model.item.ide/src/org/openhab/core/model/ide/ItemsIdeModule.xtend
./org.openhab.core.model.persistence/src/org/openhab/core/model/persistence/PersistenceRuntimeModule.xtend
./org.openhab.core.model.persistence/src/org/openhab/core/model/persistence/PersistenceStandaloneSetup.xtend
./org.openhab.core.model.persistence/src/org/openhab/core/model/persistence/formatting/PersistenceFormatter.xtend
./org.openhab.core.model.persistence/src/org/openhab/core/model/persistence/scoping/PersistenceScopeProvider.xtend
./org.openhab.core.model.persistence/src/org/openhab/core/model/persistence/generator/PersistenceGenerator.xtend
./org.openhab.core.model.persistence/src/org/openhab/core/model/persistence/validation/PersistenceValidator.xtend
./org.openhab.core.model.persistence/src/org/openhab/core/model/persistence/serializer/PersistenceSemanticSequencer.xtend
./org.openhab.core.model.persistence/src/org/openhab/core/model/persistence/serializer/PersistenceSyntacticSequencer.xtend
./org.openhab.core.model.rule.ide/src/org/openhab/core/model/rule/ide/RulesIdeModule.xtend
./org.openhab.core.model.rule.ide/src/org/openhab/core/model/rule/ide/RulesIdeSetup.xtend
./org.openhab.core.model.sitemap.ide/src/org/openhab/core/model/sitemap/ide/SitemapIdeSetup.xtend
./org.openhab.core.model.sitemap.ide/src/org/openhab/core/model/sitemap/ide/SitemapIdeModule.xtend
./org.openhab.core.model.item/src/org/openhab/core/model/ItemsRuntimeModule.xtend
./org.openhab.core.model.item/src/org/openhab/core/model/ItemsStandaloneSetup.xtend
./org.openhab.core.model.item/src/org/openhab/core/model/formatting/ItemsFormatter.xtend
./org.openhab.core.model.item/src/org/openhab/core/model/scoping/ItemsScopeProvider.xtend
./org.openhab.core.model.item/src/org/openhab/core/model/generator/ItemsGenerator.xtend
./org.openhab.core.model.item/src/org/openhab/core/model/validation/ItemsValidator.xtend
./org.openhab.core.model.item/src/org/openhab/core/model/serializer/ItemsSemanticSequencer.xtend
./org.openhab.core.model.item/src/org/openhab/core/model/serializer/ItemsSyntacticSequencer.xtend
./org.openhab.core.model.rule/src/org/openhab/core/model/rule/RulesStandaloneSetup.xtend
./org.openhab.core.model.rule/src/org/openhab/core/model/rule/formatting/RulesFormatter.xtend
./org.openhab.core.model.rule/src/org/openhab/core/model/rule/scoping/RulesScopeProvider.xtend
./org.openhab.core.model.rule/src/org/openhab/core/model/rule/validation/RulesValidator.xtend
./org.openhab.core.model.rule/src/org/openhab/core/model/rule/jvmmodel/RulesJvmModelInferrer.xtend
./org.openhab.core.model.rule/src/org/openhab/core/model/rule/serializer/RulesSemanticSequencer.xtend
./org.openhab.core.model.rule/src/org/openhab/core/model/rule/serializer/RulesSyntacticSequencer.xtend
./org.openhab.core.model.rule/src/org/openhab/core/model/rule/RulesRuntimeModule.xtend
./org.openhab.core.model.persistence.ide/src/org/openhab/core/model/persistence/ide/PersistenceIdeSetup.xtend
./org.openhab.core.model.persistence.ide/src/org/openhab/core/model/persistence/ide/PersistenceIdeModule.xtend
./org.openhab.core.model.thing/src/org/openhab/core/model/thing/internal/GenericThingProvider.xtend
./org.openhab.core.model.thing/src/org/openhab/core/model/thing/ThingRuntimeModule.xtend
./org.openhab.core.model.thing/src/org/openhab/core/model/thing/ThingStandaloneSetup.xtend
./org.openhab.core.model.thing/src/org/openhab/core/model/thing/formatting/ThingFormatter.xtend
./org.openhab.core.model.thing/src/org/openhab/core/model/thing/scoping/ThingScopeProvider.xtend
./org.openhab.core.model.thing/src/org/openhab/core/model/thing/generator/ThingGenerator.xtend
./org.openhab.core.model.thing/src/org/openhab/core/model/thing/validation/ThingValidator.xtend
./org.openhab.core.model.thing/src/org/openhab/core/model/thing/serializer/ThingSyntacticSequencer.xtend
./org.openhab.core.model.thing/src/org/openhab/core/model/thing/serializer/ThingSemanticSequencer.xtend
./org.openhab.core.model.script/src/org/openhab/core/model/script/internal/engine/ServiceTrackerThingActionsProvider.xtend
./org.openhab.core.model.script/src/org/openhab/core/model/script/internal/engine/ServiceTrackerActionServiceProvider.xtend
./org.openhab.core.model.script/src/org/openhab/core/model/script/ScriptRuntimeModule.xtend
./org.openhab.core.model.script/src/org/openhab/core/model/script/formatting/ScriptFormatter.xtend
./org.openhab.core.model.script/src/org/openhab/core/model/script/scoping/ScriptScopeProvider.xtend
./org.openhab.core.model.script/src/org/openhab/core/model/script/ScriptStandaloneSetup.xtend
./org.openhab.core.model.script/src/org/openhab/core/model/script/interpreter/ScriptInterpreter.xtend
./org.openhab.core.model.script/src/org/openhab/core/model/script/validation/ScriptValidator.xtend
./org.openhab.core.model.script/src/org/openhab/core/model/script/jvmmodel/ScriptJvmModelInferrer.xtend
./org.openhab.core.model.script/src/org/openhab/core/model/script/serializer/ScriptSyntacticSequencer.xtend
./org.openhab.core.model.script/src/org/openhab/core/model/script/serializer/ScriptSemanticSequencer.xtend
```
</details>

That said, removing Rules DSL and DSL transformation from openhab-core means also removing Xtend and Xbase.  It does not mean removig Xtext.

In fact Xtend/Xtext would be very fine, if it can run without the Rules grammar.  That is to be able to write .xtend files in automation/jsr223/, which are either interpreted or converted to .java files, but do not have this `rule NAME when … then … end` grammar extras (or limitations).

# Execting a rule to itself in parallel

> As I see it, there are only two ways you can make it so the same rule can run more than once at a time, and this applies to all the languages.
> * Rules can only be stateless. No more variables, no cache (unless all that locking stuff gets straightened out), simply a few lines of code which neither depend upon nor save any data from one run to the next.
> * A separate instance (ScriptEngine) is created on the fly for every execution of a rule so if you have the same rule running twice, each has it's own thread, own ScriptEngine, and in all other ways are completely seprate from each other without anything shared. They are essentially two separate rules.


To run a rule in parallel would actually mean that the only thing a rule does is to start a new thread and execute a function in that thread.  Then the rule returns (immediately), before the function in the thread has completed.  This can be done even now and as it can be done now, the rules can run with the same amount of ScriptEngines as now and share data in the same way as now.  The only drawback is that a rule, which returns immediately cannot return a value.  But this value is useful only, if a …

Okay, above we talk about rules, but we mean rule actions.  So if a rule has many actions, when each action returns, it passes data to the next action.  And when the action returns immediately, then the next action cannot use the return value of the previous action.  Except the previous action returned a Promise… and who needs several actions to a rule… That said, as it is now, as people can fire threads from (slow) rules and return immeditely, the same rule can run virtually in parallel to itself.

The other only drawback is that starting new threads/executing actions in thread pools will eat more resources than now.

However the above does not apply to transformations, as they need a String return value immediately.

Nadahar:
> To run a rule in parallel would actually mean that the only thing a rule does is to start a new thread and execute a function in that thread. Then the rule returns (immediately), before the function in the thread has completed. This can be done even now and as it can be done now, the rules can run with the same amount of ScriptEngines as now and share data in the same way as now. The only drawback is that a rule, which returns immediately cannot return a value. But this value is useful only, if a …
> 
> Okay, above we talk about rules, but we mean rule actions. So if a rule has many actions, when each action returns, it passes data to the next action. And when the action returns immediately, then the next action cannot use the return value of the previous action. Except the previous action returned a Promise… and who needs several actions to a rule… That said, as it is now, as people can fire threads from (slow) rules and return immeditely, the same rule can run virtually in parallel to itself.
> 
> The other only drawback is that starting new threads/executing actions in thread pools will eat more resources than now.
> 
> However the above does not apply to transformations, as they need a String return value immediately.

I don't understand what you're trying to say here. If you launch new threads (which you can basically do anywhere where you're allowed to "run code", those new threads aren't "rules". They are just threads, they don't interact with OH as part of a rule, they aren't treated as such, and you're "completely on your own" as to what these achieve. I don't think such "abusive techniques" need to be a part of the regular documentation.

If doing that serves a purpose for you, please go ahead and do it, but for me, it's very difficult to understand what I could achieve by doing something like this.

rkoshak:
As for removing Rules DSL from OH, you can go tilt at that windmill. As far as I'm concerned, the developers/maintainers have spoken definitively. Rules DSL is not going to be removed from core.

> To run a rule in parallel would actually mean that the only thing a rule does is to start a new thread and execute a function in that thread.

And those separate runs of the function *share data*. In fact, they don't just share data, they share memory. 

step | function 1 | function 2
-|-|-
1 | starts running |
2  | sets variable `a` to `"foo"` |
3  |  | starts running
4  |  | sets variable `a` to `"bar"`|
5  | `console.info(a)` |
6  | | `console.info(a)` 

What gets logged in step 5? What gets logged in step 6? In both cases `"bar"`. function 2 overwrote `"foo"` and now *all* executing functions only see `"bar"`.

*This* is why you can't run the same script more than once at the same time except under the restrictions I outlined above. 

And yes, this is *exactly* the sorts of things we had to deal with back when rules could run at the same time.

> and who needs several actions to a rule… 

Anyone using "simple UI rules" (i.e. using any action that isn't an "Inline Script"). Granted this is exceptionally limited right now but Scripts are not the only Rule Action.

dilyanpalauzov:
The focus of the current changeset is to write down what cannot run in parallel, what can run in parallel, what only sometimes runs in parallel - file based rules, UI rules, file-script, UI scripts, actions, profiles, profiles linked to scripting transformations, profiles linked to non-scripting transformations, scripting transformations, non-scripting transformations…

Currently it is not clear when a non-scripting transformation reuses a thread (runs sequentially - is started, after it finished) and when it runs in a new thread (parallel to itself).

But as an exception to what I just said:
> developers/maintainers have spoken definitively. Rules DSL is not going to be removed from core.

why does https://github.com/openhab/openhab-core/issues/3233 exist as it is?

rkoshak:
> why does https://github.com/openhab/openhab-core/issues/3233 exist as it is?

Because he was also in favor of moving Rules DSL to an add-on. But the consensus of the rest of the maintainers was to leave it part of core. So that Issue was never worked and it probably should have been closed.

florian-h05:
> Because he was also in favor of moving Rules DSL to an add-on. But the consensus of the rest of the maintainers was to leave it part of core. So that Issue was never worked and it probably should have been closed.

I don't think there was a consesus against it, just no one implementing it.

Let me know when you think this PR is correct and ready @rkoshak

rkoshak:
I think the text as written in the PR is correct for now. But they might change based on some of the PRs that are still open I think. It's probably better to merge this now and wait for those changes rather than holding back. So I'll say it looks good to me.

concepts/rules.md
```diff
@@ -21,6 +21,14 @@ openHAB can send a notification to your phone.
 Other systems may have a concept of _Automations_, _Tasks_, and other terms.
 In openHAB, rules are used to implement all of these concepts.
 
+Rules can be executed, either when fired by their triggers, or when called explicitly - from a script, another rule, MainUI widget.
```
florian-h05:
```suggestion
Rules can be executed, either when fired by their triggers, or when called explicitly - from a script, another rule, or a Main UI widget.
```

concepts/rules.md
```diff
@@ -21,6 +21,14 @@ openHAB can send a notification to your phone.
...
+When a rule is fired by its trigger, the execution of one and the same rule does not happen in parallel.
+If a rule is triggered for execution, while the rule is currently running, it will be queued and run later.
+
+When a rule is called explicitly - from a script, another rule with [`RuleManager.runNow()`](https://www.openhab.org/javadoc/latest/org/openhab/core/automation/rulemanager), from a MainUI widget, then this rule can be executed in parallel.
```
florian-h05:
```suggestion
When a rule is called explicitly - e.g. from a script, another rule with [`RuleManager.runNow()`](https://www.openhab.org/javadoc/latest/org/openhab/core/automation/rulemanager), or from a Main UI widget - then this rule can be executed in parallel.
```

configuration/rules-dsl.md
```diff
@@ -801,7 +801,8 @@ Caveat: Please note the semicolon after the return statement which terminates th
 
 ### Concurrency Guard
 
-If a rule triggers on UI events it may be necessary to guard against concurrency.
+If a rule is explicitly run from another script, rule, MainUI widget, instead of a trigger, the rule can be started before the current execution has ended.
```
florian-h05:
```suggestion
If a rule is explicitly run from another script, rule, or a Main UI widget, instead of a trigger, the rule can be started before the current execution has ended.
```

Nadahar:
The one thing I don't get with documenting that rules can be executed in parallel when started "externally", is that this doesn't actually *work* unless the rule is written in a language that allows concurrency *and* synchronization is handled in the rule script.

By documenting that it can be done, without mentioning the caveats, it does sound like this is something you could/should do, not that it will get you into a world of trouble.

concepts/rules.md
dilyanpalauzov: HTTP POST `/rest/rules/{rule-uid}/runnow/` also executes the rule extraordinary.  The lack of “or” is supposed to indicate that the list is not exclusive.

configuration/rules-dsl.md
dilyanpalauzov: “BasicUI” appears here 20 times (`git grep "BasicUI"|wc -l`), “Basic UI” appears 10 times, “MainUI” appears 38 times and “Main UI” appears 76 times.

With or without space after Main and Basic and before UI?

dilyanpalauzov: In the openhab-website repository, “Main UI” appears 66 times, MainUI appears 8 times, not counting variable and directory names; BasicUI appears approximately 19 times and "basic ui" appears approximately 24 times, not counting translations in .properties files.

dilyanpalauzov:
The purpose of the current changes is to indicate to rule writers when they should use locks.  I think indicating what can be a critical section is enough.  If somebody wants to include in the documentation of openHAB how to deal with critical sections, such additions can be made independent of the current one.

> I think it would be good to mention that automation add-ons can put additional synchronisation into place.

This should be documented by the automation add-ons.  I see no reason to bother the reader here with extra sentences, if the reader is not going to use such automation add-ons.

> … By documenting that it can be done, without mentioning the caveats, it does sound like this is something you could/should do, not that it will get you into a world of trouble.

Do you want to propose concrete sentences for inclusion?

Nadahar:
> Do you want to propose concrete sentences for inclusion?

Personally, I would prefer that https://github.com/openhab/openhab-core/pull/5069 was merged to get an end to this madness, and there would be no reason to try to express the details of what is essentially "broken".

As has been pointed out several times, there are numerous problems here. JavaScript, which seems to be the most broadly used rule scripting language today, doesn't support this at all, and will probably fail if you try to do this. Rules DSL can do it, but has some flaws that means that you risk leaving the lock in a locked state, where any thread that tries to run the rule will be waiting forever/hung. Other languages like Python and Ruby have varying levels of "issues" doing this, and it would need to be tested (I can't predict what would happen). Using locks from either isn't very convenient, and the users of those languages aren't used to it.

There's also the fact that OH itself doesn't expect the same rule to run in parallel. Everything is written with the assumption that things happen sequentially, so it's actually "just luck" if it works to do it concurrently. You also have the very "fundamental challenge" that rules will usually react to some state change and make another change as a consequence. Having multiple versions of this logic running at one, will usually be a bad thing, regardless of thread-safety concerns. You will quickly come to the situation where the wrong decision is made, because another thread is in the middle of the same evaluation, so by the time a decision is made, the circumstances under which the decision was just made, is no longer valid. This is why I'm saying that it's hard to imagine when it would actually be desirable for the same rule to run concurrently.

I'm sure there are cases where it would be desirable, but I'm pretty sure they would be a very small minority of cases.

When it comes to transformations, I see it differently, they "should" be able to run in parallel IMO (but there are challenges). For rules, I think the most flexible would be if you could enable concurrent execution for individual rules, but that it was disabled by default. But, that should mean that they could run in parallel, whether triggered by a trigger or "manually", via a timer or another rule.  And it would require some changes to the underlying system in OH, because currently, there is no way to allow multiple threads to handle a rule.

I can't see that there is any situation where having it like it is, with this "special behavior" when the rule is executed using `runNow()`, is desirable, so I think it would be a very bad thing if users starts to adapt to this.

Even if you don't explain how to use locks, it should be explained that you *need to* do it, if you run rules in parallel and ignore all the other challenges with doing this. I would also say that the other challenges should be mentioned, like the fact that you can't really evaluate a state and assume that it stays true until you take action. It should also be explained that not all scripting languages can do this.

That said, maybe too much is made from this little tweak to the documentation. IMO, it would be more important to get https://github.com/openhab/openhab-core/pull/5069 moving and avoid all these problems.

dilyanpalauzov:
I suggest in the beginning of December, if openhab/openhab-core#5069 is still pending, that the currently proposed changes are merged, so that they are available when OH 5.1 is released.

Or I have even better suggestion: the current changes are integrated, and if openhab/openhab-core#5069 is applied, then a notice is inserted, that the information is relevant for OH < 5.1. (I use OH 4.3).

Will openhab/openhab-core#5069 also handle the cases with UI Widgets triggered events? (Does it utilize runnow under the hood?)

It is true that in a parallel world a state can be read and handled later, where the state has changed in the meantime and this is not mentioned explicitly.  It is however implied by the fact that things can run in parallel.  Even with sequential rule execution, a rule executed when item A has changed, if the rule reads the state of item B and then does something based on the states of A and B, when the rule actions are done the state of B can be completely different.  So the current rewording does not solve all possible problems and is not encyclopedia for concurrency, but it makes things, specific to openHAB, which things cannot be read elsewhere, clearer.

rkoshak:
> Or I have even better suggestion: the current changes are integrated, and if https://github.com/openhab/openhab-core/pull/5069 is applied, then a notice is inserted, that the information is relevant for OH < 5.1. (I use OH 4.3).

That's not how the docs work. In rare cases, a change can be made to the docs for previous versions of OH, but otherwise the "next.openhab.org/docs" (i.e. latest) is intended to be for 5.1 and not have a bunch of stuff referring to older versions. 

> Will https://github.com/openhab/openhab-core/pull/5069 also handle the cases with UI Widgets triggered events? (Does it utilize runnow under the hood?)

Yes.

> It is true that in a parallel world a state can be read and handled later, where the state has changed in the meantime and this is not mentioned explicitly. 

If you are referring to an Item's state, yes. When you get the state of the Item, you get it at that moment. The Item may change state later but what ever variable you saved the state to will still be the old state. This is why for Item triggered rules the name of the triggering Item and the state/command that caused the rule to trigger are captured in the `event`. 

This has *always* been the case, but I don't know if it's explicitly mentioned in the docs anywhere. It's mostly just a side effect of the OH architecture.

dilyanpalauzov:
> That's not how the docs work. In rare cases, a change can be made to the docs for previous versions of OH,

I do not see it this way, some aspects of openHAB are under-documented, this is e.g. the circumstance whether a rule can run in parallel to itself, if a transformation can run in parallel to itself, the available JSR223 presets - media, lifecycle, provider; the content of the listed presets, the circumstance that `.things`, `.items`, `.sitemap`, `.persistence` files can have comments - `//` until the end of the line or `/*` mutlilined comment `*/`.  When writing about these under-documented things they are sometimes relevant for all previous versions but the documentation for the previous versions is rarely updated.  So it is best when writing new information  to spell for which versions of openHAB it relates, or does not relate, if this can be easily determined and makes sense.

rkoshak:
And yet the policy for the docs has always been to not do iot that way. :shrug: I'm not a maintainer on this repo but you will probably run into issues getting you PR accepted if you go against the policy.

dilyanpalauzov:
Aligning to policies, like [removing all questions from a proposed change](https://github.com/openhab/openhab-docs/pull/2578#issuecomment-3448065504), does not mean that [within reasonable times the changes will be accepted](https://github.com/openhab/openhab-docs/pull/2487).

dilyanpalauzov:
> I suggest in the beginning of December, if https://github.com/openhab/openhab-core/pull/5069 is still pending, that the currently proposed changes are merged, so that they are available when OH 5.1 is released.

As it is short before a release, can the current change be included in 5.1?

Nadahar:
> As it is short before a release, can the current change be included in 5.1?

That's a good question - I don't know who can answer it.

configuration/rules-dsl.md
florian-h05: It should be Main UI and Basic UI IMO.

concepts/rules.md
```diff
@@ -21,6 +21,14 @@ openHAB can send a notification to your phone.
 Other systems may have a concept of _Automations_, _Tasks_, and other terms.
 In openHAB, rules are used to implement all of these concepts.
 
+Rules can be executed, either when fired by their triggers, or when called explicitly - from a script, another rule, or a Main UI widget.
+When a rule is fired by its trigger, the execution of one and the same rule does not happen in parallel.
+If a rule is triggered for execution, while the rule is currently running, it will be queued and run later.
+
+When a rule is called explicitly - e.g. from a script, another rule with [`RuleManager.runNow()`](https://www.openhab.org/javadoc/latest/org/openhab/core/automation/rulemanager), or from a Main UI widget - then this rule can be executed in parallel.
```
florian-h05, line 28: In parallel with other explicit invocations and triggered executions, correct?

configuration/rules-dsl.md
florian-h05:
```suggestion
If a rule is explicitly run from another script, rule, a Main UI widget, instead of a trigger, the rule can be started before the current execution has ended.
```

concepts/rules.md
Nadahar: In parallel with "anything", since it's run in using a "normal" thread pool instead of the reserved rule thread.

concepts/rules.md
```diff
@@ -21,7 +21,7 @@ openHAB can send a notification to your phone.
 Other systems may have a concept of _Automations_, _Tasks_, and other terms.
 In openHAB, rules are used to implement all of these concepts.
 
-Rules can be executed, either when fired by their triggers, or when called explicitly - from a script, another rule, MainUI widget.
+Rules can be executed, either when fired by their triggers, or when called explicitly - from a script, another rule, or a Main UI widget.
```
dilyanpalauzov: There should be no `or` here.  It is possible to execute a rule by calling RunNow.  So the list is not exclusive and therefore there should be no final `or`.

Nadahar: That's not correct, that's not a rule in English and the first form is just invalid/incomplete. Perhaps this rule exists in another language?

dilyanpalauzov: Omitting conjunctions in a list it discussed at https://english.stackexchange.com/questions/15970/omitting-and-in-a-sentence .

https://www.litcharts.com/literary-devices-and-terms/asyndeton says “Lists that do not include conjunctions leave open the possibility that there could be more elements within a series.”

Nadahar: They are discussing "creative writing", not formal documentation, and the way it's used in the example there is much more "acceptable" than simply ending a sentence in what feels like without completing it.

What I would typically do in such a situation, where I don't want to give the impression that the list is exclusive, is:
"for example from a script, another rule or a Main UI widget."

rkoshak: > That's not correct, that's not a rule in English and the first form is just invalid/incomplete. Perhaps this rule exists in another language?

It's not a rule of grammar for English in general, but there can be rules set down in style guides and context.

Usually, in technical or academic writing, asyndetons are to be avoided or used sparingly and in a way that does not disrupt the formal tone of the writing. 

OH docs do not really follow any specific writing style guide (e.g. MLA, Chicago, AP, IEEE, etc.) but it would qualify as technical and therefore more formal writing. Probably the guide that would be most applicable to the OH docs would be IEEE which, while not mentioning asyndeton by name, has rules which effectively discourage its use in favor of syndetic writing. I can find the exact reference if desired, but this AI summary matches with what I know.

> The IEEE Editorial Style Manual and associated technical writing standards do not explicitly mention "asyndeton" by name. Instead, they provide strict grammatical rules for series and lists that effectively discourage asyndetons in favor of clarity and precision. 
> 
>The consensus for IEEE style can be summarized through its established rules on series and clarity:
Requirement for Conjunctions: IEEE style generally requires a coordinating conjunction (e.g., "and", "or") before the final term in a series to ensure the relationship between items is unambiguous.
The Serial (Oxford) Comma: IEEE dictates using a comma after each term in a series of three or more, followed by a conjunction (e.g., "A, B, and C"). This rule directly contradicts the structure of asyndeton ("A, B, C").
Clarity over Rhetoric: In technical fields like engineering and computing, the primary goal is unambiguous communication. Asyndetons can create "choppy" prose or leave the reader uncertain if a list is complete, which violates IEEE’s emphasis on professionalism and consistency.
Deferral to Chicago Style: For grammatical points not explicitly covered in its own manual, IEEE officially defers to The Chicago Manual of Style. Chicago permits asyndeton in narrative or literary prose but suggests it be avoided in formal or technical contexts where logic and flow are paramount.
> 
> Context | Preferred (Syndetic) | Discouraged (Asyndetic)
> -- | -- | --
> Technical Series | The system requires a sensor, a processor, and a battery. | The system requires a sensor, a processor, a battery.
> Author Lists | J. Smith and W. Jones [9]. | J. Smith, W. Jones [9].
> Equations | I=1,2,3,…,ncap I equals 1 comma 2 comma 3 comma … comma n𝐼=1,2,3,…,𝑛 | I=1,2,3cap I equals 1 comma 2 comma 3𝐼=1,2,3 (unclear if it continues)

In formal writing, if you wanted to omit the "or" and indicate that the list is not complete use "etc.".

```
Rules can be executed, either when fired by their triggers, or when called explicitly - from a script, another rule, a Main UI widget, etc.
```

While this seems to violate the rule, the "or" is implit in the word "etcetera" so can be omitted.

dilyanpalauzov:
> > Where are these bugs described? Can I have an example reproducing this? Can a lock be handled as a try-with-resources, so after try is left, the lock is released, without having something in finally?Where are these bugs described? Can I have an example reproducing this? Can a lock be handled as a try-with-resources, so after try is left, the lock is released, without having something in finally?
> 
> I haven't used Rules DSL in many years and do not care to. But it used to be the case that if you had a type error in the try block (e.g. tried to do an operation with two Objects of incompatible types), Rules DSL would throw _and catch_ and log a NullException. Because it caught its own exception, the rule just exits and is never given the chance to run the finally.
> 
> There were other errors that got generated deep in the Xtend call stack which cause the rule to just stop. The exception never bubbles back up to the rule so the catch and finally never execute.

This is the explanation for the above, and how can it be avoided:

From [XbaseInterpreter.java](https://github.com/eclipse-xtext/xtext/blob/main/org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/interpreter/impl/XbaseInterpreter.java) two methods are overridden in core.model.interpreter.Script.xtend: `invokeFeature(JvmIdentifiableElement feature, XAbstractFeatureCall featureCall, Object receiverObj, IEvaluationContext context, CancelIndicator indicator)` and `_doEvaluate(XCastedExpression castedExpression, IEvaluationContext context, CancelIndicator indicator)`.

For the latter the method in the super class does `throw new EvaluationException(new ClassCastException(typeName))` when a cast cannot be performed.  If that cast is in a try{…}catch…finally{…} clase, the exception is caught by Xbase and the catch/finally blocks are executed.  The overriden method in Script.xtend instead does `throw new ScriptExecutionException(new ScriptError("Could not cast … ", castedExpression));`.  This aborts the Xbase Interpreter and the interpreter does not execute the catch…finally blocks.  https://github.com/openhab/openhab-core/pull/5329 does change `throw new ScriptExecutionException(new ScriptError(…))` with `throw new EvaluationException(new ClassCastException(…))`.  This will allow to catch cast exceptions within DSL Rules/Script and execute the `catch` and `finally` blocks.

Similar change is needed also in the overriden method `invokeFeature(JvmIdentifiableElement, XAbstractFeatureCall, Object, IEvaluationContext, CancelIndicator)` so that calling that method does not abort the Xbase interpreter by executing `throw new ScriptExecutionException(new ScriptError(`, and the interprerer executes `catch` and `finally` statements.

Nadahar:
> This is the explanation for the above, and how can it be avoided:

This is very useful, I get the feeling that very few OH developers are very "deep" into Xbase/text/tend these days, with it suffering as a consequence. I don't know if myself because I can't open those packages without Eclipse going into crazy build loops, which prevents me from getting to know the code.

It sounds to me like these fixes could get rid of some "deep rooted" problems.

PR #2580, opened 2025-10-06
Add guidelines for event source

ccutrer:


developers/utils/events.md
```diff
@@ -40,7 +40,19 @@ The topics of openHAB events are divided into the following four parts: `{namesp
 The type of an event is represented by a string, usually the name of the concrete event implementation class, e.g. `ItemCommandEvent`, `ItemUpdatedEvent` .
 This string type presentation is used by event subscribers for event subscription (see chapter "Receive Events") and by the framework for the creation of concrete event instances.
 
-The event source is optional and represents the name of the source identifying the sender.
+The event source identifies the sender.
+Not all senders will set the source.
+The source should be structured, and of the form `<sending package>[$<actor>][><sending package$<actor>]+`
```
florian-h05: Don't we want to make this a real arrow?

```suggestion
The source should be structured, and of the form `<sending package>[$<actor>][-><sending package$<actor>]+`
```

developers/utils/events.md
```diff
@@ -40,7 +40,19 @@ The topics of openHAB events are divided into the following four parts: `{namesp
...
+The sending package is the identifier of the component that sent the event.
```
florian-h05: Can you please include that this is (meant to be) the package name?

developers/utils/events.md
```diff
@@ -40,7 +40,19 @@ The topics of openHAB events are divided into the following four parts: `{namesp
...
+The optional actor piece is an identifier of some sort that identifies the end user or device that initiated the event.
```
florian-h05:
```suggestion
The optional actor piece is an identifier of some sort that identifies the end user, device or rule that initiated the event.
```

developers/utils/events.md
```diff
@@ -40,7 +40,19 @@ The topics of openHAB events are divided into the following four parts: `{namesp
...
+If another event is triggered in response to an initial event, it should preserve the original source, and append (separated by `>`) another identifier for itself.
```
florian-h05:
```suggestion
If another event is triggered in response to an initial event, it should preserve the original source, and append (separated by `->`) another identifier for itself.
```

See my first comment.

developers/utils/events.md
```diff
@@ -40,7 +40,19 @@ The topics of openHAB events are divided into the following four parts: `{namesp
...
+A few examples:
+
+- `org.openhab.ui$admin` would mean the event was initiated by a user named `admin` via MainUI.
```
florian-h05:
```suggestion
- `org.openhab.ui$admin` would mean the event was initiated by a user named `admin` via Main UI.
```

developers/utils/events.md
ccutrer: I purposely used a single character here both for simplicity of splitting a string, and to not use a hyphen character that is common in hostnames, GUIDs, and thing/channel UIDs.

florian-h05: But isn't it easier to split by `->` than by `>` only?
We could also use `=>` as arrow ...

florian-h05:
@ccutrer I‘ve seen your add-on PRs, but I think it would be good to merge the documentation here first. Please let me know what you think about my comments.

developers/utils/events.md
ccutrer: I used `=>`

PR #2585, opened 2025-10-13
JSR223: Add `busEvent` to the ScriptAction preset

dilyanpalauzov:
Depends on https://github.com/openhab/openhab-core/pull/5064

### DIFF
--- a/configuration/jsr223.md
+++ b/configuration/jsr223.md
@@ -421,9 +421,12 @@ scriptExtension.importPreset("RuleFactories")
 
 This preset can be useful for scheduling asynchronous code execution with `org.openhab.core.automation.module.script.action.Timer`.
 
-| Variable          | Description                                                        |
-|-------------------|--------------------------------------------------------------------|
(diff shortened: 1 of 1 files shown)


dilyanpalauzov:
My question from https://github.com/openhab/openhab-core/pull/5064#issuecomment-3397492778 how will the `events` variable from the `default` preset be different from the `busEvent` variable from the `ScriptAction` preset was not answered.  This is now the same value exposed over two different jsr223-variables, as the next Groovy script demonstrates by printing `true`:
```groovy
org.slf4j.LoggerFactory.getLogger("G").error(Boolean.valueOf(
  scriptExtension.importPreset("ScriptAction").get("busEvent") ==  scriptExtension.importPreset("default").get("events")
).toString())
```

florian-h05:
Exactly, those two are now the same.
Previously, they were different classes, but had similar methods with similar signature. The core change reduced code duplication, and providing two ways of accessing the BusEvent doesn't hurt.

PR #2586, opened 2025-10-15
ScriptExecution actions: callScript() returns a value

dilyanpalauzov:
… because the loaded script is in Xbase, in Xbase everything is an expression, there are no statements.

In any case, here a test: in openhab/scripts/a.script I put
```
logError('A', "STARTED")
7
```
and in rules/z.rules
```
rule "change"
when Item r3 changed
then
  logError("B", "when Item r3 changed")
  logError("D", "M " + callScript("a"))
end

```
Then I get in openhab.log:
```
2025-10-15 19:30:56.904 [ERROR] [org.openhab.core.model.script.A   ] - when Item r3 changed
2025-10-15 19:30:56.932 [ERROR] [org.openhab.core.model.script.B   ] - STARTED
2025-10-15 19:30:56.935 [ERROR] [org.openhab.core.model.script.D   ] - M 7
```
so the value 7 (last line/exrpession of the script) was received by the rule.

PR #2594, opened 2025-11-12
getting_started: Persistence can be configured over UI

dilyanpalauzov:
Since openHAB 4.0 Persistence and Transformation can be configured over Main Ui.

I do not dare to ask here if there are other things, which cannot be configured yet over UI, justifying the remaining part of the line, as [I will be invited to delete the question](https://github.com/openhab/openhab-docs/pull/2586#issuecomment-3427656253).

### DIFF
--- a/tutorials/getting_started/index.md
+++ b/tutorials/getting_started/index.md
@@ -59,7 +59,7 @@ Cons:
 
 - Can be slower to reach one's goal if you already know what you're doing and if there's a lot of bulk changes or duplicating required.
 - Harder to remove obsolete stuff.
-- Some things cannot be configured with the UI yet (e.g. persistence).
+- Some aspects cannot be configured with the UI yet.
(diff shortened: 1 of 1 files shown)


rkoshak:
There are enough add-ons that required manually editing files (e.g. Exec binding whitelist) the sentence should remain. But all the main OH concepts (Things, Items, Semantic Model, Transformations, Persistence, Rules, Sitemaps, Widgets) can now be edited through MainUI so removing the parenthetical makes sense without replacing it with a list of examples.

dilyanpalauzov:
I have changed the sentence from

> Some things cannot be configured with the UI yet.

to 

> Some aspects cannot be configured with the UI yet.

as “things“ can mean something containing channels and supplying items with data.  But this is not meant here.

florian-h05 (approved):
I actually think the UI can do everything file-based config can do.

PR #2599, opened 2025-11-23
[homie] Add Homie logo

ccutrer:
Since https://github.com/openhab/openhab-addons/pull/19700, it will be a separate listing in the addon store, and thus needs its own logo.


lsiepel:
LGTM

lolodomo:
To be merged also

PR #2603, opened 2025-11-26
Update description for 'group' parameter in config-xml

rkoshak:
Based on the discussion at https://github.com/openhab/openhab-webui/issues/3457 we determined the description of the `group` context on this doc was incorrect. This PR corrects the description.

### DIFF
--- a/developers/addons/config-xml.md
+++ b/developers/addons/config-xml.md
@@ -127,7 +127,7 @@ The following contexts require a specific format of the content:
   <tr><td>url</td><td>text</td><td>web url</td><td><code>&lt;input type="url"/></code></td></tr>
   <tr><td>item</td><td>text</td><td>Item name</td><td>custom input field</td></tr>
   <tr><td>thing</td><td>text</td><td>UID of a thing</td><td>custom input field</td></tr>
-  <tr><td>group</td><td>text</td><td>group name to which this parameter belongs</td><td></td></tr>
+  <tr><td>group</td><td>text</td><td>name of a Group Item</td><td></td></tr>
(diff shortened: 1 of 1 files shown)


Nadahar:
~~Sorry, I think this is wrong. This isn't about Item groups, but parameter groups,~~ unless I am completely confused <--

Nadahar:
To repeat what I've [stated before](https://github.com/openhab/openhab-webui/issues/3457#issuecomment-3544673995), the `group` field here must contain the "name"/identifier of the `parameterGroup` to which the parameter should belong.

The whole confusion here stems from the fact that there's both a "context" called "group", and there's a field called `group`, but these serve entirely different purposes.

rkoshak:
> To repeat what I've https://github.com/openhab/openhab-webui/issues/3457#issuecomment-3544673995, the group field here must contain the "name"/identifier of the parameterGroup to which the parameter should belong.

No, that's the `groupName` parameter. 

> parameter.groupName	Sets a group name for this parameter (optional).

I didn't change that.

This is changing the description on the  `context` "group` which we all agreed couldn't possibly be referencing the parameter group. It makes no sense to have a context for that.  And the JavaDoc for https://www.openhab.org/javadoc/latest/org/openhab/core/config/core/configdescriptionparameter#getContext() says that the "group" context is a Group's Item name, not the parameter group's name.

> group: A valid group item "name"

> The whole confusion here stems from the fact that there's both a "context" called "group", and there's a field called group, but these serve entirely different purposes.

The field is called `groupName`, or in this doc it's `parameter.groupName`. And I didn't change anything related to that. I only changed the description for the context.

Nadahar:
Soo sorry, my mistake. I thought that you changed the other one 😯

PR #2611, opened 2025-12-15
Update event source guidelines

ccutrer:
Avoid confusing syntax, rename package to bundle, and add and update examples to match actual implementation.

### DIFF
--- a/developers/utils/events.md
+++ b/developers/utils/events.md
@@ -43,9 +43,12 @@ This string type presentation is used by event subscribers for event subscriptio
 The event source identifies the sender.
 Not all senders will set the source.
 The source should be structured, consisting of a series of concatenated components (separated by `=>` to indicate delegation), and each component consists of a bundle name with an optional actor name (separated by `$`).
-Bundle names cannot contain either the delegation separator, or the actor separator.
+For example, given `org.openhab.ui.basic$default:03=>org.openhab.core.io.rest` we have two components separated by the `=>`.
+The first component has the bundle `org.openhab.ui.basic` and the actor `default:03`.
+The second component only has the bundle `org.openhab.core.io.rest`.
+Bundle names can contain neither the delegation separator, nor the actor separator.
 Actor names cannot contain the delegation separator.
-The sending bundle is meant to be the OSGi bundle name of the component that sent the event.
+The sending bundle is meant to be the OSGi bundle name of the component that sent the event, though this isn't a strict rule and may be violated in some cases, especially for something like `org.openhab.core` that has many different packages in it.
 For non-Java components generating a source (such as the Android and iOS apps), it should still be in Java package name format, as if the component were a Java package.
 The actor piece is an identifier of some sort that identifies the end user, device, rule, or other identifier of a specific instance that initiated the event, if one is relevant.
 If an event is passed through multiple components before being processed, it should preserve the original source, and append (separated by `=>`) another component for itself.
@@ -56,9 +59,9 @@ A few concrete examples:
 - `org.openhab.io.homekit$1467397f-c2e7-4b15-a7dc-315331ced2db` would mean the event was initiated from the HomeKit addon by a user identified by a UUID.
 - `org.openhab.ui.basic$default:03=>org.openhab.android$my-phone=>org.openhab.io.openhabcloud$user@gmail.com=>org.openhab.core.io.rest` would mean the event was initiated by using a sitemap in the Android openHAB app, and proxied through myopenhab.org using the user `user@gmail.com` (and finally through the REST API, but without a user).
 - `org.openhab.core.io.console` would mean the event was sent via the Karaf console.
-- `org.openhab.core.automation` would mean the event was triggered by a command sent in a rule action in most langauges.
+- `org.openhab.core.automation` would mean the event was triggered by a command or update sent in a rule action in most langauges. Some languages might provide a more detailed source.
 - `org.openhab.automation.jrubyscripting$rule:virtual_lighting:98` would mean the event was sent by an action written in JRuby using the helper library, with rule ID `virtual_lighting:98`.
-- `org.openhab.binding.mqtt$mqtt:thing:mything:mychannel` would mean the event was sent by a specific channel from the MQTT binding (updating a state).
+- `org.openhab.binding.mqtt$mqtt:thing:mything:mychannel` would mean the event was sent by the `mychannel` channel of the `mything` thing from the MQTT binding (updating a state).
 - `org.openhab.core.persistence` would mean the event was sent by the persistence engine (restoring the state).
 - `org.openhab.core.autoupdate` would mean the event was sent by the AutoUpdate manager (updating the state automatically).
 


ccutrer:
cc @florian-h05 @kaikreuzer

developers/utils/events.md
```diff
@@ -42,18 +42,24 @@ This string type presentation is used by event subscribers for event subscriptio
...
-- `org.openhab.android$my-phone=>org.openhab.io.openhabcloud$user@gmail.com` would mean the event was initiated in the Android openHAB app, and proxied through myopenhab.org using the user `user@gmail.com`.
-- `org.openhab.core.io.console=>org.openhab.core.automation$f8e461d0d9` would mean an event was sent via the Karaf console, which then triggered a rule with ID `f8e461d0d9`, triggering the current event.
-- `org.openhab.binding.mqtt$mqtt:thing:mything:mychannel` would mean an event was sent by a specific channel from the MQTT binding.
+- `org.openhab.ui.basic$default:03=>org.openhab.android$my-phone=>org.openhab.io.openhabcloud$user@gmail.com=>org.openhab.core.io.rest` would mean the event was initiated by using a sitemap in the Android openHAB app, and proxied through myopenhab.org using the user `user@gmail.com` (and finally through the REST API, but without a user).
+- `org.openhab.core.io.console` would mean the event was sent via the Karaf console.
+- `org.openhab.automation.jrubyscripting$rule:virtual_lighting:98` would mean the event was sent by a JRuby rule with ID `virtual_lighting:98`.
```
Nadahar: I'm not sure that this a good idea, what exactly constitutes a "JRuby rule"? A rule can have multiple conditions and actions, written in different languages. So, all rules should probably use a uniform "actor" part that doesn't try to differentiate on the language used.

The "bundle" rule means that all rules would have to use `org.openhab.core.automation`, since that's where `Rule` is defined, so there would be no clear way to identify that it actually originates from a rule.

ccutrer: A rule written with the JRuby helper library is able to populate the Rule ID. DSL rules have no way to do so, so are handled generically with the static source of `org.openhab.core.automation`. The source comes from https://github.com/openhab/openhab-jruby/blob/23a04794d81a90686badb3367a55a1b3c70fcdec/lib/openhab/core/items/item.rb#L644, thus the bundle `org.openhab.automation.jrubyscripting`, and not `org.openhab.core.automation`.

BUT... I'll also add an example for a DSL rule, and clarify wording around rule/action.

Nadahar: It's not that simple, the special case is those rules created by the helper library, not the DSL rules. A rule created in the UI, or by a YAML file, or any other generic way, will behave like "a DSL rule" - that also applies if the rule uses JRuby.

Only rules that implement `SimpleRule` can probably use this "custom source", and I'm not sure why one would even bother making them have a special source. What value does knowing this add to the recipient? A `SimpleRule` is, as the name suggests, a limited `Rule` implementation that can't do many things that other rules can, but has *one* good ability - the ability to share the context with the script that creates them.

ccutrer: Okay, let me rephrase - `DSL rules have no way to do so. Any other language may be able to do so, but to my knowledge has not done so as of writing.`

I'm not sure what you're talking about for `SimpleRule`? The source is sent by the BusEventActionService, which can be called by any sort of module (be it action, condition, whatever).

Knowing a more precise source of a command will be extremely useful for end users debugging their rules when looking at logs. It can also help for rule implementation preventing infinite loops - I have a few rules that check the source of a command, and do nothing if it's from a particular other rule.

ccutrer:
cc @rkoshak, as our de facto users' advocate :)

developers/utils/events.md
Nadahar: > Okay, let me rephrase - `DSL rules have no way to do so. Any other language may be able to do so, but to my knowledge has not done so as of writing.`
>
> I'm not sure what you're talking about for SimpleRule? The source is sent by the BusEventActionService, which can be called by any sort of module (be it action, condition, whatever).

I think you're missing my point - I'm just saying that framing it as if it's DSL that "uses `org.openhab.automation`" would be wrong. You know better than me how the JRuby helper lib works, I'm not even sure exactly under which circumstances these rule-created events occur, but are you saying that any event generated using the JRuby helper library will apply this "custom" source? Even if it's just one of several actions in a rule? I assume that it's also possible to send events from JRuby that doesn't use the function in the helper library, so what this source would actually identify is the use of the helper library, not the JRuby script engine.

> Knowing a more precise source of a command will be extremely useful for end users debugging their rules when looking at logs. It can also help for rule implementation preventing infinite loops - I have a few rules that check the source of a command, and do nothing if it's from a particular other rule.

I agree that knowing what rule that triggers it is useful, but I'd think that the rule UID is much more relevant for that than e.g. `org.openhab.automation.jrubyscripting`.

What about rules that use different languages or a language in combination with built-in conditions and actions. Should they also be considered e.g. "a JRuby rule"? I assume that it's the helper lib in use by the action that generates the event that will actually determine what form the source will have, I'm just asking if this is useful. I would think that, what would be relevant to know regarding source for a rule generated event is the fact that the source *is* a rule, and the rule UID.

Nadahar:
If you rename package to bundle, you should probably do the same here:

https://github.com/openhab/openhab-core/blame/c3deec6d2121ae1bc1a0f0f5a463e310397236d4/bundles/org.openhab.core/src/main/java/org/openhab/core/events/AbstractEvent.java#L101-L113

That said, using package would allow more detail in some situations, e.g. `org.openhab.core.item` instead of just `org.openhab.core`.

developers/utils/events.md
```diff
@@ -42,18 +42,30 @@ This string type presentation is used by event subscribers for event subscriptio
...
-- `org.openhab.binding.mqtt$mqtt:thing:mything:mychannel` would mean an event was sent by a specific channel from the MQTT binding.
+- `org.openhab.ui.basic$default:03=>org.openhab.android$my-phone=>org.openhab.io.openhabcloud$user@gmail.com=>org.openhab.core.io.rest` would mean the event was initiated by using a sitemap in the Android openHAB app, and proxied through myopenhab.org using the user `user@gmail.com` (and finally through the REST API, but without a user).
+- `org.openhab.core.io.console` would mean the event was sent via the Karaf console.
+- `org.openhab.core.automation` would mean the event was triggered by a command sent in a rule action in most langauges.
+- `org.openhab.automation.jrubyscripting$rule:virtual_lighting:98` would mean the event was sent by an action written in JRuby using the helper library, with rule ID `virtual_lighting:98`.
+- `org.openhab.binding.mqtt$mqtt:thing:mything:mychannel` would mean the event was sent by a specific channel from the MQTT binding (updating a state).
```
rkoshak: Instead of "by a specific channel" say "by the 'mychannel' channel of the 'mything' thing" or something like that. The more the explanation can point to specific parts of the example, the easier it is for readers to understand and apply.

developers/utils/events.md
```diff
@@ -42,18 +42,30 @@ This string type presentation is used by event subscribers for event subscriptio
 
 The event source identifies the sender.
 Not all senders will set the source.
-The source should be structured, and of the form `<sending package>[$<actor>][=><sending package[$<actor>]]+`
-The sending package is meant to be the Java package name of the component that sent the event.
+The source should be structured, consisting of a series of concatenated components (separated by `=>` to indicate delegation), and each component consists of a bundle name with an optional actor name (separated by `$`).
```
rkoshak, line 45: I know there are lots of examples later, but it might be worth adding a "for example" here.

> For example, given `org.openhab.ui.basic$default:03=>org.openhab.core.io.rest` we have two components separated by the "=>". The first component has the bundle "org.openhab.ui.basic" and the actor "default:03". The second component only has the bundle "org.openhab.core.io.rest". 

This part is mainly just making the mapping between the terms just defined to a concrete example which can help a lot with understanding.

developers/utils/events.md
rkoshak: > so what this source would actually identify is the use of the helper library, not the JRuby script engine.

I would assume it uses the BusEvent Actions. For example https://www.openhab.org/javadoc/latest/org/openhab/core/automation/module/script/action/busevent#postUpdate(org.openhab.core.items.Item,java.lang.String,java.lang.String). 

In that case, I imagine the source is what ever the user sets it to. As of right now, I suspect none of the other helper libraries or supported languages except for jRuby wrap this to supply the `source` yet so when calling `sendCommand` or `postUpdate` from these would not set the source at all. 

The end user has the option to supply a source if they choose (even in rules DSL) if they use the BusEvent Actions directly instead of through the helper libraries. 

And if I were to guess how it's implemented in jRuby, it should work as @ccutrer describes for jRuby Script Actions and Conditions. This is because when the user calls `postUpdate(foo)` (for example) that ultimately ends up at a call to `org.openhab.core.automation.module.script.action.BusEvent.postUpdate(itemName, foo, event.source+"=>org.openhab.automation.jrubyscripting$rule:virtual_lighting:98')` (I'm guessing at how to access the source that triggered the rule here). 

It will work slightly differently for JS text based rules than UI rules because for reasons I don't understand text file created rules in JS only knows the file name, not the ruleUID. But that's not yet implemented so not of concern here. I only bring it up to illustrate that how this is handled is largely going to be very specific to the individual automation add-ons implementations, specifically their helper libraries.

Nadahar:
I never got to that because of the discussion about the package/bundle name around rules, but except for the fact that I think there should be something that identifies it as a rule (aka `org.openhab.core.rule`) for all rules instead, I think the rest looks good. It's certainly much clearer and easier to understand 👍

developers/utils/events.md
```diff
@@ -42,18 +42,30 @@ This string type presentation is used by event subscribers for event subscriptio
...
+Bundle names cannot contain either the delegation separator, or the actor separator.
```
florian-h05:
```suggestion
Bundle names can contain neither the delegation separator, nor the actor separator.
```

Personally I think that's easier to read than double negation.

developers/utils/events.md
florian-h05: > It's not that simple, the special case is those rules created by the helper library, not the DSL rules. A rule created in the UI, or by a YAML file, or any other generic way, will behave like "a DSL rule" - that also applies if the rule uses JRuby.

JS Scripting actually sets the source based on ruleUID or filename, regardless whether you use file-based rules or ScriptAction/ScriptConditions. This is done by the helper library when calling `items.getItem(...).sendCommand(...)` or `items.getItem(...).postUpdate(...)`.
If I wanted to, I could implement that in the add-on itself, but that would be way more complicated.

developers/utils/events.md
```diff
@@ -42,18 +42,30 @@ This string type presentation is used by event subscribers for event subscriptio
...
-- `org.openhab.android$my-phone=>org.openhab.io.openhabcloud$user@gmail.com` would mean the event was initiated in the Android openHAB app, and proxied through myopenhab.org using the user `user@gmail.com`.
-- `org.openhab.core.io.console=>org.openhab.core.automation$f8e461d0d9` would mean an event was sent via the Karaf console, which then triggered a rule with ID `f8e461d0d9`, triggering the current event.
-- `org.openhab.binding.mqtt$mqtt:thing:mything:mychannel` would mean an event was sent by a specific channel from the MQTT binding.
+- `org.openhab.ui.basic$default:03=>org.openhab.android$my-phone=>org.openhab.io.openhabcloud$user@gmail.com=>org.openhab.core.io.rest` would mean the event was initiated by using a sitemap in the Android openHAB app, and proxied through myopenhab.org using the user `user@gmail.com` (and finally through the REST API, but without a user).
+- `org.openhab.core.io.console` would mean the event was sent via the Karaf console.
+- `org.openhab.core.automation` would mean the event was triggered by a command sent in a rule action in most langauges.
```
florian-h05:
```suggestion
- `org.openhab.core.automation` would mean the event was triggered by a command or update sent in a rule action. Some languages might provide a more detailed source.
- `org.openhab.automation.jsscripting$rule:MyRule` would mean the event was triggered by a command or update sent from a Script Action or Script Condition using JavaScript Scripting of a rule with the UID `MyRule`.
```

developers/utils/events.md
Nadahar: My point was more about whether there is any value in specifying which add-on/helper library that generated the event, instead of just the fact that the event originated from a rule with a specific rule UID.

But, this is probably not worth all the discussion, it's not that big a point. The most important thing is that the rule UID is there, and that there's a way to recognize that the source includes a rule UID. Most people will probably use only one scripting language, so it won't matter that much if the source identifiers look different from different rule sources.

florian-h05: > My point was more about whether there is any value in specifying which add-on/helper library that generated the event, instead of just the fact that the event originated from a rule with a specific rule UID.

I now understand your point. I think it makes sense to denote the language of the rule and not only the rule UID, and this makes it also consistent with logger names (at least for JS, which logs to `org.openhab.automation.jsscripting.file.<filename>` or `org.openhab.automation.jsscripting.rule.<ruleUID>`; and `org.openhab.automation.jsscripting$rule:ruleUID` etc. as source).

ccutrer: @rkoshak: yup, you pretty much nailed it, with the additional point that a user is still allowed to overwrite the source that is sent if they want to (just like if they were to use BusEvent themselves). And to @florian-h05: yup, same - the source the JRuby helper library set literally comes from the same methods that determine what logger name to use.

> I agree that knowing what rule that triggers it is useful, but I'd think that the rule UID is much more relevant for that than e.g. org.openhab.automation.jrubyscripting.

@Nadahar: I _think_ you're saying you don't care about a bundle name there, you only want the rule UID? The problem is the bundle name is essentially setting the namespace (or tagging it, as you alluded to earlier) to be able to interpret the actor. I think you're also trying to get to the point that you want to be able to extract the rule UID from a source without caring about which language it was written in. While I see how that would be nice, I don't think it's a critical or simple thing to do. It looks like JS and Ruby already use the same actor format. And BusEvent can't infer a rule UID itself, so it's always going to be dependent on a helper library providing it. And at least in file-based Ruby (and I assume JS ruby) it's possible to send commands and update items _outside_ the context of a rule execution. In that case, it ends up just using the filename (and without the `rule:` prefix). And Ruby does one more trick that if it happens to be a profile or script, it uses those as the prefix, not `rule`.

ccutrer:
I've accepted all specific suggestions, and also added a line about bundle name not necessarily being a hard and fast rule (AutoUpdateManager technically already violates it).

developers/utils/events.md
Nadahar: Yes, my idea was that it would be more uniform if all events originating from rules had a form similar to `org.openhab.core.rule$<ruleUid>`, but you're correct that it wouldn't work anyway, because the rule UID probably can't be inferred if not explicitly provided by the caller. Which makes "my idea" pointless - so please go ahead as planned 😉

PR #2623, opened 2025-12-27
Clarification: Rules DSL derives from XBase and is similar to Xtend

dilyanpalauzov:
* update hyperlinks to Xbase Language Reference and Xtend Documentation

Before this change there were contradicting statements, whether DSL is based on Xbase or on Xtend.


florian-h05:
@openhab/core-maintainers Please confirm.

PR #2628, opened 2025-12-30
Document how to make thread dumps

Nadahar:
Time and time again, I need to tell people how to make thread dumps, and I finally decided to document it to not have to look up how to every time.

Thread dumps are very useful for figuring out deadlocks or why a system slows down to a crawl.

Feel free to point out formatting or spelling issues.

### DIFF
--- a/administration/runtime.md
+++ b/administration/runtime.md
@@ -183,7 +183,7 @@ Created dump zip: 2025-08-08_171434.zip
 You will find this ZIP in the `$OPENHAB_USERDATA` folder, usually `/var/lib/openhab/` on Linux systems.
 
 ::: tip Note
-The thread dumps created from Karaf lacks some important information for troubleshooting blocked threads.
+The thread dumps created from Karaf lack some important information for troubleshooting blocked threads.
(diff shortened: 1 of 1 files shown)


administration/runtime.md
```diff
@@ -182,4 +182,72 @@ Created dump zip: 2025-08-08_171434.zip
 
 You will find this ZIP in the `$OPENHAB_USERDATA` folder, usually `/var/lib/openhab/` on Linux systems.
 
+::: tip Note
+The thread dumps created from Karaf lacks some important information for troubleshooting blocked threads.
```
andrewfg: `lacks` => `lack`

andrewfg (changes requested):
Just one small comment below..

Also wondering if we should document https://github.com/openhab/openhab-core/pull/4948 here?

Otherwise LGTM.

Nadahar:
> Also wondering if we should document [openhab/openhab-core#4948](https://github.com/openhab/openhab-core/pull/4948) here?

That might not be a bad idea, but my immediate need is to have somewhere to reference every time I need to ask people to make a thread dump.

PR #2633, opened 2026-01-04
Improve transformation docs

florian-h05:


configuration/transformations.md
```diff
@@ -40,36 +39,42 @@ Transformation files need to be placed in the directory `$OPENHAB_CONF/transform
     ```
 
     Usage of Transformations in the [label parameter of Sitemap elements]({{base}}/ui/sitemaps.html#element-type-text) works the same way.
+
+    When creating Item's through the UI, the content of the `[]` brackets has to be put into the `pattern` field of the State Description Item metadata.
```
jimtng:
```suggestion
    When creating an Item through the UI, the content of the `[]` brackets has to be put into the `pattern` field of the State Description Item metadata.
```

configuration/transformations.md
```diff
@@ -40,36 +39,42 @@ Transformation files need to be placed in the directory `$OPENHAB_CONF/transform
...
+    Transformations can sometimes be used directly in [binding add-ons](/addons/#binding).
+    For example, transforming an openHAB ON command into "action=powerup" for sending to a device.
+    Bindings can either implement their own way of handling transformations or use the Channel Transformation provided by the framework.
+    If, and how, transformation can be used will be described the individual binding documentation.
 
 ## Script Transformation
 
-The script transformation is available from the framework and needs no additional installation.
-It allows transforming values using any of the available scripting languages in openHAB (JSR-223 or DSL).
+The script transformation is provided by the framework and needs no additional installation of a [transformation add-on](/addons/#transform).
```
jimtng: This can be confusing because apart from rulesdsl, you do need to install jsscripting, jruby, etc as the underlying scripting engine.

configuration/transformations.md
```diff
@@ -40,36 +39,42 @@ Transformation files need to be placed in the directory `$OPENHAB_CONF/transform
...
+It allows transforming values using any of the available scripting languages in openHAB (Rules DSL and those provided by [automation add-ons](/addons/#automation)).
```
jimtng: So this seems contradictive. Perhaps rephrase both sentences somehow? Because here, we probably should tell them that they need to install one of the scripting language addons.

florian-h05:
@jimtng Better now?

jimtng (approved):
perfect, thanks!

PR #2637, opened 2026-01-07
DSL examples: remove optional brackets

dilyanpalauzov:
https://eclipse.dev/Xtext/documentation/305_xbase.html#xbase-expressions-lambda contains „in Xbase parentheses are optional for method calls“ and https://eclipse.dev/Xtext/xtend/documentation/203_xtend_expressions.html#feature-calls repeats “A simple name can refer to a local field, variable or parameter. In addition it can point to a method with zero arguments, since empty parentheses are optional.”

Note that the change in configuration/actions.md is spelled for “```php`, and PHP is shown in the final page, while the changes in configuration/rules-dsl.md are after ”```java” or after “```javascript” and as such are displayed explicitly at https://next.openhab.org/docs/configuration/rules-dsl.html.  However in practice the snippets are for DSL, as the name of the changed file implies.

### DIFF
--- a/configuration/actions.md
+++ b/configuration/actions.md
@@ -167,7 +167,7 @@ then
         }
     } else {
         logInfo("rules", "Timer canceled")
-        myTimer?.cancel()
+        myTimer?.cancel
         myTimer = null
     }
 end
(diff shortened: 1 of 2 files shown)


stefan-hoehn:
@openhab/core-maintainers can anyone check if this is correct?

florian-h05 (approved):

PR #2640, opened 2026-01-11
Update windows.md

spacemanspiff2007:
As requested by @florian-h05 .
The information is incorrect and should not be there.

### DIFF
--- a/installation/windows.md
+++ b/installation/windows.md
@@ -199,12 +199,6 @@ You can:
 
 ## Maintenance
 
-### Backup
-
-Make sure that you make regular backups of the **conf** and **userdata** folders, you can zip and unzip these folders too and from openHAB installations (even across most versions).
-When you have a setup that you are happy with, it would be a good idea to make a backup of the whole `C:\openHAB` folder.
-Which can be used any time after a failure.
-
 ### Updating the openHAB Runtime
 
 To start the update process, run PowerShell as an administrator and use the following commands, while replacing `x.x.x` with the wanted openHAB version.


florian-h05 (commented):
I would rather remove the sentence even across openHAB versions and add in „across the same openHAB version“.

spacemanspiff2007:
> I would rather remove the sentence even across openHAB versions and add in „across the same openHAB version“.

Are you 100% sure that restoring an old version version of the `cache` or `tmp` folder does not cause any error or issues?
Wouldn't it be better to tell the user to exclude the cache directories?

florian-h05:
I think the cache and tmp folder are only included if the full option is set when creating a backup.
But to be on the safe side, we should recommend to remote the cache and tmp directories, WDYT?

stefan-hoehn (approved):

stefan-hoehn:
Sorry, I approved but only then saw @florian-h05 's comment. @spacemanspiff2007 Anything you want to change?

spacemanspiff2007:
I think the best would be to add an information to exclude the following folders/files from a backup
- `cache`
- `config` 
- `tmp`
- `logs`
- `etc` except for `etc/log4j2.xml`
- `Californium3.properties`
- `marketplace` <-- Not sure about that one

rkoshak:
> config

This absolutely must be part of the backup. That's where all the settings are saved made through the REST API which do not go in the JSONDB. This includes which add-ons are installed, add-on specific configs, regional settings, ephemeris, etc. 

> etc except for etc/log4j2.xml

There are more files in etc that the user may have changed. For example `keys.properties` if they changed the password for the karaf console. The `keystore`, `host.key`, `host.key.pub` etc are all files the end user may have modified and which should be included in a backup. 

> marketplace <-- Not sure about that one

That should be excluded.

florian-h05 (changes requested):
I did not mean to remove this section, but correct it => it does not work across openHAB versions.

PR #2644, opened 2026-01-12
Modify openHAB service setup instructions for manual Linux installations

sihui62:
Updated the instructions for setting up openHAB as a systemd service on manual Linux installations to include the newly introduced linux.parameters file for overwriting environmental variables:

https://github.com/openhab/openhab-linuxpkg/pull/233

Tested here:
https://community.openhab.org/t/etc-openhab-linux-parameters/163246/12

### DIFF
--- a/installation/linux.md
+++ b/installation/linux.md
@@ -557,7 +557,7 @@ A cleaner approach is to create a Linux service.
 The following instructions are intended for a Linux init system based on **systemd** (e.g. Debian 8 / Ubuntu 15.x and newer).
 This will allow you to register openHAB as a service, so that it runs at startup and automatically restarts if openHAB crashes.
 The service will be running with the privileges of the user "openhab" and expects the openHAB files under `/opt/openhab`.
-Additionally, if you need to modify environment variables, create the file `/opt/openhab/conf/misc/linux.parameters` with your changes or additions.
+Additionally, if you need to modify environment variables, create the file `/opt/openhab/conf/linux.parameters` with your changes or additions.
 
 Create the file `/usr/lib/systemd/system/openhab.service` with the following content:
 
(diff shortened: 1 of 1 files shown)


installation/linux.md
```diff
@@ -557,6 +557,7 @@ A cleaner approach is to create a Linux service.
 The following instructions are intended for a Linux init system based on **systemd** (e.g. Debian 8 / Ubuntu 15.x and newer).
 This will allow you to register openHAB as a service, so that it runs at startup and automatically restarts if openHAB crashes.
 The service will be running with the privileges of the user "openhab" and expects the openHAB files under `/opt/openhab`.
+In addition if you need to add or change environmental variables, create the file `/opt/openhab/conf/misc/linux.parameters` with your changes or additions.
```
stefan-hoehn: "Additionally, if you need to modify environment variables, create the file..."

installation/linux.md
```diff
@@ -557,6 +557,7 @@ A cleaner approach is to create a Linux service.
 The following instructions are intended for a Linux init system based on **systemd** (e.g. Debian 8 / Ubuntu 15.x and newer).
 This will allow you to register openHAB as a service, so that it runs at startup and automatically restarts if openHAB crashes.
 The service will be running with the privileges of the user "openhab" and expects the openHAB files under `/opt/openhab`.
+Additionally, if you need to modify environment variables, create the file `/opt/openhab/conf/misc/linux.parameters` with your changes or additions.
```
florian-h05: https://github.com/openhab/openhab-linuxpkg/pull/233 uses `/etc/openhab/linux.parameters` as path.
So this one here should rather be `/opt/openhab/conf/linux.parameters` IMO.

installation/linux.md
```diff
@@ -573,7 +574,7 @@ User=openhab
 Group=openhab
 
 WorkingDirectory=/opt/openhab
-#EnvironmentFile=-/etc/default/openhab
+EnvironmentFile=-/opt/openhab/conf/misc/linux.parameters
```
florian-h05: See above.

sihui62: I'm fine with that, as it works in both locations. The reason for /misc was:

https://community.openhab.org/t/etc-openhab-linux-parameters/163246/12?u=sihui

> For me personally I choose another location for the linux.parameters file, because in the root directory of `conf` it looks strange as it is the only file. I have put it under `.../conf/misc`.



Screenshot 2026-03-17 141502

florian-h05: I understand that idea, but I prefer keeping the file out of folders that openHAB itself is looking for files.

PR #2676, opened 2026-03-14
Update the iOS docs

### LINKED ISSUE #2675: iOS App page outdated, appstore link missing
The iOS app page was last updated by @digitaldan , seemingly before the V2 app was released. It should be updated with a current link and a description of features of the V2 app. The V1 app should only be mentioned as backwards-compatible version.

I can submit a PR for the update, but maybe someone else is faster than me ;)

digitaldan:
Closes #2675


florian-h05:
Can you please fix the markdownlint issues?

And another question: Is this copied from the iOS app repo or only maintained here? I'm asking because if its copied, I would setup the automatic docs integration similarly to Alexa docs, Garmin app, etc.

florian-h05:
@stefan-hoehn It seems this is copied over from the iOS repo, we can merge it for now but I would prefer to implement automatic doc integration for this as I did for Garmin and SailfishOS.

florian-h05 (approved):

PR #2693, opened 2026-04-03
Linkplay logo

digitaldan:
see https://github.com/openhab/openhab-addons/pull/19554


stefan-hoehn:
@digitaldan The image is too big (it should be less than 30kb). I tweaked and reduced the size. Can you use the attached one instead?
linkplay

florian-h05 (approved):

PR #2696, opened 2026-04-06
[energyforecast] Add logo for new binding

weymann:
Add logo for new introduced binding [energyforecast](https://github.com/openhab/openhab-addons/pull/20325)


stefan-hoehn:
We usually prefer SVGs but I couldn't find this logo as an SVG either, so let's move forward with the png.

stefan-hoehn (approved):

stefan-hoehn (changes requested):
Now that I look again on the png it is actually rather big. We usually want them around the size less the 30kB (see /developers/addons/index.md)

```
In case you don't have an SVG version of the logo, you can instead provide a PNG version, these are the requirements for PNG logos:

- PNG (transparency is preferred)
- 512x512 pixels or smaller in one dimension, if it's not a square logo
- Less than 30kB
```

weymann:
> Now that I look again on the png it is actually rather big. We usually want them around the size less the 30kB (see /developers/addons/index.md)
> 
> ```
> In case you don't have an SVG version of the logo, you can instead provide a PNG version, these are the requirements for PNG logos:
> 
> - PNG (transparency is preferred)
> - 512x512 pixels or smaller in one dimension, if it's not a square logo
> - Less than 30kB
> ```

Ok,
tried svg converter and IMHO delivered good result. Dimesions 512x512 are fitting and size is now ~19Kb

stefan-hoehn (approved):
@weymann Great, that is even better 👏

PR #2699, opened 2026-04-11
More Xbase style

dilyanpalauzov:


configuration/actions.md
```diff
@@ -198,7 +198,7 @@ For example:
 ```java
 var thingStatusInfo = getThingStatusInfo("zwave:device:c5155aa4:node2")
 
-if ((thingStatusInfo !== null) && (thingStatusInfo.getStatus().toString() == "ONLINE")) {
+if (thingStatusInfo?.status?.toString == "ONLINE") {
```
stefan-hoehn, line 201: I am not aware that the null-safe navigation operator is available in Java, is it? So, either the code here or ```java is wrong.

configuration/actions.md
```diff
@@ -109,8 +109,8 @@ All HTTP Actions can have a last `timeout` parameter added in ms. eg. `sendHttpP
 
 For example:
 
-```javascript
-val headers = newHashMap("Cache-control" -> "no-cache")
+```java
+val headers = #{"Cache-control" -> "no-cache"}
```
stefan-hoehn, line 113: I am not aware that #{...} is either java or javascript.

stefan-hoehn (changes requested):

configuration/actions.md
dilyanpalauzov: This is DSL Rules/Xbase/DSL Script syntax.  `java` is used to force highlighting for DSL/XBase snippets, on many other places the same approach is used.

configuration/actions.md
dilyanpalauzov: `val` is neither Java nor JavaScript, too.  `newHashMap()` without terminating semi-colon is neither Java, nor JavaScript, too.

configuration/actions.md
rkoshak: @dilyanpalauzov is corerct. There isn't an ```xbase tag available to do syntax highlighting. The tag that comes closest to doing it well is ```java for Rules DSL blocks. I think this is handled inconsistently throughout the docs but we should standardize on using one tag for all Rules DSL and ```java is the best candidate. The only other alternative is to extend the upstream tools we use to add xtend as a formatting tag.

PR #2707, opened 2026-04-25
DSL Rules: Thing based trigger always need quotes

dilyanpalauzov:
Due the way thing UIDs are constructed:  `<binding>`:`<bridge>`:…  the name always contains colon. So it must be quoted unconditionally.


lolodomo (approved):
LGTM

florian-h05 (approved):

PR #2710, opened 2026-05-01
WebSocket API: Document Item type for PCM Audio WebSocket

florian-h05:


florian-h05:
For the listeningItem, we have this config description declaring it has to be a Switch: https://github.com/openhab/openhab-core/blob/392241a02989634a327f0ba92449475073a563d1/bundles/org.openhab.core.voice/src/main/resources/OH-INF/config/voice.xml#L32

For the locationItem, String is logical.

PR #2711, opened 2026-05-02
Update documentation for sitemap Group element behavior

mherwege:
Clarify behavior of Group elements with and without child widgets. Group elements do support child widgets.


lolodomo (approved):
LGTM

florian-h05 (approved):

PR #2714, opened 2026-05-05
Mention ability to define several sitemaps in a file

lolodomo:
Mention the new .sitemaps extension.
Remove the old constraint between sitemaop name and sitemap file name.

### DIFF
--- a/ui/sitemaps.md
+++ b/ui/sitemaps.md
@@ -18,7 +18,8 @@ This page is structured as follows:
 
 [[toc]]
 
-Sitemaps are text files with the `.sitemap` extension, and are stored in the `$OPENHAB_CONF/sitemaps` directory.
+Sitemaps are text files with the `.sitemap` or `.sitemaps` extension, and are stored in the `$OPENHAB_CONF/sitemaps` directory.
+They can contain any number of sitemap definitions since version 5.2.
 Sitemaps follow the syntax described in this article.
 
 For easy editing of sitemap definition files, we suggest to use one of the [openHAB supporting editors]({{base}}/configuration/editors.html).
@@ -111,13 +112,13 @@ The `sitemap` element is **mandatory** in a Sitemap definition.
 This element shall be the first line in the sitemap file, and the following code block comprises the entire Sitemap definition.
 
 ```java
-sitemap <sitemapname> label="<title of the main screen>" {
+sitemap <sitemapname> [label="<title of the main screen>"] {
     [all sitemap elements]
 }
 ```
 
-- `sitemapname` shall always be equal to the Sitemaps file name, e.g. the `sitemapname` in a sitemap file named `demo.sitemap` must be "demo"
-- `label` is free text and will be shown as the title of the main screen.
+- `sitemapname` is the identifier of the Sitemap definition.
+- `label` is an optional free text and will be shown as the title of the main screen.
 
 (Note that the element `sitemap` is written with a lower case "s".)
 
--- a/ui/sitemaps.md
+++ b/ui/sitemaps.md
@@ -19,7 +19,7 @@ This page is structured as follows:
 [[toc]]
 
 Sitemaps are text files with the `.sitemap` or `.sitemaps` extension, and are stored in the `$OPENHAB_CONF/sitemaps` directory.
-They can contain any number of sitemap definitions since version 5.2.
+They can contain any number of sitemap definitions since openHAB 5.2.
 Sitemaps follow the syntax described in this article.
 
 For easy editing of sitemap definition files, we suggest to use one of the [openHAB supporting editors]({{base}}/configuration/editors.html).


ui/sitemaps.md
```diff
@@ -116,7 +117,7 @@ sitemap <sitemapname> label="<title of the main screen>" {
 }
 ```
 
-- `sitemapname` shall always be equal to the Sitemaps file name, e.g. the `sitemapname` in a sitemap file named `demo.sitemap` must be "demo"
+- `sitemapname` is the identifier of the Sitemap definition.
```
dilyanpalauzov, line 120: The next line should suggest, that `label=` is optional.  This can be expressed by changing line 114 , as is done on line 216:
```diff
- sitemap <sitemapname> label="<title of the main screen>" {
+ sitemap <sitemapname> [label="<title of the main screen>"] {
```

lolodomo: Done

ui/sitemaps.md
```diff
@@ -111,13 +112,13 @@ The `sitemap` element is **mandatory** in a Sitemap definition.
...
-sitemap <sitemapname> label="<title of the main screen>" {
+sitemap <sitemapname> [label="<title of the main screen>"] {
     [all sitemap elements]
 }
 ```
 
-- `sitemapname` shall always be equal to the Sitemaps file name, e.g. the `sitemapname` in a sitemap file named `demo.sitemap` must be "demo"
-- `label` is free text and will be shown as the title of the main screen.
+- `sitemapname` is the identifier of the Sitemap definition.
+- `label` is an optional free text and will be shown as the title of the main screen.
```
mherwege, line 121: You can also define an optional icon for sitemaps. I use it in the sitemap list in the UI sitemap list page. I don’t know it it used by any of the UIs. Should we add that now as well?

lolodomo: The icon is not used by Basic UI. I have no idea what format is even expected for this icon.

@openhab/android-maintainers @openhab/ios-maintainers : is this icon used in Android or iOS app ?
I suggest to document it in a separate PR and only if it used by one sitemap UI.

maniac103: It isn't used in the Android app.

mueller-ma: AFAIK the Sitemap icon is displayed in the app drawer on Android: https://github.com/openhab/openhab-android/blob/main/mobile/src/main/java/org/openhab/habdroid/ui/MainActivity.kt#L1161

dilyanpalauzov: I tried it: in openHAB-Android the sitemap icon is utilized, here for `sitemap ab label="A2" icon="colorpicker" {`:

image

maniac103: > AFAIK the Sitemap icon is displayed in the app drawer on Android:

Thanks for that correction, I completely forgot about that.
FWIW, what we're expecting there is the same format as for - say - sitemap widget icons.

lolodomo: Control has to be added to sitemap providers and then documentation could then be charged.

mherwege: From what sitemap provider is it missing? The only thing I am aware off is that the DSL provider does not use the icon syntax, but a string for it, but that should still work. It can be improved though.
BasicUI might be enhanced as well to show the icon on its sitemap selection page. But these are all optional enhancements, not a dependency.
FWIW, in my work to be able to link sitemaps (https://github.com/openhab/openhab-core/pull/5543),I also use this icon.

lolodomo: DSL and YAML providers should include a check of the proper icon syntax, the same as for the widget icon.

ui/sitemaps.md
```diff
@@ -18,7 +18,8 @@ This page is structured as follows:
 
 [[toc]]
 
-Sitemaps are text files with the `.sitemap` extension, and are stored in the `$OPENHAB_CONF/sitemaps` directory.
+Sitemaps are text files with the `.sitemap` or `.sitemaps` extension, and are stored in the `$OPENHAB_CONF/sitemaps` directory.
+They can contain one or several sitemap definitions.
```
dilyanpalauzov: I think this sentence `They can contain one or several sitemap definitions.` is not necessary.  `.items` files can also contain zero or more items, and this flexibility is nowhere stated.

dilyanpalauzov: Strictly speaking the sentence is also not correct.  In openHAB 5.1.4  when I create an empty `.sitemap` file the system logs `DSL model 'a.sitemap' is either empty or cannot be parsed correctly!`  With openHAB 5.2.0 build 5369 there is no such warning, so apparetly the `.sitemap` files can also contain zero or more sitemap definitions.

mherwege: It used to be one and exactly one. We changed it to allow any number of sitemaps. Like you can have an empty .items file, you can now also have an empty .sitemap(s) file. It doesn’t make much sense but it is possible.
I think it is good to be explicit for sitemaps that it can be multiple. It was always 1 single one since OH 1.0, so being explicit about it to make the point is worth it.

lolodomo: I will adjust the sentence as it is true that it can now contain 0 sitemap.
Similar to .items and .things and .rules files.

dilyanpalauzov: The text was underspecified.  Like it was not mentioned that `.persistence` files must contain `strategy{}` even if contains no definitions.  And likewise for DSL Rules it was not stated, that global variables cannot use previously defined global variables.  DSL Scripts have not documented, that from Xbase “Collection literals” could not be used in openHAB, but now there is no point to writing explicitly, that collection literals can be used.

These restrictions were not documented, so it is strange to write now down, that the restrictions do not exist.  I just see no need to spell explicitly that a restriction does not exist, if the assumption is, that the restriction (arbitrary number of sitemap defintions in a file) does not exist. 

Users, which have one sitemap definition per file, do not have to change anything, or to worry.

dilyanpalauzov: Irrespective of this change, the release notes for 5.2. could state, that `.sitemap` files can contain zero or more sitemap definitions.

dilyanpalauzov:
> They can now contain any number of sitemap definitions.

As these are not release notes, the word “now” should not be used.  I have been told that the documentation of openHAB is not supposed to describe how thing were in older versions, just describe the current version : https://github.com/openhab/openhab-docs/pull/2573#issuecomment-3461788572 .

Moreover at https://github.com/openhab/openhab-docs/pull/2567/ for the "Provider preset” I wrote “The preset is available since openHAB 5.0.”.  That last sentence was removed.

My understanding is that the openHAB-documentation refers only to the current⇔upcoming version, and it does not make statements about outdated, previous versions. 

That said, there is no point to emphasize here, that in the past one sitemap could be included per `.sitemap` file.

In addition, users read this sitemap.md file once, when they create a sitemap.  Users are not supposed to reread the file to find out, that things have changed.  This is to be included in the Release Notes, if it is considered as important enough the fact that a single sitemap file can from 5.2. on contain many sitemaps.

dilyanpalauzov:
Instead of `now` the text should be `since version 5.2`, so that users can understand the meaning, even if they read this in 10 years.  This does not change my opinion, that for consistency the sentence should disapear.

ui/sitemaps.md
```diff
@@ -18,7 +18,8 @@ This page is structured as follows:
 
 [[toc]]
 
-Sitemaps are text files with the `.sitemap` extension, and are stored in the `$OPENHAB_CONF/sitemaps` directory.
+Sitemaps are text files with the `.sitemap` or `.sitemaps` extension, and are stored in the `$OPENHAB_CONF/sitemaps` directory.
+They can contain any number of sitemap definitions since version 5.2.
```
florian-h05:
```suggestion
They can contain any number of sitemap definitions since openHAB 5.2.
```

florian-h05 (approved):

PR #2715, opened 2026-05-05
Enhance time trigger for DSL rule

lolodomo:


dilyanpalauzov:
It should be demonstrated, that hours must always be two digits: `08:00`.

lolodomo:
Good idea

lolodomo:
> It should be demonstrated, that hours must always be two digits: `08:00`.

This will be no more the case.
So I should probably change again the example.

configuration/rules-dsl.md
```diff
@@ -173,6 +173,7 @@ You can either use some pre-defined expressions for timers or use a [cron expres
 ```java
 Time is midnight
 Time is noon
+Time is 8:00
```
florian-h05, line 176:
```suggestion
Time is 08:00
```

??

florian-h05 (changes requested):

configuration/rules-dsl.md
lolodomo: No, it is deliberate, to show that this syntax is accepted.

florian-h05: Ah okay, I now re-read the discussion here. My fault.

florian-h05 (approved):

PR #2718, opened 2026-05-08
Move documentation into src and make docs processing a single source of truth

jimtng:
Move repository documentation into a new src/ layout and add tooling to support the migration and source-driven docs pipeline. Centralize docs-processing logic so other repos (notably openhab-website) can source the processing code from openhab-docs, removing duplicated processing logic and ensuring the website build always uses the latest processing logic.

Highlights:

- Moved ruby scripts into `scripts/`
- Documents moved: bulk relocation of documentation into src to match the exact final structure of the `docs/` directory.
- No more manually adjusting the ruby script for any modifications in the structure. Everything in src/ will be processed and deployed automatically.
- Removed some old directories / pages
- Website integration: [openhab-website clones openhab-docs](https://github.com/openhab/openhab-website/pull/560) during preprocessing and loads docs-processing ruby code from that clone, eliminating duplication, reducing maintenance, and creating a single source of truth for docs processing.


jimtng:
@florian-h05 this is what we discussed earlier about deduplicating the processing scripts between website and docs. Also see its counterpart in openhab/openhab-website#560

jimtng:
The web site build that relies on this new PR is green now! This PR must be merged after being rebased on the latest main so that we can pick up and migrate the latest documents into `src/`

jimtng:
I won't have a lot of availability in the next month. Switching this to draft for now.

florian-h05 (commented):
Overall LGTM, many thanks!
Let's tackle this now, right after the 5.2 release where we don't expect a lot of doc contributions.

FYI, we also need to update the other GHA workflows and the Maven/Groovy workflow in the final branch. Probably this could be migrated to Ruby as well, would be nice to have a single pipeline in place. Will have a look at it.

jimtng:
> FYI, we also need to update the other GHA workflows and the Maven/Groovy workflow in the final branch.

Can you point me to what you're referring to (url links)?

florian-h05:
Not sure what to do with the fetch_docs_* workdlows: https://github.com/openhab/openhab-docs/tree/main/.github/workflows

They update docs in the final-stable branch upon a release of e.g. openHABian - since the changes here only apply to main and final, these workflows should probably be kept unmodified until openHAB 5.3 release.

What has to be adjusted however is this:

- https://github.com/openhab/openhab-docs/blob/final/pom.xml
- https://github.com/openhab/openhab-docs/blob/final/process_addons.groovy
- https://github.com/openhab/openhab-docs/blob/final/process_thing_types.groovy
- https://github.com/openhab/openhab-docs/blob/final/update-external-resources.sh

It is only in the final-* branches and used by our Jenkins CI to integrate the add-ons docs into the final docs that are then picked up by openhab-website.

florian-h05:
@jimtng Just FYI, I am currently working on a consolidation of the whole fetch external docs pipeline into Ruby scripts (or more exactly, Antigravity is working on it).

scripts/lib/add_placeholders.rb
```diff
@@ -36,6 +36,17 @@ def add_placeholder_pages
...
 
+    puts "🔍 Checking target_file: #{target_file} (absolute: #{target_file.expand_path}) - exist? #{target_file.exist?}"
+
+    # If target file exists and is not a placeholder, skip it
+    if target_file.exist?
+      content = target_file.read
+      unless content.include?("migrated from another repository")
+        puts "  ⏭️ Skipping placeholder for #{path} (actual content present)"
+        next
+      end
```
jimtng: We might as well just skip if a file exists at all?

scripts/lib/add_placeholders.rb
```diff
@@ -8,8 +8,8 @@ def add_placeholder_pages
   puts "➡️ Adding placeholder pages for preview"
 
   pages = [
-    "addons/integrations/homekit",
-    "addons/integrations/openhabcloud",
+    "docs/addons/integrations/homekit",
+    "docs/addons/integrations/openhabcloud",
```
jimtng: There isn't actual src/docs/addons/xxx. These (and the sidebar above) are linked / referring to the actual addon documentations, unless you're planning to duplicate the readme of homekit and openhabcloud into src/docs/addons/ ?

florian-h05: Good catch, fill fix that :+1:

scripts/fetch-external-content.rb
```diff
@@ -0,0 +1,226 @@
...
+require "fileutils"
+
+# Ensure we can load files from scripts/lib
+$LOAD_PATH.unshift(File.join(File.dirname(__FILE__), "lib"))
+
+require "repo_manager"
+require "jfrog_fetcher"
+require "addon_processor"
+require "thing_types_processor"
+require "process_utils"
```
jimtng:
```suggestion
require_relative "lib/repo_manager"
require_relative "lib/jfrog_fetcher"
require_relative "lib/addon_processor"
require_relative "lib/thing_types_processor"
require_relative "lib/process_utils"
```

florian-h05 (approved):
LGTM, thanks -- I've moved the RFC of the fetch external content into a dedicated branch, will open a PR with that.

PR #2728, opened 2026-06-12
Rules DSL: Document new commands

Nadahar:
This documents the changes that were introduced in https://github.com/openhab/openhab-core/pull/5481.

PR #2739, opened 2026-07-02
Add sidebar menu depth

Nadahar:
This has been split from #2728. It adds extra depth to the sidebar menu, making it much easier to navigate long pages. It does so by using heading levels to create "chapters".

Apparently, this change must be done somewhere else as well, so I'm just submitting it here as a starting point.


florian-h05 (approved):

florian-h05:
> Apparently, this change must be done somewhere else as well, so I'm just submitting it here as a starting point.

I can submit it at the openhab-website repo.

PR #2749, opened 2026-07-06
Fetch external documentation contents through a Ruby pipeline

florian-h05:
This aims at replacing the combination of Maven plugins and Groovy scripts in the final branch at fetching and integrating external contents (e.g., openhab-addons) into the documentation.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,7 +19,8 @@ project.pbxproj
 .vscode
 
 # Built docs
+addons
 docs
-addons/integrations
 vuepress
+
 .DS_Store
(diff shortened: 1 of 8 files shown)


.gitignore
```diff
@@ -19,7 +19,9 @@ project.pbxproj
 .vscode
 
 # Built docs
+.vuepress/thing-types.json
+addons
```
Hilbrand, line 23: This line (`addons`) excludes the `images/addons` folder. I don't think that should be done?

florian-h05: Just pushed a fix :+1:

PR #2757, opened 2026-07-15
fetch-external-content: Fetch bundled openhab-js type definitions from GH release

florian-h05:
This fetches the bundled openhab-js type definitions `openhab.d.ts` from the GitHub release assets according to the ohjs version currently used in JS Scripting.
`openhab.d.ts` is then available on the website and the openhab-docs final branch. It can be used to provide improve LLM code generation.

### DIFF
--- a/scripts/fetch-external-content.rb
+++ b/scripts/fetch-external-content.rb
@@ -93,6 +93,27 @@ def update_external_repositories
   end
 end
 
+def fetch_openhab_js_types
+  pom_path = File.join(RESOURCE_FOLDER, "openhab-addons/bundles/org.openhab.automation.jsscripting/pom.xml")
+  unless File.exist?(pom_path)
+    warn "  ⚠️ JS Scripting POM not found at #{pom_path}"
+    return
+  end
+
+  pom_content = File.read(pom_path)
+  version_match = pom_content.match(/<ohjs.version>(.*?)<\/ohjs.version>/)
+  unless version_match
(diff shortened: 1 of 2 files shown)


florian-h05:
FYI @digitaldan

digitaldan:
Thanks @florian-h05 , question however, we do something similar for the MainUI, publish the widget definition file which the MCP service reads and caches form the users live system, this decouples the the MCP service from the version of the MainUI, so if you install an upgraded version of the MainUI (or an updated version of the MCP binding), it always has the latest  definitions that the user is actually running.   I was hoping to do something similar here, especially since not only can the JSScripting binding be upgraded, and MCP as well,  but a user might install a newer version of openhab-js, which is probably more common, so multiple ways there could be a difference.

Worst case we could always just default to stable/next versions of the type def from docs, and just assume it's "close enough" .  I was also thinking if we did that, we would do something similar to this PR if we did not want to modify our docs just to support it (i guess it does make it available to other agents looking at our docs, which reminds me, i have been seeing more and more docs have an alternative markdown output just for agents, maybe something we look at).

i'm not sure if you had thought about if this would be pulled at build time or run time by the MCP binding?

florian-h05:
We could easily build and bundle it with the JS Scripting add-on at JS Scripting build time.
I guess over the bundle context or something like that you could load resources of the JS Scripting bundle from within the MCP add-on. 

Wrt to manually installed openhab-js, I don’t see a way of also accounting for that without a lot of code complexity.

WDYT?

PR #2754, opened 2026-07-10
Append existing frontmatter from addon docs

jimtng:
Fixes https://github.com/openhab/openhab-addons/pull/21130#issuecomment-4934334229

In order to support https://github.com/openhab/openhab-website/pull/565 we need to include / merge the existing frontmatter from the document with the ones that we're generating.

Notes:
- The existing frontmatter is extracted and merged into the default one that we constructed here. Duplicate keys will be merged with the existing ones overriding our generated defaults.
- Some quick syntactical changes were automatically performed by Rubocop when saving the file.

PR #2760, opened 2026-07-17
Append existing frontmatter from addon docs - take 2

jimtng:
Add fixes for #2754

### DIFF
--- a/scripts/lib/addon_processor.rb
+++ b/scripts/lib/addon_processor.rb
@@ -2,6 +2,7 @@
 
 require "rexml/document"
 require "fileutils"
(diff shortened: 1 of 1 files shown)


jimtng:
@florian-h05 The second commit fixed the problems.

- I missed the custom quotations that was inserted manually
- By default the yaml output wrapped long lines and the "add metadata" code manually parses the frontmatter instead of ingesting it using a yaml parser, so it missed the line wraps. I have set the yaml output here to disable line wrapping.

florian-h05 (approved):
LGTM now!

PR #2766, opened 2026-08-19
Add examples for restoring only a group of Items, fix a doubled word

ML19821:
# Description

## Problem

The persistence page explains `restoreOnStartup` and warns against restoring everything, but has no example for the common setup in between: persist all Items, restore only a selected group. It also contains a doubled word ("strategies strategies") in the Items section.

## Changes

- Add two examples to "Restoring Item States on Restart" for the common setup the page warns towards but never shows: persist all Items, restore only the members of one group — once via the additive entry semantics (group entry carries only `restoreOnStartup`), once with disjoint entries (`*, !gRestore*`) for the case where the group entry carries store strategies of its own.
- Note explicitly that a store strategy listed in several entries matching the same Item stores the Item once per entry — a consequence of the additive semantics the page did not spell out before.
- Fix the doubled word "strategies strategies".

Note: for the additive example, the persistence health check currently reports a false "missing store strategy" warning — known issue openhab/openhab-core#5777, fix in review (openhab/openhab-core#5778). The configuration itself works as documented.

The example mirrors a configuration running on the author's production system.

This patch was developed with AI assistance (Claude); every commit carries an AI-assisted-by trailer.

### DIFF
--- a/src/configuration/persistence.md
+++ b/src/configuration/persistence.md
@@ -176,7 +176,7 @@ where `<itemlist>` is a comma-separated list consisting of one or more of the fo
   Note that `*` is NOT a wildcard match character in this context.
 - `!<groupName>*` all members of this group, but not the group itself, to be excluded from persistence defined by the other elements in this comma-separated `<itemList>`.
 The entries are additive.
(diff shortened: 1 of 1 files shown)


ML19821:
@florian-h05 the markdown check on this PR never started, it was waiting for workflow approval and expired today. Could you start it again? markdownlint with the repository's config reports nothing on the changed file.

